PR #135, opened 2014-02-10
Fixed Dependency issue and added if statement to init.pp

flakrat:
Fixed dependency version bug in Modulefile for the puppetlabs/apache module

Added if statement around the "class { 'puppet::config': } -> ..." code in init.pp
  So that a node that defines puppet::agent: and puppet::server: false won't have
  any changes applied to their puppet.conf. Without this, the puppet.conf is always
  potentially modifiable, even when not intended.


mmoll:
@flakrat what's the status here?

mmoll:
no reaction. closing this, I have a rebased version in the was_pr_135 branch of my fork and will probably open a new PR an some days.

PR #162, opened 2014-06-05
Make foreman integration optional

### LINKED ISSUE #154: Drop puppet-foreman requirement
In my opinion, puppet-puppet should not require puppet-foreman but the other way around, puppet should not be aware if foreman is going to use it or not.

It might require changes on puppet-foreman too.

That prevents the module to being used on it's own.

ekohl:
Instead of including foreman::params, the values are copied. That means 
foreman::params no longer needs to be included.

The next use of foreman is in puppet::server::config where it sets up 
reporting and ENC. This is wrapped in a new boolean $server_foreman. The user
will still need to take care of clearing $server_reports and
$server_external_nodes.

Fixes #154

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -268,6 +268,9 @@
 # $server_facts::                  Should foreman receive facts from puppet
 #                                  type:boolean
 #
+# $server_foreman::                Should foreman integration be installed
+#                                  type:boolean
+#
 # $server_foreman_url::            Foreman URL
 #
 # $server_foreman_ssl_ca::         SSL CA of the Foreman server
@@ -276,10 +279,7 @@
 #
 # $server_foreman_ssl_key::        Key for authenticating against Foreman server
 #
-#
 # $server_puppet_basedir::         Where is the puppet code base located
-#
-# $server_puppet_home::            Puppet var directory
 #
 # $server_enc_api::                What version of enc script to deploy. Valid
 #                                  values are 'v2' for latest, and 'v1'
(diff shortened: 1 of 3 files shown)


ekohl:
Rebased.

mmoll:
Seems to work just fine on my existing Debian/wheezy installation, so :+1:
Still have to test it on a fresh install.
The dependency should also be deleted from the metadata, or is there something like "recommends" for puppet modules? :)

ekohl:
Not sure about removing the dependency. Have to think about it.

roidelapluie:
@ekohl The dependecy should be removed and placed into README.md according to puppetlabs doc:

https://docs.puppetlabs.com/guides/style_guide.html

8.1 Dependencies

Hard dependencies must be declared explicitly in your module’s metadata.json file. Soft dependencies should be called out in the README.md, and must not be enforced as a hard requirement in your metadata.json. A soft dependency is a dependency that is only required in a specific set of use cases. (As an example, see the rabbitmq module.)

ekohl:
@roidelapluie it's good to know puppet has some guidelines about this. Still, if we default to true it'll be most use cases and the default install will be broken.

mmoll:
this needs a rebase now... Personally I'm tending to remove the dependency (esp. because of the good behaviour in librarian to resolve dependencies), but would merge it either way.

ekohl:
Since we default server to false, I now agree because it's not needed if you pass no parameters.

mmoll:
thanks!

domcleal:
@ekohl it seems the dep is still in metadata.json?

domcleal:
Oh, and since this removes a parameter (albeit fairly useless) we should either bump the major or keep it but deprecated.

roidelapluie:
@domcleal it is not only the API tha changes here but also the scope of the module as it can now be the reference module to setup puppet (without pulling all the dependencies of puppet-foreman) so I would go for a major bump

roidelapluie:
^ once we will have remove the dep from metadata.json

domcleal:
#230

PR #176, opened 2014-08-05
Ensure ca cert is accessible by the puppet group

lfisher47:
When the umask on the system is set to 0077, the puppet cert generate command sets the files to only be readable by the user.  Since we are not controlling the individual files, the easiest to way to set the permissions is to run with the a more global umask.


mmoll:
This should be fixed now by #273 et al. @lfisher47, could you check please?

lfisher47:
It looks like #273 takes care of the issue as a clean install on a system with a umask of 0077 worked for me.  I only tried with a version above 3.4.0+ of puppet though.  Closing.

PR #182, opened 2014-09-22
add PuppetDB configuration on puppet master side

mmoll:
more a WIP for discussion. comments please :newspaper:

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,7 @@ Optional support for installation of a Puppetmaster (using server => true)
   * Configurable support for static or git-backed dynamic environments (requires puppet-git module)
   * Storeconfig options (off, ActiveRecord or PuppetDB)
   * Passenger support (requires puppet-apache and puppet-passenger modules)
+  * PuppetDB integration (requires puppetlabs-puppetdb module)
 
 # Installation
 
@@ -79,6 +80,15 @@ wrapper classes or even your ENC (if it supports param classes). For example:
       server_git_repo_path     => '/var/lib/gitolite/repositories/puppet.git',
       server_post_hook_name    => 'post-receive.puppet',
       server_post_hook_content => 'puppetserver/post-hook.puppet',
+    }
+
+    # Want to integrate with an existing PuppetDB?
+    class { '::puppet':
+      server               => true,
+      server_puppetdb_host => 'mypuppetdb.example.com',
+      server_reports       => 'puppetdb',
+      storeconfigs         => true,
+      storeconfigs_backend => 'puppetdb',
     }
 
 Look in _init.pp_ for what can be configured this way, see Contributing if anything
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -243,6 +243,14 @@
 #                                  --vardir.  The default is an empty array.
 #                                  type:array
 #
+# $server_puppetdb_host::          PuppetDB host
+#
+# $server_puppetdb_port::          PuppetDB port
+#                                  type:integer
+#
+# $server_puppetdb_swf::           PuppetDB soft_write_failure
+#                                  type:boolean
+#
 # === Advanced server parameters:
 #
 # $server_httpd_service::          Apache/httpd service name to notify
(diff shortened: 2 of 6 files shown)


mmoll:
rebased

ekohl:
I wouldn't be against this, but I'm hesitant to add it as an unconditional dependency. I'm leaning to doing so because IMHO PuppetDB is a valuable addition in many cases.

mmoll:
now with added test. no idea what's going on with 2.7 there...

ekohl:
I wonder if a gem broke compatibility with 2.7.

mmoll:
I messed up metadata.json. All :green_heart: now

ekohl:
Maybe add a test to the 'with no customer parameters' that it should not contain the puppetdb class?

Other than that there's a puppetlabs recommendation that for soft dependencies it should be in README rather than metadata.json. See https://github.com/theforeman/puppet-puppet/pull/162#issuecomment-72476020 as well.

mmoll:
all done, I think

ekohl:
:+1: from my side. @domcleal?

mmoll:
needed a rebase again

README.md
```diff
@@ -71,6 +72,15 @@ wrapper classes or even your ENC (if it supports param classes). For example:
       server_external_nodes => '',
     }
 
+    # Want to integrate with an existing PuppetDB?
+    class { '::puppet':
+      server               => true,
+      server_puppetdb_host => 'mypuppetdb.example.com',
+      server_reports       => 'puppetdb',
```
domcleal: Nitpick, should we keep 'foreman' in here too?

manifests/init.pp
```diff
@@ -415,6 +423,9 @@
   $server_foreman_ssl_key        = $puppet::params::client_ssl_key,
   $server_facts                  = $puppet::params::server_facts,
   $server_puppet_basedir         = $puppet::params::server_puppet_basedir,
+  $server_puppetdb_host          = $puppet::params::server_puppetdb_host,
```
domcleal: validate_string on this?

manifests/init.pp
```diff
@@ -415,6 +423,9 @@
...
+  $server_puppetdb_port          = $puppet::params::server_puppetdb_port,
```
domcleal: validate_int?

manifests/init.pp
domcleal: Oh no, it's `undef`

domcleal:
Otherwise :+1:

manifests/init.pp
ekohl: We don't have it now, but what we could do:
```puppet
if $server_puppetdb_host {
  validate_string($server_puppetdb_host)
}
```

manifests/init.pp
ekohl: That doesn't exist afaik

manifests/init.pp
domcleal: Good idea.

manifests/init.pp
domcleal: You're right...

README.md
mmoll: As foreman got optional, I presume not, but I wouldn't mind having it also in just as example.

mmoll:
updated

domcleal:
Merged, thanks @mmoll!

README.md
ekohl: I do think people will be surprised if suddenly they don't have foreman reports anymore so I'd prefer having it in the example.

ekohl: And just now I see it was already merged.

domcleal: Ah, sorry.  Please feel free to push it straight to the repo, I guess it'd be `server_reports => 'puppetdb,foreman'`

ekohl: No problem, it's just a minor comment :)

ekohl: Fixed in d640db097b39531c5385ccfcffdb28ad4f405d89

PR #200, opened 2014-12-02
Allow to specify bind ip address of puppet master

roidelapluie:


mmoll:
please rebase

roidelapluie:
Done

manifests/params.pp
```diff
@@ -7,6 +7,7 @@
   $version             = 'present'
   $user                = 'puppet'
   $group               = 'puppet'
+  $ip                  = undef
```
domcleal: Should this be 'server_ip' for consistency with the class parameter name?

domcleal:
One question, otherwise :+1: 

We should probably check about setting the bind address on Puppet Server (and possibly the master webrick service) some time.

ekohl:
So I recall puppetlabs-apache does not like mixing IPs and names, but since it's undef it's fine. :+1: once the consistency issue is solved.

manifests/params.pp
roidelapluie: No, @domcleal 
```

$server_dir = $puppet::params::dir,
+ $server_ip = $puppet::params::ip,
$server_port = $puppet::params::port,
```

roidelapluie: or you have to modify the whole params.pp?

domcleal: Fair point!

domcleal:
I guess we'll overlook the inconsistency for now?

mmoll:
it's consistently "wrong", so :+1:

domcleal:
@ekohl does that mean this isn't going to work as-is?  IIRC this is what I hit when I tried the puppet-foreman patch: https://github.com/theforeman/puppet-foreman/pull/255#issuecomment-70073734

ekohl:
@domcleal as-is it shouldn't be any different from the current code, so I think this works. It may error out in unexpected ways for users.

domcleal:
I think we should just give it a test before merging, to check it doesn't fail like puppet-foreman..

ekohl:
I don't see why puppet-foreman would fail on that PR, so more testing would be very good.

roidelapluie:
Closing after inactivity

Issue #206, opened 2014-12-23
basemodulepath is written in [master] section. Mistake?


confiq:
When I'm using the following:
```puppet 
 class {
    '::puppet':
      server_common_modules_path => ['/etc/puppet/modules/X','/etc/puppet/modules/Y'],
```
module puppet writes config under [master] section
```
    basemodulepath   = /etc/puppet/modules/X:/etc/puppet/modules/Y
```
However, according to the [puppet docs](https://docs.puppetlabs.com/puppet/3.7/reference/dirs_modulepath.html#configuring-the-modulepath) it should be in _[main]_ section.
Even when I'm running command ```puppet config print modulepath``` configuration in [master] is not working and ```puppet module list``` ignores basemodulepath in  [master] 
This should be easy fix which even I might be able to fix it but the question is, is it a bug?

domcleal:
:+1: sounds fine to move to `[main]`.

domcleal:
Resolved in #213 and 447c744, thanks @confiq!

PR #213, opened 2015-01-05
issue #206

confiq:
basemodulepath is written in [master] section.
It should be simple as this.

### DIFF
--- a/templates/puppet.conf.erb
+++ b/templates/puppet.conf.erb
@@ -47,3 +47,7 @@
 
     syslogfacility = <%= @syslogfacility %>
 <% end -%>
+<% if scope.lookupvar("puppet::server_directory_environments") -%>
+    environmentpath  = <%= scope.lookupvar("puppet::server_envs_dir") %>
+    basemodulepath   = <%= [scope.lookupvar("puppet::server_common_modules_path")].flatten.join(':') %>
+<% end -%>
--- a/templates/server/puppet.conf.erb
+++ b/templates/server/puppet.conf.erb
@@ -14,12 +14,8 @@
     storeconfigs_backend = <%= @server_storeconfigs_backend %>
 <% end -%>
     strict_variables = <%= scope.lookupvar("puppet::server_strict_variables") %>
-
-<% if scope.lookupvar("puppet::server_directory_environments") -%>
-    environmentpath  = <%= scope.lookupvar("puppet::server_envs_dir") %>
-    basemodulepath   = <%= [scope.lookupvar("puppet::server_common_modules_path")].flatten.join(':') %>
-<% elsif scope.lookupvar("puppet::server_git_repo") ||
-        scope.lookupvar("puppet::server_dynamic_environments") -%>
+<% if !scope.lookupvar("puppet::server_directory_environments")  && 
+    ( scope.lookupvar("puppet::server_git_repo") || scope.lookupvar("puppet::server_dynamic_environments") ) -%>
     manifest       = <%= scope.lookupvar("puppet::server_envs_dir") %>/$environment/manifests/site.pp
     modulepath     = <%= scope.lookupvar("puppet::server_envs_dir") %>/$environment/modules
 <% end -%>
--- a/spec/classes/puppet_server_config_spec.rb
+++ b/spec/classes/puppet_server_config_spec.rb
@@ -180,8 +180,10 @@ describe 'puppet::server::config' do
 
       it 'should configure puppet.conf' do
         should contain_concat_fragment('puppet.conf+30-master').
-          with_content(%r{^\s+environmentpath\s+= /etc/puppet/environments$}).
           with_content(%r{^\s+config_version\s+= git --git-dir /etc/puppet/environments/\$environment/.git describe --all --long$})
+
+        should contain_concat_fragment('puppet.conf+10-main').
+          with_content(%r{^\s+environmentpath\s+= /etc/puppet/environments$})
       end
     end
 
@@ -221,7 +223,7 @@ describe 'puppet::server::config' do
       end
 
       it 'should configure puppet.conf' do
-        should contain_concat_fragment('puppet.conf+30-master').
+        should contain_concat_fragment('puppet.conf+10-main').
           with_content(%r{^\s+environmentpath\s+= /etc/puppet/environments\n\s+basemodulepath\s+= /etc/puppet/environments/common:/etc/puppet/modules:/usr/share/puppet/modules$})
       end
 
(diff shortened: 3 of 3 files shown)


templates/agent/puppet.conf.erb
```diff
@@ -32,4 +32,4 @@
     runinterval       = <%= scope.lookupvar('::puppet::runinterval') %>
     noop              = <%= scope.lookupvar('::puppet::agent_noop') %>
     configtimeout     = <%= scope.lookupvar('::puppet::configtimeout') %>
-    usecacheonfailure = <%= scope.lookupvar('::puppet::usecacheonfailure') %>
+    usecacheonfailure = <%= scope.lookupvar('::puppet::usecacheonfailure') %>
```
ekohl: Please keep the newline at the end of the file here.

templates/puppet.conf.erb
```diff
@@ -47,3 +47,13 @@
 
     syslogfacility = <%= @syslogfacility %>
 <% end -%>
+
+    
+<% if scope.lookupvar("puppet::server_directory_environments") -%>
+    environmentpath  = <%= scope.lookupvar("puppet::server_envs_dir") %>
+    basemodulepath   = <%= [scope.lookupvar("puppet::server_common_modules_path")].flatten.join(':') %>
+	<% elsif scope.lookupvar("puppet::server_git_repo") ||
```
ekohl: Odd indenting here.

templates/puppet.conf.erb
```diff
@@ -47,3 +47,13 @@
...
+	        scope.lookupvar("puppet::server_dynamic_environments") -%>
+    manifest       = <%= scope.lookupvar("puppet::server_envs_dir") %>/$environment/manifests/site.pp
```
ekohl: Should the legacy way also be in ```[main]``` rather than ```[server]```?

ekohl:
We have tests to verify the contents of the config files which should be updated as well.

templates/puppet.conf.erb
confiq: I was trying "Geppeto" as IDE and it made it wrong. Will fix in next commit.

templates/puppet.conf.erb
confiq: I assume that "legacy way" is when we use `server_dynamic_environments` or `server_git_repo` with `manifest` directory. 
Anyway `manifest` should be in `[master] ` [section](https://docs.puppetlabs.com/puppet/latest/reference/dirs_manifest.html#with-puppet-master) and not `[main]` or `[server]`.
I will try to fix this too.

ekohl: I believe that ```[master]``` is an environment while this is the generic case since it refers to ```$environment```. That means it should be at least in ```[server]``` as it is now. That way you have the directory environments behaviour even though that was never intentionally supported.

confiq:
I've upgraded the tests to new code. 
Can you code review it and tell me if I've missed something?

templates/puppet.conf.erb
```diff
@@ -47,3 +47,7 @@
 
     syslogfacility = <%= @syslogfacility %>
 <% end -%>
+<% if scope.lookupvar("puppet::server_directory_environments") -%>
+    environmentpath  = <%= scope.lookupvar("puppet::server_envs_dir") %>
+    basemodulepath   = <%= [scope.lookupvar("puppet::server_common_modules_path")].flatten.join(':') %>
+<% end -%>
```
ekohl: We prefer line endings here.

templates/server/puppet.conf.erb
```diff
@@ -29,4 +26,4 @@
 <% end -%>
 <% scope.lookupvar("puppet::server_additional_settings").sort_by {|k, v| k}.each do |key, value| -%>
     <%= key %> = <%= value %>
-<% end -%>
+<% end -%>
```
ekohl: We prefer line endings here.

ekohl:
Other than the trivial nitpicks I think it's good. If you can fix those and squash everything in a single commit, I think it's good to go.

confiq:
I don't PR very often so, when you say "_squash everything in a single commit_" you mean git rebase squash? Should I rewrite git history?

ekohl:
Yes, I do mean rewrite git history. We should really document this, but http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html or http://davidwalsh.name/squash-commits-git describe the proces.

confiq:
Ohh, there is no problem rewriting history, just didn't know it's something we usually do in PR.
Anyway, I've squash all my commits to one. 
Cheers

ekohl:
We prefer it in our repos to keep our history sane, but I know many people don't care.

Anyway, :+1:

templates/server/puppet.conf.erb
```diff
@@ -15,10 +15,7 @@
 <% end -%>
     strict_variables = <%= scope.lookupvar("puppet::server_strict_variables") %>
 
-<% if scope.lookupvar("puppet::server_directory_environments") -%>
-    environmentpath  = <%= scope.lookupvar("puppet::server_envs_dir") %>
-    basemodulepath   = <%= [scope.lookupvar("puppet::server_common_modules_path")].flatten.join(':') %>
-<% elsif scope.lookupvar("puppet::server_git_repo") ||
+<% if scope.lookupvar("puppet::server_git_repo") ||
```
domcleal: These conditionals aren't quite right now for a couple of cases.. it used to work thanks to the "elsif":

1. When both server_directory_environments & server_git_repo are true, it used to only add environmentpath/basemodulepath.  I think now it will add env/basemodulepath into `[main]` and also add manifest/modulepath in `[master]` which isn't right.
2. ditto with server_dynamic_environments

I think this if statement needs to be: `if !server_directory_environments && (server_git_repo || server_dynamic_environments)` (pseudo-code)

confiq: You got good point there. I've commit and squash all commits. Can you check again?
Travis CI is happy but we might add this check for the future.

confiq:
Hi,
Please tell me if this PR is missing something else...

templates/puppet.conf.erb
```diff
@@ -47,3 +47,8 @@
...
+
```
ekohl: Nitpick: empty line here

ekohl:
Trivial nitpick, but I think it's good. @domcleal?

templates/puppet.conf.erb
confiq: I've remove \n. If/When @domcleal approve it I'll squash the commits before merging

domcleal:
:+1: tests fine.

ekohl:
:+1:

domcleal:
Merged as 447c744, thanks @confiq!

PR #214, opened 2015-01-05
Apply server_environments_{owner,group,mode} to environments subdirs

martinpfeifer:
If Puppet doesn't manage those directories, they are owned by root and are not writable by e. g. a special code deployment user.


ekohl:
I'd prefer to make those values parameters so they can be overridden per environment (but still have a sane default).

martinpfeifer:
I made them parameters which default to server_environments_{owner,group,mode}.

domcleal:
:+1:

ekohl:
Thanks!

PR #215, opened 2015-01-06
add windows support

cristi1979:
add windows support for puppet

### DIFF
--- a/manifests/agent/service.pp
+++ b/manifests/agent/service.pp
@@ -10,8 +10,10 @@ class puppet::agent::service {
         enable    => true,
       }
 
-      cron { 'puppet':
-        ensure => absent,
+      if $::osfamily != 'windows' {
+        cron { 'puppet':
+          ensure => absent,
+        }
       }
     }
     'cron': {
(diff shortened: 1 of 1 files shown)


manifests/agent/service.pp
```diff
@@ -29,11 +31,23 @@
...
-        hour    => $times[0],
-        minute  => $times[1],
+      if $::osfamily == 'windows' {
+        scheduled_task { 'puppet':
+          ensure  => present,
+          enabled => true,
+          command => $command,
+          trigger => {
+            schedule   => daily,
+            start_time => $times[0],
```
ekohl: This is a different schedule. Maybe move the ip_to_cron part to the non-windows part and just try to parse runinterval directly?

cristi1979: This document: https://docs.puppetlabs.com/puppet/3.6/reference/resources_scheduled_task_windows.html it says:
Puppet does not support “every X minutes” type triggers.

Also there is PUP-1291 that waits to be implemented.

I couldn't find how to do this type of scheduling with puppet on windows.

Also I noticed that $times[0] may be '*' which is not ok with windows. Should the schedule for windows be set to an hard coded hour, or removed?
The only thing that stops the module from running in windows is the cron type.

ekohl: Maybe just error out of the user tries to set cron on windows? Then on absent you can just hide the cron part inside the if block. That way we only support puppet as deamon on windows which IMHO is good enough for now.

mmoll:
Looks good to me :+1: 
Out of curiosity are you using that module for Windows nodes? Does it work just out of the box with the defaults?

ekohl:
:+1: if you can squash it.

cristi1979:
i will close this and opens a new PR with only 1 commit

ekohl:
You can force push to a branch in which case you don't need to open a new PR.

PR #218, opened 2015-01-08
Bugfix: removed dynamic_environment from config_version if statement

visibilityspots:
When using the param server_dynamic_environments I got this warning in my puppet runs:
```puppet
Warning: Setting config_version is deprecated in puppet.conf. See http://links.puppetlabs.com/env-settings-deprecations
   (at /usr/lib/ruby/site_ruby/1.8/puppet/settings.rb:1141:in `issue_deprecation_warning')
```

Since the dynamic environments are taking over from the config-file environments setting this config_version configuration didn't make much sense I think when the dynamic_environments param is setted to true?

### DIFF
--- a/templates/server/puppet.conf.erb
+++ b/templates/server/puppet.conf.erb
@@ -14,13 +14,12 @@
     storeconfigs_backend = <%= @server_storeconfigs_backend %>
 <% end -%>
     strict_variables = <%= scope.lookupvar("puppet::server_strict_variables") %>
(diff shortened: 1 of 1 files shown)


ekohl:
Are you sure you wouldn't rather be using ```server_directory_environments``` which is the proper puppet way of supporting directory based environments. I must admit we have a few to many options here to support and it's hard to see.

ekohl:
No problem. Just a signal that we should improve our documentation.

PR #220, opened 2015-01-24
Updated agent config and tests to work on Windows.

genebean:


mmoll:
Doesn't change anything on my Debian and looks sensible, so :+1: from me in general, however, there might be some overlap with $foreman::params::puppet_home regarding $puppet::params::dir_prefix.

Nitpick: some empty lines contains white space, visible in the diff.

mmoll:
maybe @cristi1979 can say something about the Windows part?

genebean:
I can certainly go in an clean up the whitespace... it's an artifact of Geppetto trying to help format things. I'll do that when I get back home in a few hours.

genebean:
I went through and cleaned up the whitespace

manifests/agent/install.pp
```diff
@@ -1,8 +1,15 @@
...
+      package { $puppet::params::client_package:
+        ensure   => $::puppet::version,
+        provider => $::puppet::params::win_provider,
+      }
+    }
 
-  package { $puppet::params::client_package:
-    ensure => $::puppet::version,
+    default : {
+      package { $puppet::params::client_package: ensure => $::puppet::version, }
```
mmoll: I'd like to keep the package block with line breaks.

metadata.json
```diff
@@ -25,6 +25,7 @@
     { "operatingsystem": "Scientific", "operatingsystemrelease": [ "6", "7" ] },
     { "operatingsystem": "Fedora",     "operatingsystemrelease": [ "19" ] },
     { "operatingsystem": "Debian",     "operatingsystemrelease": [ "6", "7" ] },
-    { "operatingsystem": "Ubuntu",     "operatingsystemrelease": [ "12.04", "14.04" ] }
+    { "operatingsystem": "Ubuntu",     "operatingsystemrelease": [ "12.04", "14.04" ] },
+    { "operatingsystem": "Windows",    "operatingsystemrelease": [ "7", "2012 R2" ] }
```
mmoll: What about older versions?

manifests/agent/install.pp
genebean: Dang Geppetto... that is something I feel is broken about their formatting... I agree this needs changing back, I just overlooked it.

metadata.json
genebean: Let me take a look at this. I saw more details in the Puppet Labs docs I used in my research.

genebean: Puppet's list of what they support is at https://docs.puppetlabs.com/guides/install_puppet/install_windows.html#supported-versions
My setting should work with Vista - 8 and Server 2008 - 2012 R2. Server 2003 would require different defaults because of the difference in the data directory (https://docs.puppetlabs.com/pe/latest/install_windows.html#data-directory)

What are your thoughts on which ones should be in this file?

mmoll: 2003 is dead anyway now, but the others should be in there, IMHO.

metadata.json
```diff
@@ -25,6 +25,10 @@
     { "operatingsystem": "Scientific", "operatingsystemrelease": [ "6", "7" ] },
     { "operatingsystem": "Fedora",     "operatingsystemrelease": [ "19" ] },
     { "operatingsystem": "Debian",     "operatingsystemrelease": [ "6", "7" ] },
-    { "operatingsystem": "Ubuntu",     "operatingsystemrelease": [ "12.04", "14.04" ] }
+    { "operatingsystem": "Ubuntu",     "operatingsystemrelease": [ "12.04", "14.04" ] },
+    { "operatingsystem": "Windows",    "operatingsystemrelease": [ 
+        "Vista", "7", "8",
+        "2008", "2008 R2", "2012", "2012 R2"
+      ] }
   ]
```
genebean: How does this look? I broke up the desktop versions and the server versions since the list is so long... it made it much easier to read.

manifests/params.pp
```diff
@@ -34,6 +33,31 @@
...
+
+    default: {
+      $dir                 = "${dir_prefix}/etc/puppet"
+      $logdir              = "${dir_prefix}/var/log/puppet"
+      $rundir              = "${dir_prefix}/var/run/puppet"
+      $ssldir              = '$vardir/ssl'
+    }
+  }
+
+  $win_provider        = 'chocolatey'
```
ekohl: I'd prefer to keep this generic, so ```$package_provider``` with ```undef``` for non-windows.

manifests/agent/install.pp
```diff
@@ -1,8 +1,17 @@
 # Install the puppet client installation
 class puppet::agent::install {
+  case $::osfamily {
```
ekohl: If the provider is generic, you can remove the os-specific logic here.

ekohl:
I have no experience with puppet on windows, so I trust you on it. I'd prefer to keep the os-specific logic out of the classes where possible, but not sure how you can easily do that in the config case other than create a ton of variables.

manifests/agent/install.pp
```diff
@@ -1,8 +1,13 @@
...
+  if ($::puppet::params::package_provider) {
+    package { $puppet::params::client_package:
+      ensure   => $::puppet::version,
+      provider => $::puppet::params::package_provider,
+    }
+  } else {
+    package { $puppet::params::client_package:
+      ensure => $::puppet::version,
+    }
   }
```
genebean: @ekohl I think I addressed this the way you had in mind, if not just let me know. I actually agree wholeheartedly about keeping the settings generic.

genebean:
BTW, I just realized that there is no test for `puppet::agent::install` and, with this new logic, there should be... fixing now.

manifests/agent/install.pp
ekohl: I think you can even leave out the if. When it's set to undef, the OS default should be used.

spec/classes/puppet_agent_install_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+      } end
+
+      it 'should define provider as chocolatey' do
+        should contain_package('puppet').with_provider('chocolatey')
+      end
+    end
+
+  end
+
+end
```
genebean: Still working out some tests...

manifests/agent/install.pp
```diff
@@ -1,8 +1,7 @@
 # Install the puppet client installation
 class puppet::agent::install {
-
-  package { $puppet::params::client_package:
-    ensure => $::puppet::version,
+  package { $puppet::client_package:
+    ensure   => $::puppet::version,
+    provider => $::puppet::package_provider,
```
genebean: @ekohl - I learned something new today thanks to this... didn't know it would work with `provider => undef`

spec/classes/puppet_agent_install_spec.rb
genebean: Tests are finished now

manifests/agent/install.pp
ekohl: I'll be sure to test this, but it should work.

genebean: I tested it on CentOS & it worked.

metadata.json
mmoll: This is really nitpicky now, I'd like it more like that:
```
{ "operatingsystem": "Windows", "operatingsystemrelease":
  [ "Vista", "7", "8",
    "2008", "2008 R2", "2012", "2012 R2" ]
}
```
or 
```
{ "operatingsystem": "Windows", "operatingsystemrelease":
  [
  "Vista", "7", "8",
  "2008", "2008 R2", "2012", "2012 R2"
  ]
}
```
or just put it in one line, it's JSON anyway and not meant to be human readable ;)

manifests/agent/install.pp
```diff
@@ -1,8 +1,7 @@
 # Install the puppet client installation
 class puppet::agent::install {
-
-  package { $puppet::params::client_package:
-    ensure => $::puppet::version,
+  package { $puppet::client_package:
```
mmoll: ::params::?

genebean: Based on params not being passed directly to this class or any other that I can find, it seems the standard throughout this module is to use `$puppet::some_param_name`. I thought the `::params::` that was in here before was my doing... guess I actually found a bug while modifying this manifest.

metadata.json
```diff
@@ -25,6 +25,12 @@
...
     { "operatingsystem": "Debian",     "operatingsystemrelease": [ "6", "7" ] },
-    { "operatingsystem": "Ubuntu",     "operatingsystemrelease": [ "12.04", "14.04" ] }
+    { "operatingsystem": "Ubuntu",     "operatingsystemrelease": [ "12.04", "14.04" ] },
+    { "operatingsystem": "Windows",    "operatingsystemrelease": 
+      [ 
+        "Vista", "7", "8",
+        "2008", "2008 R2", "2012", "2012 R2"
+      ]
+    }
   ]
```
genebean: @mmoll let's go with this if it works for you. Even though this is "just JSON" it does help when it's easy to read. I assume that's why the other OS's are on a single line. An alternative is to apply standard JSON formatting like puppet_blacksmith, Geppetto, and jsonlint.com do. That makes it look like this:
```json
"operatingsystem_support": [
        {
            "operatingsystem": "RedHat",
            "operatingsystemrelease": [
                "6",
                "7"
            ]
        },
        {
            "operatingsystem": "CentOS",
            "operatingsystemrelease": [
                "6",
                "7"
            ]
        },
        {
            "operatingsystem": "Scientific",
            "operatingsystemrelease": [
                "6",
                "7"
            ]
        },
        {
            "operatingsystem": "Fedora",
            "operatingsystemrelease": [
                "19"
            ]
        },
        {
            "operatingsystem": "Debian",
            "operatingsystemrelease": [
                "6",
                "7"
            ]
        },
        {
            "operatingsystem": "Ubuntu",
            "operatingsystemrelease": [
                "12.04",
                "14.04"
            ]
        },
        {
            "operatingsystem": "Windows",
            "operatingsystemrelease": [
                "Vista",
                "7",
                "8",
                "2008",
                "2008 R2",
                "2012",
                "2012 R2"
            ]
        }
    ]
```

manifests/agent/install.pp
mmoll: Right, I see now, overriding via class parameter probably didn't work.

metadata.json
mmoll: that's working for me :)

manifests/agent/install.pp
genebean: BTW @mmoll, sorry if that came across crappy... I just reread that and realized I could have worded it better.

mmoll: no problem :)

genebean:
@mmoll & @ekohl - all good now?

ekohl:
I think it's looking good, though I only did a code review for now.

/cc @domcleal

mmoll:
:+1: from me also.

genebean:
Thank you both for all the feedback! I can honestly say that I got a lot out of it and that I think this is better code as a result.

manifests/init.pp
```diff
@@ -13,8 +13,20 @@
 #
 # $group::                         Override the name of the puppet group.
 #
+# $dir_prefix::                    Override the where to find the data directory on Windows.
```
domcleal: I don't think this is used in the class, only inside params.  Does it need to be a parameter, or should we join the paths inside the class instead of params?

manifests/params.pp
```diff
@@ -34,6 +33,34 @@
...
+  case $::osfamily {
+    Windows : {
+      $dir    = "${dir_prefix}/etc"
+      $logdir = "${dir_prefix}/var/log"
+      $rundir = "${dir_prefix}/var/run"
+      $ssldir = '$confdir/ssl'
+    }
+
+    default : {
+      $dir    = "${dir_prefix}/etc/puppet"
```
domcleal: This might be answered by my previous comment, but $dir_prefix is always empty at this point for non-Windows platforms.

manifests/init.pp
mmoll: as said, this could overlap with $foreman::params::puppet_home

genebean: Good point. My thought for making it a param was that people can install things in alternate locations such as the D: drive, the problem is that I didn't code it that way. I am removing the param.

manifests/params.pp
genebean: You are correct. Like in my last comment, I had intended for this to accommodate alternate install paths such as D: on Windows and /usr/local on Linux. I am removing the param and will clean this up too.

genebean:
@domcleal updates finished... how's that?

manifests/init.pp
domcleal: Yeah, I thought that was the direction you were going in.  This works too, it's just a little more work.

manifests/init.pp
```diff
@@ -307,7 +317,12 @@
   $version                       = $puppet::params::version,
   $user                          = $puppet::params::user,
   $group                         = $puppet::params::group,
+  $dir_prefix                    = $puppet::params::dir_prefix,
```
domcleal: This line also needs removing, as the docs are now gone.

domcleal:
:+1: once the last parameter's removed.

genebean:
Line removed

manifests/init.pp
```diff
@@ -308,6 +318,10 @@
   $user                          = $puppet::params::user,
   $group                         = $puppet::params::group,
   $dir                           = $puppet::params::dir,
```
ekohl: Maybe we can use ```validate_absolute_path``` on all these dirs, though I'm not sure if that also works for windows paths.

ekohl:
I think it looks good! :+1: even without the validate_absolute_path.

mmoll:
according to the docs ```validate_absolute_path``` also works for windows

genebean:
Checking out `validate_absolute_path` now...

genebean:
Not sure we want to do this as puppet can be installed without some of the directories existing yet or they may get created as a byproduct of changing the config file... I think I would lean towards no validating the paths. A prime example of this would be using `puppet apply` to take a minimally bootstrapped agent install and making it the way you want it in production.

mmoll:
As far as I understand, the validation is only taking place on a string level, not in the file system actually. See https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/lib/puppet/parser/functions/validate_absolute_path.rb

ekohl:
Looking at https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/lib/puppet/parser/functions/validate_absolute_path.rb it only validates that it's an absolute path (as opposed a relative path), but not that it actually exists as I read it.

genebean:
Ahh, I misunderstood what it does then. I will add that in and test it tomorrow afternoon. I'll also add a test that validates a relative path fails.

genebean:
How's that look?

manifests/init.pp
```diff
@@ -414,6 +428,10 @@
   validate_array($dns_alt_names)
   validate_array($auth_allowed)
 
+  validate_absolute_path($dir)
+  validate_absolute_path($logdir)
+  validate_absolute_path($rundir)
+
```
mmoll: ssldir?

mmoll: ah, that's getting bulit from other vars, scrap that...

mmoll:
@genebean seriously sincere thanks for staying with us :bow:

mmoll:
merged as 9be5a6c, thanks @genebean :tada: :balloon:

genebean:
@mmoll Glad to do it! I actually quite enjoyed it.

genebean:
@mmoll Just for educational reasons, what was the merge process for this? I ask because I see it shows as "Closed with unmerged commits". Just trying to learn when I can.

mmoll:
I cherry-picked your commit into master, as it couldn't be merged with --ff-only. We want a linear history without merge commits from pull requests in our repos. See http://blog.spreedly.com/2014/06/24/merge-pull-request-considered-harmful/ for a short overview on that topic. BTW, feel also free to join our IRC channel(s) #theforeman(-dev) on freenode for realtime communications.

genebean:
Totally makes sense, thanks.

ekohl:
Like @mmoll said, thanks for the patch and addressing all concerns. Also, #theforeman(-dev) is a friendly and helpful place.

PR #222, opened 2015-01-27
Do not set config_version when using directory envs

ekohl:
According to the [docs](https://docs.puppetlabs.com/puppet/latest/reference/config_file_environment.html#configversion), config_version can't be used with directory environments since it will never be used. This also removes a deprecation warning users of directory environments will see.

### DIFF
--- a/spec/classes/puppet_server_config_spec.rb
+++ b/spec/classes/puppet_server_config_spec.rb
@@ -179,8 +179,7 @@ describe 'puppet::server::config' do
       end
 
       it 'should configure puppet.conf' do
-        should contain_concat_fragment('puppet.conf+30-master').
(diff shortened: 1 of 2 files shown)


domcleal:
Cool.  We should double check the result, but it looks like we default puppet::server::env's config_version parameter to puppet::server::config_version_cmd, so hopefully it'll populate config_version inside environment.conf instead.

ekohl:
I'll fix the tests and have a look at puppet::server::env. One difference is that environment.conf needs an absolute path where puppet.conf can look at $PATH. That makes it harder to be portable.

ekohl:
Rebased so the tests pass

mmoll:
thanks @ekohl :tada:

Issue #157, opened 2014-05-22
Add support for [pre|post]run_command in agent configuration


lvg:
Add support for pre- and postrun_command options in the puppet.conf agent section. Values are expected to be strings (including quotation when using spaces)

PR #223, opened 2015-01-27
add support for [pre|post]run_command for agent

mmoll:
closes GH-157

### DIFF
--- a/templates/agent/puppet.conf.erb
+++ b/templates/agent/puppet.conf.erb
@@ -33,3 +33,9 @@
     noop              = <%= scope.lookupvar('::puppet::agent_noop') %>
     configtimeout     = <%= scope.lookupvar('::puppet::configtimeout') %>
     usecacheonfailure = <%= scope.lookupvar('::puppet::usecacheonfailure') %>
+<% unless [nil, :undefined, :undef].include?(scope.lookupvar('::puppet::prerun_command')) -%>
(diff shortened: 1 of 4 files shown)


templates/agent/puppet.conf.erb
```diff
@@ -33,3 +33,9 @@
     noop              = <%= scope.lookupvar('::puppet::agent_noop') %>
     configtimeout     = <%= scope.lookupvar('::puppet::configtimeout') %>
     usecacheonfailure = <%= scope.lookupvar('::puppet::usecacheonfailure') %>
+<% if scope.lookupvar('::puppet::prerun_command') and !scope.lookupvar('::puppet::prerun_command').to_s.empty? -%>
+    prerun_command    = <%= scope.lookupvar('::puppet::prerun_command') %>
```
ekohl: Is undef.to_s.empty empty?

mmoll: not really, but if the first check is false, the second doesn't get evaluated (at least I'm not seeing any errors)

domcleal: Sometimes (maybe depending on how the class is called) undef gets passed through as nil.. or maybe that's just in rspec-puppet, so this would work.  But often with scope.lookupvar, undef will return :undef or :undefined...

mmoll:
updated paramter check and tests

domcleal:
:+1:

ekohl:
Thanks!

PR #224, opened 2015-01-28
do not use validate_string on pre/postrun_command

mmoll:


domcleal:
:+1:

ekohl:
Any particular reason you don't want to validate this?

domcleal:
Our system tests failed, we're getting:

    [ERROR 2015-01-28 13:03:11 verbose] validate_string(): wrong number of arguments (0; must be > 0)

when Kafo runs its validation checks before the main catalog.

On closer inspection I see that validate_string(undef) is meant to fail, so given the default value we probably shouldn't be validating them.  Or we should use '' as the default value..

ekohl:
Ah, just read IRC. I wouldn't mind an empty string as default but I have no strong preference now. :+1: either way.

domcleal:
Merged, thanks @mmoll!

(I looked at the wrong Travis window, I hope it passes...)

PR #227, opened 2015-01-31
Add parameter server_rack_arguments

treydock:
* Default is an empty array
* Elements are added to ARGV in Puppet server's rack config.ru

This is an ugly work around to segfaults that occur when running Puppet under Passenger in Ruby 1.8.7 (CentOS 6.x).  The only way I've found to work around the segfaults is to enable profiling by appending the "--profile" flag to ARGV in config.ru.

### DIFF
--- a/templates/server/config.ru.erb
+++ b/templates/server/config.ru.erb
@@ -18,6 +18,9 @@ ARGV << "--rack"
 # to ~puppet/.puppet
 ARGV << "--confdir" << "<%= scope.lookupvar('::puppet::server_dir') %>"
 ARGV << "--vardir"  << "<%= scope.lookupvar('::puppet::server_vardir') %>"
+<% (@server_rack_arguments || []).each do |server_rack_argument| -%>
+ARGV << "<%= server_rack_argument %>"
+<% end -%>
 
 # NOTE: it's unfortunate that we have to use the "CommandLine" class
 #  here to launch the app, but it contains some initialization logic
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -239,6 +239,10 @@
 #                                  Example: {trusted_node_data => true, stringify_facts => true, ordering => 'manifest'}
 #                                  type:hash
 #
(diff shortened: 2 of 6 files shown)


ekohl:
:+1: I think the test failure is unrelated our tests are broken on puppet 2.7.

templates/server/config.ru.erb
```diff
@@ -18,6 +18,9 @@ ARGV << "--rack"
 # to ~puppet/.puppet
 ARGV << "--confdir" << "<%= scope.lookupvar('::puppet::server_dir') %>"
 ARGV << "--vardir"  << "<%= scope.lookupvar('::puppet::server_vardir') %>"
+<% (scope.lookupvar('::puppet::server_rack_arguments') || []).each do |server_rack_argument| -%>
```
domcleal: I think this is causing the test failure, as server_rack_arguments can be `:undefined` (or `:undef`)

treydock:
@domcleal the option server_rack_arguments shouldn't be undefined since the default in puppet::params is an empty array.  Since the test failed in puppet::passenger spec my guess is the tests are behaving such that it's like calling `class { 'puppet::passenger': }` directly rather than what I'd assume to be the only public class, `class { 'puppet': }`.

I think I only way to handle "undef" properly in all versions of Puppet is to reference the value using instance variable which is defined in the template's calling class.  https://docs.puppetlabs.com/guides/templating.html#testing-for-undefined-variables

Will update.

treydock:
PR updated.  Ran into issue after removing my local Gemfile.lock and running `bundle install` that is related to rspec-puppet now being at version 2.0.0.  The use of `subject` no longer works for the methods in `spec/lib/module_spec_helper.rb`, but instead the subject argument must be passed using `catalogue`.  The unit tests wouldn't pass without that change so hopefully it's okay that I've included that change in this PR.

treydock:
Disregard my comments about issues with rspec-puppet.  I see now that my branch was not up-to-date with master and @domcleal already addressed that issue.  Updating PR yet again and sorry for the noise.

mmoll:
thanks @treydock :tada:

PR #229, opened 2015-02-04
Let puppet::server::service notify passenger

ekohl:
This complements https://github.com/theforeman/puppet-foreman/pull/282

### DIFF
--- a/manifests/server/passenger.pp
+++ b/manifests/server/passenger.pp
@@ -84,4 +84,5 @@ class puppet::server::passenger (
     require              => Class['::puppet::server::rack'],
   }
 
+  Class['puppet::server::service'] ~> Class['puppet::server::passenger']
 }


domcleal:
So aside from the simple test failures, there's one about a dependency cycle as the puppetmaster class is part of the config stage (in server.pp) and so having a service class notify a config class is backwards.

What we're really trying to do is have the service class notify Exec[puppet_server_rack-restart] which is part of the rack class, which is called by the passenger class.  Unfortunately rack is also a dependency of passenger, so we have `rack -> passenger -> config -> service` or similar at the moment.

I think it's probably best to move the exec from rack into the service class, or maybe a service::rack class.

mmoll:
what's the status here? :)

mmoll:
can this get closed after #456?

domcleal:
I think so, it looks like the same thing.

PR #230, opened 2015-02-04
Remove the strict puppet-foreman requirement

ekohl:
This amends ff942a72bf1918fa42bea77e82f7c1d7d83e69ef / https://github.com/theforeman/puppet-puppet/issues/154 which made the code change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,23 @@ Standalone agent with defaults:
 
     echo include puppet | puppet --modulepath /path_to/extracted_tarball
 
+# Integration with foreman
+
+With the 3.0.0 release the Foreman integration became optional. Note that you
+need the [puppet-foreman module](https://github.com/theforeman/puppet-foreman)
+installed.
+
+It will still by default install the Foreman integration when `server` is true,
+so if you wish to run a puppet server without Foreman, it can be disabled by
+setting `server_foreman` to false:
+
+    class { '::puppet':
+      server                => true,
+      server_foreman        => false,
+      server_reports        => 'store',
+      server_external_nodes => '',
+    }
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -30,6 +30,21 @@ Standalone agent with defaults:
 
     echo include puppet | puppet --modulepath /path_to/extracted_tarball
 
+# Integration with foreman
+
+With the 3.0.0 release the foreman integration became optional. Note that you
```
domcleal: *Foreman

README.md
```diff
@@ -30,6 +30,21 @@ Standalone agent with defaults:
...
+need the [puppet-foreman module](https://github.com/theforeman/puppet-foreman)
+installed.
+
+It will still by default install the foreman integration when server is true,
```
domcleal: ditto

README.md
```diff
@@ -64,13 +79,6 @@ wrapper classes or even your ENC (if it supports param classes). For example:
       server_post_hook_content => 'puppetserver/post-hook.puppet',
     }
 
-    # Perhaps you want to install without foreman?
-    class { '::puppet':
-      server                => true,
-      server_reports        => 'store',
```
domcleal: Do you need to keep these params in the above example?

ekohl: You're right, we still default to ```foreman``` there.

README.md
ekohl: You mean literal with backticks? Isn't that a github flavor instead of plain markdown?

ekohl:
Amended

domcleal:
Thanks!

PR #232, opened 2015-02-05
Fix SSL default variable names to match params.pp

domcleal:
Fixes kafo putting the string "puppet::params::client_ssl_cert" etc as the value for these parameters, resulting in:

    Feb  5 10:33:32 localhost puppet-master[18099]: Report processor failed: Could not send report to Foreman at https://foreman-el7.example.com/api/reports: No such file or directory - puppet::params::client_ssl_cert


ekohl:
:+1: odd that it even compiled in the first place.

domcleal:
I think because we use a pick() later on, pure Puppet would cause it to pick another without a problem as the parameter would be undef, but kafo doesn't evaluate parameter defaults in quite the same way...

ekohl:
That could explain it. I suppose we should put some more effort into the strict variables.

PR #234, opened 2015-02-06
Make theforeman/git a soft dependency

domcleal:
Following the change for theforeman/foreman and puppetlabs/puppetdb.


mmoll:
:+1:, would be nice to link puppet-git in the README

domcleal:
Ah yes, which reminds me, shall I link to the Puppet Forge page do you think rather than the GitHub repo?

mmoll:
Personally I would link the forge page, but it seems that gitrepo is the standard at the moment.

domcleal:
Hopefully that's standardised properly now  :)

mmoll:
yes, thanks @domcleal!

PR #235, opened 2015-02-08
puppet still need foreman as a dependency

cristi1979:
Since the module uses class {'foreman::puppetmaster', it needs to have foreman as a dependency.

The same for git module and puppetdb

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -11,7 +11,10 @@
   "tags":         ["foreman", "puppet", "puppetmaster", "puppet-server"],
   "dependencies": [
     { "name": "theforeman/concat_native", "version_requirement": ">= 1.3.0 < 2.0.0" },
+    { "name": "theforeman/foreman",       "version_requirement": ">= 3.0.0 < 4.0.0" },
+    { "name": "theforeman/git",           "version_requirement": ">= 1.0.0 < 2.0.0" },
     { "name": "puppetlabs/apache",        "version_requirement": ">= 1.2.0 < 2.0.0" },
+    { "name": "puppetlabs/puppetdb",      "version_requirement": ">= 4.0.0 < 5.0.0" },
     { "name": "puppetlabs/stdlib",        "version_requirement": ">= 2.0.0 < 5.0.0" }
   ],
   "requirements": [


domcleal:
Our intention was to only keep these as soft dependencies, what do you think?  Maybe we should make the default value of server_foreman false.

cristi1979:
This will make users put the foreman and puppetdb modules as a dependency to an other custom module (a false dependency).

We install an entire environment by doing something like this (but this should apply when installing modules from puppet forge also):

    puppet module install environment_name-role --module_repository=http://pulp_forge.company.net/pulp_puppet/forge/repository/puppet --target-dir /path --config /dev/null

So if foreman doesn't exist as a dependency, it will not be installed. The same is for puppetdb in our case.

Most of our configuration is in hiera, so we don't touch the puppet code to set a puppet master or a puppet agent. We do it by setting the appropriate hostgroup in foreman and using hiera overrides.

domcleal:
Typically I think users would be either using something like r10k or librarian-puppet and could add the dependency explicitly in their Puppetfile.  e.g. if you're expecting to integrate puppet-puppet with puppetlabs/puppetdb, surely you'd just add the latter module too?

https://docs.puppetlabs.com/guides/style_guide.html#dependencies seems to suggest we're doing the right thing for soft dependencies (though I remain unsure about the foreman default).

ekohl:
I'm not quite sure either, but `server` defaults to `false` so it's already non-default. Maybe we should set the default to false in the module and default it to true in the installer answer file.

domcleal:
@ekohl yeah, I was looking at that the other day, I think it'd be a good idea.

cristi1979:
Feel free to close the ticket whenever you want.

ekohl:
@cristi1979 got any hints on how we could have communicated this better? Would a mail to the ML have helped?

cristi1979:
No, it was just an expectation of mine. A lot of modules do this, so it's not something unique that needs to be announced.

domcleal:
I'll be sure to make it clear in the CHANGELOG when we release it too.

PR #237, opened 2015-02-10
Fixes puppet apply with strict_variables

sathieu:
On local, $::puppetmaster is not set, which errors out with strict variables.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -88,7 +88,11 @@ class puppet::params {
   $remove_lock               = true
 
   # Custom puppetmaster
-  $puppetmaster              = $::puppetmaster
+  if defined('$trusted') and $::trusted['authenticated'] == 'local' {
+    $puppetmaster            = undef
+  } else {
+    $puppetmaster            = $::puppetmaster
+  }
 
   # Hashes containing additional settings
   $additional_settings   =      {}


manifests/params.pp
```diff
@@ -74,7 +74,11 @@
   $agent                     = true
 
   # Custom puppetmaster
-  $puppetmaster              = $::puppetmaster
+  if $trusted['authenticated'] == 'local' {
```
ekohl: This should be ```$::trusted['authenticated']``` according to puppet-lint.

sathieu:
@ekohl Sorry, fixed.

manifests/params.pp
```diff
@@ -74,7 +74,11 @@
   $agent                     = true
 
   # Custom puppetmaster
-  $puppetmaster              = $::puppetmaster
+  if $::trusted['authenticated'] == 'local' {
```
ekohl: It appears the tests exploded because it's undefined. I think it needs a check if ```$::trusted``` is defined.

sathieu:
@ekohl PR updated

ekohl:
:+1:

ekohl:
BTW, we prefer a linear history so it'd be great if you could squash the commits. If not we'll just do it when merging.

manifests/params.pp
```diff
@@ -74,7 +74,11 @@
   $agent                     = true
 
   # Custom puppetmaster
-  $puppetmaster              = $::puppetmaster
+  if $::trusted and $::trusted['authenticated'] == 'local' {
```
domcleal: Here's a weird observation - if you enable strict_variables but not trusted_node_data then Puppet throws an error saying ::trusted is undefined.  It seems you need to do `if defined('$trusted') and $::trusted['authenticated'] == 'local'` instead. (The lack of "::" in the defined call is deliberate for it to work.)

sathieu: Yes. But how behave defined('$var') on old puppet?

sathieu:
squash done.

manifests/params.pp
domcleal: I think it's a very old function.

sathieu:
According to https://docs.puppetlabs.com/references/stable/function.html#defined, defined('$var') is since 3.6. But this would go to the else on previous versions.

so we have to choose between (context=puppet apply):
- fix with strict_variables=true and puppet >=3.6 (and error on strict variable if puppet <3.6), or
- fix with strict_variables=true and trusted_node_data=true (and error on not defined $trusted if  trusted_node_data=false)

domcleal:
Ah, interesting.  I think the combination of Puppet < 3.6 and strict_variables is perhaps less important.

ekohl:
I agree that strict_variables=true and trusted_node_data=false resulting in an error is fine.

domcleal:
Oh, I didn't express that clearly - I think (although don't have a strong opinion) that making strict_variables=true with trusted_node_data=false work, even if that's only on Puppet >= 3.6 is preferable to supporting strict_variables=true with trusted_node_data=true on Puppet 3.4 and 3.5.

mmoll:
:+1:

ekohl:
:+1:

mmoll:
Merged, merci @sathieu!

sathieu:
@mmol, @domcleal, @ekohl thanks for your carefull review ;-)

PR #238, opened 2015-02-11
Update passenger.pp

cristi1979:
fix for ssl_ca_crl=false


ekohl:
:+1: since setting ```$puppet::server_ca``` to false will cause ```ssl_ca_crl``` to be false.

domcleal:
Good catch, thanks @cristi1979!

PR #240, opened 2015-02-13
Extending support for additional settings

beddari:
Additional settings for [main] and [agent] config file sections

... possibly not quite ready yet ;) Let's see how the tests goes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,9 +12,9 @@ The Puppet master is configured under Apache and Passenger by default, unless
 to `puppetserver` to switch to the JVM-based Puppet Server.
 
 Many puppet.conf options for agents, masters and other are parameterized, with
-class documentation provided at the top of the manifests.
-`server_additional_settings` can be used to supply master options that are
-missing parameters.
+class documentation provided at the top of the manifests. In addition, there
+are hash parameters for each configuration section that can be used to supply
+any options that are not explicitly supported.
 
 ## Environments support
 
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -117,6 +117,9 @@
 #
 # $pluginfactsource::              URL to retrieve Puppet facts from during pluginsync
 #
+# $additional_settings::           A hash of additional main settings.
+#                                  type:hash
+#
 # == puppet::agent parameters
 #
 # $agent::                         Should a puppet agent be installed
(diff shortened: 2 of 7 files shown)


manifests/init.pp
```diff
@@ -140,6 +143,9 @@
 #
 # $service_name::                  The name of the puppet agent service.
 #
+# $agent_additional_settings::     A hash of additional settings.
```
domcleal: A hash of additional settings for "main"?

domcleal:
:+1:

README.md
```diff
@@ -13,7 +13,7 @@ to `puppetserver` to switch to the JVM-based Puppet Server.
 
 Many puppet.conf options for agents, masters and other are parameterized, with
 class documentation provided at the top of the manifests.
-`server_additional_settings` can be used to supply master options that are
+The `*additional_settings` parameters can be used to supply options that are
 missing parameters.
```
ekohl: Maybe add something like 'there is a parameter for each configuration section' or explicitly name them? It is that I know what's going on here, but as a new user I couldn't place this.

beddari: Ok I was just trying not to make a list ;) Explicit is best I think, let me try.

manifests/params.pp
```diff
@@ -76,6 +76,10 @@
   # Custom puppetmaster
   $puppetmaster              = $::puppetmaster
 
+  # Hashes containing additional settings
+  $additional_settings       = {}
```
ekohl: I wouldn't mind if this was explicitly ```main_additional_settings``` as well.

ekohl:
In general I think the parameters are a good addition. Just want them to be as clear as they can be.

manifests/params.pp
beddari: Well, yeah, was thinking the same, except that the master one is called 'server'. Looking at the semantics of the other main options I opted for just additional_settings. But main is fine ...

ekohl: There's an argument for both. @domcleal?

domcleal: I'm ambivalent...

ekohl: I think right now consistency with the others (so leave main) out is better. IIRC anything in main also counts for [master] and [agent] so that means it's really the generic additional settings. So leave it as it is now, but I reserve the right to change my opinion again ;)

beddari: My pattern is to abstract the config file structure in Puppet but map kv pairs in hashes. As such "main", "agent" and "master" would be the ideal user UX. I think if you want to go all the way, that would be overall the best approach. But beware, I'm in the crazy Hiera-model camp ;)

beddari:
* Updated docs
* Hash validation

ekohl:
:+1:

mmoll:
thanks @beddari!

PR #241, opened 2015-02-16
FreeBSD agent support

mmoll:
For the master configuration there's more stuff missing from puppetlabs-apache, so that's not going to happen for now.

The server_vardir parameter is gone, which is theoretically OK because the version is bumped anyway. puppet-foreman_proxy would need to be updated in one line where this is used as default for a puppet_home setting.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -35,7 +35,7 @@ class puppet::params {
   $syslogfacility      = undef
 
   case $::osfamily {
-    Windows : {
+    'Windows' : {
       # Windows prefixes normal paths with the Data Directory's path and leaves 'puppet' off the end
       $dir_prefix = 'C:/ProgramData/PuppetLabs/puppet'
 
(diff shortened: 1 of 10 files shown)


ekohl:
We also have the group set to root in puppet::server::env

ekohl:
Also, if you run the master using webrick, does it work then?

mmoll:
First comment addressed.
Regarding server support, with some hacks it works, however the whole /usr vs. /usr/local thing would need to be addressed more globally. If wanted, I can try to tackle that, but would prefer that to happen in a seperate PR.

ekohl:
This already existed, but I wonder if we should differentiate what level of support we have per OS in the README since agent runs on more than the server.

mmoll:
Maybe, but I think the non Debian/RH based user number will be fairly low anyway...
Now also included in the PR is support for DragonflyBSD 3.6+, as it's using pkgng packages also and it's working just fine.

ekohl:
:+1:

manifests/params.pp
```diff
@@ -122,7 +133,7 @@
   # Owner of the environments dir: for cases external service needs write
   # access to manage it.
   $server_environments_owner   = $user
-  $server_environments_group   = 'root'
+  $server_environments_group   = $root_group
```
domcleal: root_group isn't defined in the Windows OS family case... does it matter? :)

ekohl: For strict vars I think it does.

mmoll: @genebean what would be the best choice for that group on Windows? Is "Administrators" also working on non-English Windows variants?

ekohl: Maybe undef?

mmoll: "Administrators" seems a good choice according to https://support.microsoft.com/kb/243330

genebean: I would think undef because, generally speaking, you will want it to inherit permissions from it's parent folder. There are also some interesting issues with permissions on Windows via Puppet outlined at https://docs.puppetlabs.com/puppet/3.6/reference/resources_file_windows.html

genebean: FWIW, if you don't want to do `undef`, then I agree that `Administrators` would be the next best thing.

fraenki:
:+1: tested on FreeBSD 10.1 running puppet 3.6.2

manifests/params.pp
```diff
@@ -43,13 +43,26 @@
       $logdir = "${dir_prefix}/var/log"
       $rundir = "${dir_prefix}/var/run"
       $ssldir = '$confdir/ssl'
+      $vardir = "${dir_prefix}/var"
+      $root_group = 'Administrators'
```
ekohl: Per suggestion from @genebean I'd go for undef here.

mmoll: ok, changed

domcleal:
:+1:

domcleal:
Thanks @mmoll!

PR #242, opened 2015-02-19
modulesync: new lint checks, add CONTRIBUTING doc

domcleal:
Lots to fix.


ekohl:
Tests should now pass.

mmoll:
:+1:

manifests/server.pp
```diff
@@ -21,7 +21,7 @@
     if $::puppet::server_git_repo {
       $config_version_cmd = "git --git-dir ${::puppet::server_envs_dir}/\$environment/.git describe --all --long"
     } else {
-      $config_version_cmd = ''
+      $config_version_cmd = undef
```
domcleal: I'm slightly concerned about this, plus using `if scope.lookupvar(..)` in the template, as setting a variable to undef sometimes results in :undef symbols and things in ERB templates with lookupvar.  Should we make it a local variable in `puppet::server::config` to avoid lookupvar?

ekohl: I was considering the same and even wrote some code to play with it. I strongly considered adding all template parameters and get rid of ```scope.lookupvar```. Would that be something you agree on?

domcleal: :+1:

ekohl:
Rebased

domcleal:
Thanks very much @mmoll and @ekohl, great job!

PR #243, opened 2015-02-23
bump minor version

mmoll:
after changing (server_)vardir in 2c17dfa


domcleal:
3.0.0's fine, we haven't released it yet :)

mmoll:
ah, yay.

PR #244, opened 2015-02-24
Add remove_lock parameter

mcanevet:
We want to be able to get around puppet locking system using `--agent_disabled_lockfile /nonexisting`, but by default the puppet lock is removed on Debian.
This patch will allow us to set `remove_lock` to false.


ekohl:
:+1: pending green tests.

domcleal:
Thanks @mcanevet!

PR #245, opened 2015-02-25
Add new boolean parameter stringify_facts

martinpfeifer:
This parameters must be set in the [main] or [agent] section not in the
[master] section. That's why specifing this in the server_additional_settings
parameter will not work.

### DIFF
--- a/templates/puppet.conf.erb
+++ b/templates/puppet.conf.erb
@@ -64,3 +64,6 @@
 <% scope.lookupvar("puppet::additional_settings").sort_by {|k, v| k}.each do |key, value| -%>
     <%= key %> = <%= value %>
 <% end -%>
+<% if @stringify_facts == false -%>
+    stringify_facts = <%= @stringify_facts.to_s %>
+<% end -%>
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -122,6 +122,9 @@
 # $additional_settings::           A hash of additional main settings.
 #                                  type:hash
 #
+# $stringify_facts::               Whether to convert facts to strings. Defaults to true.
+#                                  type: boolean
+#
 # == puppet::agent parameters
 #
 # $agent::                         Should a puppet agent be installed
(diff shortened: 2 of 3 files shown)


templates/puppet.conf.erb
```diff
@@ -64,3 +64,6 @@
 <% scope.lookupvar("puppet::additional_settings").sort_by {|k, v| k}.each do |key, value| -%>
     <%= key %> = <%= value %>
 <% end -%>
+<% if @stringify_facts == false -%>
```
ekohl: Why only when it's false?

ekohl:
We also have ```additional_settings``` for ```[main]``` and ```agent_additional_settings``` for ```[agent]``` precisely for this use case. I'm fine with this change though, especially since it fixes an incorrect example.

templates/puppet.conf.erb
martinpfeifer: The puppet internal default is currently true (see https://docs.puppetlabs.com/references/latest/configuration.html#stringifyfacts). I thought if the parameter doesn't deviate from the default (set to true or undef) i don't set it in the config at all.

martinpfeifer:
I added this a while ago locally but didn't push to GitHub. In the meantime https://github.com/FILIADATAGmbH/puppet-puppet/commit/4cc6629697c111b73aee628f41158f0e7db81bbb added what caused my changes initially.
Not sure if its better to have this as explicit parameter or handle it via the generic additional settings interface.

ekohl:
I think that mostly depends on how common it is to set the parameter and I can't really judge about this one.

templates/puppet.conf.erb
mmoll: I think this is OK, as the default is ```true``` now and the whole option will be gone in Puppet 4 by puppetlabs/puppet#3212

mmoll:
A test would be nice :slot_machine:

martinpfeifer:
I now think the generic interface is sufficient to configure this setting as it's probably not that common to configure it at all. I'm closing this pull request and create a new one that fixes only the documentation.

PR #246, opened 2015-03-03
Add parser option to [master] section of puppet.conf

danfoster:
This PR adds the `parser` option to the `[master]` section of `puppet.conf` allowing you to specify the parser puppet should use.
It defaults to `parser = current`, which is the default if this option is left out, see https://docs.puppetlabs.com/references/latest/configuration.html#parser .

This allows module users to override the `::puppet::server_parser` to use the future parser.

### DIFF
--- a/spec/classes/puppet_server_config_spec.rb
+++ b/spec/classes/puppet_server_config_spec.rb
@@ -356,4 +356,18 @@ describe 'puppet::server::config' do
       end
     end
   end
+
+  describe 'with server_parser => future' do
+    let :pre_condition do
+      "class {'puppet':
+        server => true,
+        server_parser => 'future',
+      }"
+    end
+
+    it 'should configure future parser' do
+      should contain_concat_fragment('puppet.conf+30-master').
+        with_content(/^\s+parser\s+=\s+future$/)
+    end
+  end
 end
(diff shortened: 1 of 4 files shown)


ekohl:
Does puppet < 3.2 ignore the setting or error out?

domcleal:
Generally I think Puppet ignores unknown settings, as we have a number in our templates which are recent additions to Puppet.

ekohl:
If it works on puppet < 3.2, :+1: from me.

mmoll:
A small test for the setting actually set to future would be nice, also the variable could be checked with validate_re. But I'm visually fine with it also as-is. :+1:

spec/classes/puppet_server_config_spec.rb
```diff
@@ -89,6 +89,7 @@
         with_content(/^\s+node_terminus\s+= exec$/).
         with_content(/^\s+ca\s+= true$/).
         with_content(/^\s+ssldir\s+= \/var\/lib\/puppet\/ssl$/).
+        with_content(/^\s+parser\s+=/).
```
domcleal: `=\s+current$/`

domcleal:
Tested on Puppet 2.7.26, works fine!  If you can fix the above test, add validate_re and optionally add a test, that would be great.

danfoster:
Thanks for the feedback and testing on older versions of puppet.

I believe I've addressed all 3 suggestions.

mmoll:
:+1:

domcleal:
Thanks @danfoster!  Merged as 864e175.

PR #248, opened 2015-03-06
Anchor Foreman puppet master configuration inside server::config

domcleal:
Pre-req to fixing theforeman/puppet-foreman#129 and removing ordering and
implicit dependency of foreman::puppetmaster on theforeman/puppet-puppet.


ekohl:
By using anchors, will it automatically be constrained to this class and thus automatically require install and notify the service? If so :+1:

domcleal:
@ekohl yep, as this config class is constrained too

ekohl:
Thanks! Who'd have thought it was such a small change :)

PR #249, opened 2015-03-09
Use puppet reload instead of puppet restart

sathieu:
This avoid killing the current puppet agent run

### DIFF
--- a/manifests/agent/service.pp
+++ b/manifests/agent/service.pp
@@ -4,10 +4,12 @@ class puppet::agent::service {
   case $::puppet::runmode {
     'service': {
       service {'puppet':
-        ensure    => running,
-        name      => $puppet::service_name,
-        hasstatus => true,
-        enable    => true,
+        ensure     => running,
+        name       => $puppet::service_name,
+        hasstatus  => true,
+        hasrestart => $puppet::agent_restart_command != undef,
+        enable     => true,
+        restart    => $puppet::agent_restart_command,
       }
 
       if $::osfamily != 'windows' {
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -191,6 +191,11 @@ class puppet::params {
 
   # Puppet service name
   $service_name = 'puppet'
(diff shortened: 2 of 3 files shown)


sathieu:
Tested on Debian with and without puppetlabs repo. Tested on CentOS6 with puppetlabs repos

mmoll:
At least FreeBSD has no "reload" for the puppet service, but I get the problem...

manifests/agent/service.pp
```diff
@@ -4,10 +4,12 @@
...
-        ensure    => running,
-        name      => $puppet::service_name,
-        hasstatus => true,
-        enable    => true,
+        ensure     => running,
+        name       => $puppet::service_name,
+        hasstatus  => true,
+        hasrestart => true,
+        enable     => true,
+        restart    => '/usr/sbin/service puppet reload',
```
ekohl: I don't think this works on windows. Agent has a decent support matrix so maybe this should be in params.pp and undef for most OS-es.

sathieu:
Here's an updated patch.

manifests/params.pp
```diff
@@ -191,6 +191,11 @@
 
   # Puppet service name
   $service_name = 'puppet'
+  $agent_restart_command = $::osfamily {
```
mmoll: missing case? :)

manifests/agent/service.pp
```diff
@@ -4,10 +4,12 @@
   case $::puppet::runmode {
     'service': {
       service {'puppet':
-        ensure    => running,
-        name      => $puppet::service_name,
-        hasstatus => true,
-        enable    => true,
+        ensure     => running,
+        name       => $puppet::service_name,
+        hasstatus  => true,
```
ekohl: I suppose we can drop this now we only support puppet 2.7.0+

mmoll:
works for me on FreeBSD

ekohl:
Code looks good, :+1:

mmoll:
merged as 8846c84, thanks @sathieu!

PR #250, opened 2015-03-11
modulesync: enable future parser tests

domcleal:


mmoll:
rebased.

mmoll:
Fixed some tests, maybe somebody else could have a look onto the apache/rack thing still failing?

There are quite some module parts untested, so even if all the tests suceed, the module will not really be future_parser ready.

However, I think we should just bring that in so at least new PRs with tests will have an indicator about it.

ekohl:
I don't know if this change works, but let's see what Travis says.

ekohl:
Looks like ```$puppet::server_app_root``` isn't set. This can be solved through a pre_condition in the tests.

mmoll:
Solved this by using rspec-puppet-facts also there, should turn :green_heart:

mmoll:
mispushed there :(
but the PR tests are green.

mmoll:
rebased.

domcleal:
:+1: from me.

ekohl:
Minor comments inline. :+1: for the general change.

domcleal:
Fixed the indentation and removed the check for EL6, it now runs on all OSes.

domcleal:
And fails on all OSes due to the app_root being missing, curious.  Will investigate more soon.

I don't think the conditional here was working reliably (maybe r-p-f version differences): https://github.com/domcleal/puppet-puppet/commit/472b6002cf348e80ef8919026befddee456ec546#diff-2192e49e7ea16301b4ba87a40eeb3a86R5 as the previous run (https://travis-ci.org/domcleal/puppet-puppet/jobs/64176632) doesn't seem to have run any tests from that class.

ekohl:
That is the same error we saw as before the if check. I wonder if the tests were even executed or we just masked it.

ekohl:
I still think the spec needs a pre-condition which includes the main puppet class so the variable is set.

domcleal:
> I still think the spec needs a pre-condition which includes the main puppet class so the variable is set.

I agree that'd work.  I took a slightly different route, which was to parameterise puppet::server::rack and explicitly initialise it, is that OK?

ekohl:
:+1:

mmoll:
Thank You, Gentlemen!
:clap:

PR #251, opened 2015-03-12
fix documentation regarding additional parameters

martinpfeifer:
As mentioned in https://github.com/theforeman/puppet-puppet/pull/245 here the fixed documentation.


ekohl:
Thanks!

PR #253, opened 2015-03-21
Add options to let the puppet master listen for HTTP connections too

whefter:
(load balancer/reverse proxy scenarios), protected by default Deny all. Configurable listen port and allow directive.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -171,6 +171,19 @@
 # $server_ca::                     Provide puppet CA
 #                                  type:boolean
 #
+# $server_http::                   Should the puppet master listen on HTTP as well as HTTPS.
+#                                  Useful for load balancer or reverse proxy scenarios. Note that
+#                                  the HTTP puppet master denies access from all clients by default,
+#                                  allowed clients must be specified with $server_http_allow.
+#                                  type:boolean
+#
+# $server_http_port::              Puppet master HTTP port; defaults to 8139.
+#                                  type:integer
+#
+# $server_http_allow::             Array of allowed clients for the HTTP puppet master. Passed
+#                                  to Apache's 'Allow' directive.
+#                                  type:array
+#
 # $server_reports::                List of report types to include on the puppetmaster
 #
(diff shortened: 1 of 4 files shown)


manifests/init.pp
```diff
@@ -171,6 +171,19 @@
...
+#                                  the HTTP puppet master denies access from all clients by default,
+#                                  allowed clients must be specified with $server_http_allow.
+#                                  type:boolean
+#
+# $server_http_port::              Puppet master HTTP port; defaults to 8139.
+#                                  type:integer
+#
+# $server_http_allow::             Array of allowed clients for the HTTP puppet master. Passed
+#                                  to Apache's 'Allow' directive.
+#                                  type:integer
```
mmoll: should probably be type:array

mmoll:
Tested, works for me. :+1: 
@domcleal? @ekohl?

whefter:
Oops, yeah. Corrected that, and rebased/squashed the commits into one.

manifests/server/passenger.pp
```diff
@@ -87,4 +92,31 @@
...
+            default => $custom_fragment
+          },
+          'SetEnvIf X-Client-Verify "(.*)" SSL_CLIENT_VERIFY=$1',
+          'SetEnvIf X-SSL-Client-DN "(.*)" SSL_CLIENT_S_DN=$1',
+        ], "\n"),
+      options         => ['None'],
+      require         => Class['::puppet::server::rack'],
+    }
+  }
+}
```
ekohl: nitpick: I'd prefer a newline at the end of the file.

ekohl:
I'm unsure. My main concern is that beginners might try this. I do see the use case so maybe it could be documented in the README as an advanced scenario?

mmoll:
More documentation is always good :memo:

mmoll:
:+1:

ekohl:
:+1: but I'd like @domcleal to take a look as well.

mmoll:
merged as also @domcleal's feedback via IRC was positive, thanks @whefter! :tada:

PR #256, opened 2015-04-13
Bugfix: duplicate declaration error when combining puppetserver together...

visibilityspots:
... with puppetdb module

https://github.com/puppetlabs/puppetlabs-puppetdb#puppet_service_name

When using the puppetlabs-puppetdb module to configure puppetdb in combination with puppetserver implemention I got duplicate resource errors on the puppetserver service.

By simply issuing the if ! defined check this could be tackled on a rather proper way I suppose?

### DIFF
--- a/manifests/server/service.pp
+++ b/manifests/server/service.pp
@@ -10,7 +10,7 @@
 # $puppetserver::  Whether to start/stop the (JVM) puppetserver service
 #                  type:boolean
 #
-class puppet::server::service(
+class puppet::server::service (
   $puppetmaster = undef,
   $puppetserver = undef,
 ) {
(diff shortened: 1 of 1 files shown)


ekohl:
I don't think this is a good idea because now you have two places where it could be managed. While I haven't checked recently, there are switched for this in puppetlabs-puppetdb. We should document them, but they are in [puppetdb::master::config](https://github.com/puppetlabs/puppetlabs-puppetdb/blob/master/manifests/master/config.pp). I believe you want ```manage_storeconfigs```, ```manage_report_processor``` and ```restart_puppet``` set to ```false```.

visibilityspots:
I totally over looked those parameters! In that case you are right indeed this change isn't relevant any more.

@ekohl: thanks for pointing me in the right direction!

PR #258, opened 2015-04-16
missing depency on the foreman module

attachmentgenie:


ekohl:
This dependency was explicitly removed in 5d9f9f401b14fa85d0edf7826413a092fbf892a6 because it's an optional dependency. See https://docs.puppetlabs.com/guides/style_guide.html#dependencies as well.

PR #259, opened 2015-04-19
Fix all hosts allowed in HTTP puppetmaster

whefter:
So, slightly embarrassing, but due to the way the puppetlabs/apache module handles Order, Allow and Deny between versions 2.2/2.4, the fact that when you specify no allowed hosts, all hosts are allowed somehow slipped through. This PR should fix it by actually forcing the puppetlabs/apache module to change the Order directive to Deny, Allow by default, then adding it again for good measure, denying all again and THEN adding allowed hosts again.

Slightly convoluted, but the best I can do with the current Apache 2.2/2.4 Order/Allow/Deny thing.

### DIFF
--- a/manifests/server/passenger.pp
+++ b/manifests/server/passenger.pp
@@ -93,8 +93,15 @@ class puppet::server::passenger (
   }
 
   if $http {
+    # Order, deny and allow cannot be configured for Apache >= 2.4 using the Puppetlabs/Apache
+    # module, but they can be set to false. So, set to false and configure manually via custom fragments.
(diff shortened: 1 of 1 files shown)


ekohl:
The tests are failing because puppetlabs-concat has a new dependency. Otherwise I think this change is fine.

mmoll:
Tested on Debian/wheezy (Apache 2.2), the diff of the generated files seems ok to me :+1:

ekohl:
Thanks @whefter!

PR #263, opened 2015-05-19
Use puppetlabs-concat instead of theforeman-concat_native

mcanevet:
This works with either puppetlabs-concat 1.2.x or 2.x so can be merge without waiting for dependencies to be compatible with puppetlabs-concat 2.x IMHO.

### DIFF
--- a/spec/classes/puppet_agent_install_spec.rb
+++ b/spec/classes/puppet_agent_install_spec.rb
@@ -10,6 +10,7 @@ describe 'puppet::agent::install' do
     context "on a RedHat family OS" do
       let :facts do {
         :osfamily => 'RedHat',
+        :concat_basedir => '/foo/bar',
       } end
 
       it 'should not define provider' do
@@ -20,6 +21,7 @@ describe 'puppet::agent::install' do
     context "on a Windows family OS" do
       let :facts do {
         :osfamily => 'windows',
+        :concat_basedir => 'C:\Temp',
       } end
 
       it 'should define provider as chocolatey' do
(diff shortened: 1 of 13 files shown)


ekohl:
I didn't test it, but :+1: code wise (and I trust you tested it). I do think this would be a major version bump since the concat_native fragments were part of the API and we have recommend users to use it for full flexibility.

spec/classes/puppet_agent_install_spec.rb
```diff
@@ -37,6 +39,7 @@
 
     let :facts do {
       :osfamily => 'windows',
+      :concat_basedir => '/foo/bar',
```
mmoll: would this work on a real Windows system? :)

mmoll:
:+1: from me also, please bump the major version of this module and have a short look at the windows concat_basedir, I know it makes no difference in these kind of tests but I 'd like to have it close to reality. I'm actually a bit surprised this isn't catched in concat itself, but maybe a UNIX path is somehow converted internally.

mcanevet:
@mmoll I fixed the `concat_basedir` when using windows.
Do you want me to bump version in the `metadata.json` in this PR, or will you do it latter before releasing?

ekohl:
I think it's fine to leave bumping till later.

mcanevet:
@ekohl @mmoll @domcleal so? OK to merge this?

mmoll:
works for me :+1:

domcleal:
Thanks @mcanevet!

PR #264, opened 2015-05-19
Fix incomplete param for environment_timeout

max06:
The default of this param changed in puppet 3.7.x from 3 minutes to no caching. Running this configuration in a bigger system causes performance issues.

You can now configure the timeout in the installer, if needed. Defaults to disabled.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -333,6 +333,8 @@
 #                                  catalog from Foreman (in seconds).
 #                                  type:integer
 #
+# $server_environment_timeout::    Timeout for cached compiled catalogs (10s, 5m, ...)
+#
 # $server_ca_proxy::               The actual server that handles puppet CA.
 #                                  Setting this to anything non-empty causes
 #                                  the apache vhost to set up a proxy for all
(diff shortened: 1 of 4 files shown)


mmoll:
:+1:
Please squash your commits together into one.

A nice thing would be to use ```validate_re``` on that variable. Besides a timeframe also ```0``` and ```unlimited``` are allowed, according to the docs.

mmoll:
:+1:

ekohl:
I think it looks good, but it would be awesome if you could add a test where you do set the value.

max06:
Sorry, noob here - I'm a sysadmin, not a developer :D 

(I tested it manually in my test setup, worked so far)

mmoll:
@max06 no worries, we all started like that :)
see 864e175f for an example, how to test such a setting.

max06:
Need more time for that, doing this besides my normal job like others. And I never did any kind of unit testing before, so have to read a lot first :)

ekohl:
I don't mind merging it and one of us writing the test. @mmoll agreed?

mmoll:
sure

max06:
Would love to see the written tests for learning :)

mmoll:
mmoll/puppet-puppet@a7e0b7b is a very simple test for this setting. @ekohl if that test is OK for you, just ACK here and I'd push this two commits into master.

ekohl:
@mmoll :+1:

mmoll:
Merged as f65898e, thanks @max06!

PR #265, opened 2015-05-19
Fixed $agent_restart_command to fit Redhat 6 & 7

Thubo:
Changes:
  * Replaced selector between debian and Redhat to also distinguish
    between Redhat 6 and Redhat 7

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -195,10 +195,22 @@ class puppet::params {
 
   # Puppet service name
   $service_name = 'puppet'
-  $agent_restart_command = $::osfamily ? {
-    'Debian' => '/usr/sbin/service puppet reload',
-    'Redhat' => '/usr/sbin/service puppet reload',
-    default  => undef,
+  # Command to restart the agent
+  case $::osfamily {
+    'Debian' : { $agent_restart_command = "/usr/sbin/service ${service_name} reload" }
+    'Redhat' : {
+               case versioncmp( $::operatingsystemrelease, 7.0) {
+                 '1','0' : {
+                   $agent_restart_command = "/usr/bin/systemctl restart ${service_name}" # operatingsystemrelease is equal or greater than 7.0
+                 }
+                 '-1' : {
+                   $agent_restart_command = "/sbin/service ${service_name} reload"       # operatingsystemrelease is less than 7.0
+                 }
+               }
+    }
+    default  : {
+      $agent_restart_command = undef
+    }
   }
 
   # Foreman parameters
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -197,16 +197,15 @@ class puppet::params {
   $service_name = 'puppet'
   # Command to restart the agent
   case $::osfamily {
-    'Debian' : { $agent_restart_command = "/usr/sbin/service ${service_name} reload" }
+    'Debian' : {
+      $agent_restart_command = "/usr/sbin/service ${service_name} reload"
+    }
     'Redhat' : {
-               case versioncmp( $::operatingsystemrelease, 7.0) {
-                 '1','0' : {
-                   $agent_restart_command = "/usr/bin/systemctl restart ${service_name}" # operatingsystemrelease is equal or greater than 7.0
-                 }
-                 '-1' : {
-                   $agent_restart_command = "/sbin/service ${service_name} reload"       # operatingsystemrelease is less than 7.0
-                 }
-               }
+      $osreleasemajor = regsubst($::operatingsystemrelease, '^(\d+)\..*$', '\1') # workaround for the possibly missing operatingsystemmajrelease
+      $agent_restart_command = $osreleasemajor ? {
+        '7'     => "/usr/bin/systemctl restart ${service_name}",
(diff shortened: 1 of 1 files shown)


Thubo:
Right now I have no Redhat 7 at hand to double check, but at my Redhat 6 instances the service binary is located at /sbin/service, so I'm guessing it is at /usr/sbin/service for Redhat 7...

ekohl:
I think EL7 should use ```systemctl``` instead.

domcleal:
/sbin/service would work on both, as EL7 is post-UsrMove, so /sbin is a symlink to /usr/sbin.  But yeah, systemctl might be better.

Thubo:
Do you guys fix that or do you want me to send another pull request (however that might have to wait until tomorrow or so...)?

domcleal:
Would you mind fixing it?  We're in no rush, so at your pace!

Please just amend this existing pull request by changing the file, running `git add file.pp`, `git commit --amend` and then `git push -f` to update this branch - it'll update the PR automatically.

manifests/params.pp
```diff
@@ -195,11 +195,14 @@
 
   # Puppet service name
   $service_name = 'puppet'
-  $agent_restart_command = $::osfamily ? {
-    'Debian' => '/usr/sbin/service puppet reload',
-    'Redhat' => '/usr/sbin/service puppet reload',
-    default  => undef,
+  # Command to restart the agent
+  if $::osfamily == 'Debian' { $agent_restart_command = "/usr/sbin/service $service_name reload" }
```
domcleal: Note that tests are failing as `$service_name` should use brackets, e.g. `${service_name}`

Thubo:
A college had a RHEL 7 VM around ;) I also fixed the ${service_name}. I hope it's ok now, if not let me know :)

mmoll:
I have two small points:
- please use ```versioncmp``` from stdlib on ```$::operatingsystemrelease```, as ```$::operatingsystemmajrelease``` is (depending on the facter version) not a builtin fact.
- please use ```case``` logic (nested, where needed) instead of the ```if``` constructions.

Once you got the logic right, I'll probably have some formatting nits. ;)

Thubo:
New try ;) I hope I use the `versioncmp` the way it's meant to be used...

PS. Why case statements? I feel they are a lot less readable if nested.

mmoll:
Sorry, was in a rush when doing the first quick review, so sending you on the versioncmp track was wrong here. :bow: 

We use in other modules ```$osreleasemajor = regsubst($::operatingsystemrelease, '^(\d+)\..*$', '\1')``` to work around the possibly missing ```operatingsystemmajrelease``` fact and use it like you did it before.

Regarding the use of ```case```, many module authors (including me) prefer this often in selects bigger than just one ```if``` and ```else```. Also you're getting a ```default``` case for free and it's much more readable and extendable when having more options instead of using ```elsif``` all over again.

_Maybe_ it makes sense to move this selection also into the ```case``` for osfamily in line 37.

Also, regardless if you move the selction there or not, please use two space indentation and line breaks like in that osfamily case block.

mmoll:
also, please rebase.

manifests/params.pp
```diff
@@ -195,10 +195,21 @@
...
-    default  => undef,
+  # Command to restart the agent
+  case $::osfamily {
+    'Debian' : {
+      $agent_restart_command = "/usr/sbin/service ${service_name} reload"
+    }
+    'Redhat' : {
+      $osreleasemajor = regsubst($::operatingsystemrelease, '^(\d+)\..*$', '\1') # workaround for the possibly missing operatingsystemmajrelease
+      $agent_restart_command = $osreleasemajor ? {
+        '7'     => "/usr/bin/systemctl restart ${service_name}",
```
ekohl: Why restart and not reload?

Thubo: In my test RHEL 7 the reload is not defined in the systemd script, therefore it has to be restart.

Why the variable is called agent_restart_command ? idk ;)

mmoll: The original change was to use the "softer" reload in order to avoid killing the agent while configuring puppet itself.

ekohl: There's also ```systemctl reload-or-restart``` which I've never used but sounds relevant :)

mmoll:
any news here?

Thubo:
Quoting @mmoll: "any news here?" :)

mmoll:
I hoped my comment about the restart command calling reload was clearing up the situation...

If RH 7 really does not have this option, it must set to ```undef``` also there, so a normal restart will be done.

lvlie:
This module breaks badly on rhel 6 because /usr/sbin/service does not exist. +1

Thubo:
I decided to go with `reload-or-restart` on RH7, since for now it restarts the command, however if in the future a `reload` operation should be included in the puppet service it will be used.

mmoll:
I'm visually OK with that, @lvlie, could you test?

mmoll:
anybody?

mmoll:
merged, thanks @Thubo!

PR #268, opened 2015-05-21
Update params.pp to work on RedHat

HoogWater:
on RedHat the service command is in /sbin
changed the agent_restart_command to /sbin/service to get it working on the RedHat osfamily


ekohl:
https://github.com/theforeman/puppet-puppet/pull/265 is a more complete solution to this problem so closing this. Thanks for your contribution through!

PR #269, opened 2015-05-25
Add environment parameter

mcanevet:
Currently the environment set in puppet.conf comes from the top scope
variable. If we launch puppet with '--environment foo' for testing, it
changes the puppet.conf file (what we do no want). We used to override
$environment somewhere in our stack with the fixed environment so that
it work as expected thanks to dynamic scoping still functionnal in ERB
templates. But this workaround does not work anymore with the future
parser now they remove totally dynamic scoping.

This patch allows to set the environment in the puppet class.


raphink:
:+1: when they merge the https://github.com/camptocamp/puppet-osbase/issues/86

mmoll:
looks sensible, so :+1:

ekohl:
Thanks!

PR #270, opened 2015-05-25
Fix #262

### LINKED ISSUE #262: Impossible not to use ENC with future parser
I don't seem to be able to set `$::puppet::server_external_nodes` to a value that uses the plain terminus instead of exec. The `puppet` class checks that this value is a string, so I can't pass in `false`, but all strings in the future parser evaluate to true. Passing `undef` to `puppet` means the default param is used.

Could parameters that take a boolean be a bit more lenient about allowing actual booleans as well as strings that look like booleans?

mcanevet:


mmoll:
:+1:

ekohl:
Merged with a modified commit message. Thanks!

PR #272, opened 2015-05-27
Set mode/owner/group on common module directories

domcleal:
Ensure the mode/owner/group set by server_environments_* parameters is
applied to common module directories.


ekohl:
:+1: on green tests.

mmoll:
merged in #273 as 6c26182, thanks @domcleal!

PR #273, opened 2015-05-27
Set exec umask on Puppet 3.4+ for cert generation

domcleal:
Sorry about the refactor in the middle, the file was a bit of a mess - but I can remove it if needed.

The umask ensures the cert generation works properly under a restrictive umask, as the Puppet CA doesn't enforce file permissions.  The agent does though, which means running the agent during/after installation can make it work as it tidies some file permissions.


ekohl:
:+1: on green tests

mmoll:
thanks @domcleal!

PR #276, opened 2015-06-17
Classfile should point to $statedir/classes.txt by default.

AlexRRR:
Hi! recently I noticed that the classfile parameter is pointing to the wrong place:

The documentation states that it should be set by default to ```$statedir/classes.txt``` 

I ran into a weird problem where MCollective could not find node based on classes because it was looking at the default file path, while this module sets it to a really old default.

The Mcollective documentation (fro puppetlab) states:
> Puppet provides a list of classes applied to a node by default in /var/puppet/state/classes.txt or  /var/lib/puppet/state/classes.txt (depending on which Puppet version you are using. The latter is true for 0.25.5 onwards) , we’ll use this data with –with-class filters.

Thank you!

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -1,138 +1,138 @@
 # Default parameters
 class puppet::params {
 
-  include foreman::params
+    include foreman::params
 
-  # Basic config
-  $version             = 'present'
-  $user                = 'puppet'
-  $group               = 'puppet'
-  $dir                 = '/etc/puppet'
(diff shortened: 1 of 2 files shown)


ekohl:
I think your PR needs a serious rebase. Other than that it sounds sane.

AlexRRR:
ugh. sorry about that, will try with a different pull, too lazy to rebase.

PR #277, opened 2015-06-17
classfile should default to /classes.txt

AlexRRR:
Hi! recently I noticed that the classfile parameter is pointing to the wrong place:

The documentation states that it should be set by default to ```$statedir/classes.txt``` 

I ran into a weird problem where MCollective could not find node based on classes because it was looking at the default file path, while this module sets it to a really old default.

The Mcollective documentation (from puppetlab) states:
> Puppet provides a list of classes applied to a node by default in /var/puppet/state/classes.txt or  /var/lib/puppet/state/classes.txt (depending on which Puppet version you are using. The latter is true for 0.25.5 onwards) , we’ll use this data with –with-class filters.

Thank you!

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -30,7 +30,7 @@ class puppet::params {
   $pluginsource        = 'puppet:///plugins'
   $pluginfactsource    = 'puppet:///pluginfacts'
   # lint:endignore
-  $classfile           = '$vardir/classes.txt'
+  $classfile           = '$statedir/classes.txt'
   $hiera_config        = '$confdir/hiera.yaml'
   $syslogfacility      = undef
   $environment         = $::environment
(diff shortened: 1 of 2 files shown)


ekohl:
:+1: on green tests.

mmoll:
this probably needs also to change the exptected test results :)

AlexRRR:
Ok, I am lost now. I changed the only place where it was obvious that needed replacing and still the tests fail. Also, if I try to `rake spec` on my master (no changes there) it still fails. Some guidance would be most welcome.

ekohl:
I think the test failures are unrelated to your changes. I submitted https://github.com/theforeman/puppet-puppet/pull/278 and then the tests passed for me.

AlexRRR:
So I will wait for #278 to be merged, thanks!

mmoll:
#278 has been merged, plase rebase onto current master.

AlexRRR:
Thanks @mmoll, @ekohl it finally passed!

ekohl:
Would you mind squashing it into a single commit or do you want me to do it?

mmoll:
@AlexRRR ping?

ekohl:
Thanks!

PR #279, opened 2015-06-25
don't set deprecated configtimeout on 4.0+

mmckinst:
configtimeout was deprecated in 4.0 and will issue a warning if its set. It is set by default in the current template, when puppet 4 users use a puppet.conf generated from theforeman-puppet they'll see a warning like the one below

```
Warning: Setting configtimeout is deprecated. 
   (at /opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/settings.rb:1102:in `issue_deprecation_warning')
```

This pull requests stops it from being set by default on puppet 4 boxes.

See https://docs.puppetlabs.com/puppet/4.1/reference/deprecated_settings.html#configtimeout and https://tickets.puppetlabs.com/browse/PUP-3666 for more details.

### DIFF
--- a/templates/agent/puppet.conf.erb
+++ b/templates/agent/puppet.conf.erb
@@ -31,7 +31,9 @@
     splaylimit        = <%= scope.lookupvar('::puppet::splaylimit') %>
     runinterval       = <%= scope.lookupvar('::puppet::runinterval') %>
     noop              = <%= scope.lookupvar('::puppet::agent_noop') %>
+<% unless [nil, :undefined, :undef].include?(scope.lookupvar('::puppet::configtimeout')) -%>
     configtimeout     = <%= scope.lookupvar('::puppet::configtimeout') %>
+<% end -%>
     usecacheonfailure = <%= scope.lookupvar('::puppet::usecacheonfailure') %>
 <% unless [nil, :undefined, :undef].include?(scope.lookupvar('::puppet::prerun_command')) -%>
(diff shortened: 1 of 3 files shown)


templates/agent/puppet.conf.erb
```diff
@@ -31,7 +31,9 @@
     splaylimit        = <%= scope.lookupvar('::puppet::splaylimit') %>
     runinterval       = <%= scope.lookupvar('::puppet::runinterval') %>
     noop              = <%= scope.lookupvar('::puppet::agent_noop') %>
-    configtimeout     = <%= scope.lookupvar('::puppet::configtimeout') %>
+<% if @configtimeout and !@configtimeout.empty? -%>
```
domcleal: I think use of `@configtimeout` rather than a scope.lookupvar's causing the test failure, since configtimeout may not be in the template scope.

mmckinst:
I'll take a look and get the tests to pass this weekend.

mmckinst:
@domcleal I fixed it and updated the test to check for the existence/absence of configtimeout.

ekohl:
Looks good. Would you mind squashing your commits?

mmckinst:
@ekohl I've squashed the commits. Looks like the build, with the new updated pull request is passing too.

mmoll:
:+1:

ekohl:
Thanks!

PR #282, opened 2015-07-10
[WIP] Add puppet-agent AIO support

mcanevet:
Don't merge this yet, I'm just posting this for comments.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -45,6 +45,7 @@ class puppet::params {
       $dir_prefix = 'C:/ProgramData/PuppetLabs/puppet'
 
       $dir    = "${dir_prefix}/etc"
+      $codedir = "${dir_prefix}/etc"
       $logdir = "${dir_prefix}/var/log"
       $rundir = "${dir_prefix}/var/run"
       $ssldir = "${dir_prefix}/etc/ssl"
@@ -54,6 +55,7 @@ class puppet::params {
 
     /^(FreeBSD|DragonFly)$/ : {
       $dir        = '/usr/local/etc/puppet'
+      $codedir    = '/usr/local/etc/puppet'
       $logdir     = '/var/log/puppet'
       $rundir     = '/var/run/puppet'
       $ssldir     = '/var/puppet/ssl'
@@ -62,11 +64,21 @@ class puppet::params {
     }
 
     default : {
-      $dir        = '/etc/puppet'
-      $logdir     = '/var/log/puppet'
-      $rundir     = '/var/run/puppet'
-      $ssldir     = '/var/lib/puppet/ssl'
-      $vardir     = '/var/lib/puppet'
+      if versioncmp($::puppetversion, '4.0') < 0 {
+        $dir        = '/etc/puppet'
+        $codedir    = '/etc/puppet'
+        $logdir     = '/var/log/puppet'
+        $rundir     = '/var/run/puppet'
+        $ssldir     = '/var/lib/puppet/ssl'
+        $vardir     = '/var/lib/puppet'
+      } else {
+        $dir     = '/etc/puppetlabs/puppet'
+        $codedir = '/etc/puppetlabs/code'
+        $logdir  = '/var/log/puppetlabs/puppet'
+        $rundir  = '/var/run/puppetlabs'
+        $ssldir  = '/etc/puppetlabs/puppet/ssl'
+        $vardir  = '/opt/puppetlabs/puppet/cache'
+      }
       $root_group = undef
     }
   }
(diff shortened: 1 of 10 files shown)


ekohl:
Looks like a good start.

mmoll:
First, thanks for working on this! :+1: :bow:

Not a big thing, but for the comment: Can it somehow get detected with non-Enterprise AIO packages that it's an AIO package? If Debian or EPEL will package Puppet 4.x in the old fashioned way (I hope they do that), just comparing the version isn't the best selector.

domcleal:
> If Debian or EPEL will package Puppet 4.x in the old fashioned way (I hope they do that),

Fedora have been doing this.

mcanevet:
@mmoll @domcleal I'mhappy to learn that Fedora is packaging puppet4 the traditional way and I really hope that Debian will do the same as I really don't like the idea of a single AIO package... However I can't find a way to automagically know if we are using puppet4 AIO or traditional package...

domcleal:
Hm yeah, I can't think of a way either.  Thought there might be a fact indicating some agent confdir directory or similar, but I don't see one.

mcanevet:
@domcleal I wrote this this week end : https://github.com/camptocamp/puppet-puppet_config_providers
I think we can replace the whole puppet.conf management from an ERB template to this provider (I just have to write an augeas provider to manage pre-puppet 3.5). My idea would be to replace any default values in params.pp with `undef` so that we apply `puppet_config` resources only when we diverge from default. This new Puppet Type as the advantage of knowing where the puppet.conf file is located regardless of traditional, AIO, PE, gem packaging. And thus we could greatly simplify this module.

ekohl:
@mcanevet but we need the SSL dir at least because we use the SSL certs in other places as well. That means we at least need a fact for that.

manifests/params.pp
```diff
@@ -62,11 +62,20 @@
     }
 
     default : {
-      $dir        = '/etc/puppet'
-      $logdir     = '/var/log/puppet'
-      $rundir     = '/var/run/puppet'
-      $ssldir     = '$vardir/ssl'
-      $vardir     = '/var/lib/puppet'
+      if versioncmp($::puppetversion, '4.0') < 0 {
```
ekohl: Maybe just set a $codedir here as well? Will simplify the ocde later on.

manifests/params.pp
```diff
@@ -176,13 +194,19 @@
 
   # Passenger config
   $server_app_root = "${dir}/rack"
-  $server_ssl_dir  = "${vardir}/ssl"
+  $server_ssl_dir  = versioncmp($::puppetversion, '4.0') ? {
+    '-1'    => "${vardir}/ssl",
+    default => "${ssldir}",
```
ekohl: For this to work you need to set $ssldir explicitly for the BSDs instead of ```'$vardir/ssl'``` which relies on puppet to replace the var. It also gives a lint error now.

manifests/params.pp
mcanevet: done

manifests/params.pp
```diff
@@ -176,13 +188,19 @@
 
   # Passenger config
   $server_app_root = "${dir}/rack"
-  $server_ssl_dir  = "${vardir}/ssl"
+  $server_ssl_dir  = versioncmp($::puppetversion, '4.0') ? {
```
ekohl: Is this still needed if we set ```$ssldir``` for every OS/distro?

mmoll: at least the path itself different for FreeBSD, so probably yes

ekohl: What is it for FreeBSD? In the definition we use above it's the same as Linux so maybe that should be fixed instead.

mmoll: after clarification on IRC: I'm perfectly fine with expanding $ssldir :)

manifests/params.pp
```diff
@@ -190,10 +208,14 @@
...
     $puppetrun_cmd = '/usr/sbin/puppetrun'
-  } else {
+  } elsif versioncmp($::puppetversion, '4.0') < 0 {
     $puppetca_path = '/usr/bin'
     $puppetca_bin = 'puppet cert'
     $puppetrun_cmd = '/usr/bin/puppet kick'
+  } else {
+    $puppetca_path = '/opt/puppetlabs/bin'
+    $puppetca_bin = 'puppet cert'
+    $puppetrun_cmd = '/opt/puppetlabs/bin/puppet kick'
```
mmoll: ```puppet kick``` got removed in 4.x and is already deprecated in 3.x, maybe it makes sense to remove it here completely and handle it only in the foreman_proxy module.

manifests/params.pp
ekohl: In short: ```$server_ssl_dir = $ssldir``` should work.

ekohl:
I think this would also be a step in the right direction for http://projects.theforeman.org/issues/9588.

mmoll:
indeed, there would be even the ```is_pe``` fact for that, AFAIK

manifests/server/config.pp
```diff
@@ -117,7 +117,7 @@
     }
 
   }
-  elsif ! $puppet::server_dynamic_environments {
+  elsif ! $puppet::server_dynamic_environments and !$puppet::server_directory_environments {
```
ekohl: For my reference: separate PR in https://github.com/theforeman/puppet-puppet/pull/286

spec/classes/puppet_agent_install_spec.rb
```diff
@@ -15,8 +15,14 @@
         :puppetversion => Puppet.version,
       } end
 
-      it 'should not define provider' do
```
ekohl: I prefer the pattern of the other files you changed: set a package_name variable and don't duplicate the tests.

roidelapluie:
I have made some changes in separate PR in #286

mmoll:
see #290

PR #212, opened 2015-01-02
Added support for pluginfactsource, required for SRV and 3.7

zdenekjanda:
RC1 sucessor of #210.

PR #283, opened 2015-07-13
pluginfactsource is not correctly set when using srv records

attachmentgenie:
by adding the missing parameter to the puppet::config class, the pluginfactsource source variable is now correctly set in puppet.conf.

This change is required to make the code from #212 work.


mmoll:
:+1:

ekohl:
Thanks!

PR #286, opened 2015-07-14
Don't configure defaults environments if directory_environment is set

mcanevet:


mmoll:
looking good visually :+1:

domcleal:
I don't think this behaviour should be changed, it is deliberate that we set up the environments in $server_environments whether you're using directory envs or config envs (they were written to be equivalent where possible).

What's the reason you're changing it, it might help me understand what might need to change?

jlambert121:
I was going to write a PR for this as well, but it looks like it's already here.  We manage out environments with r10k outside of any modules, puppet creates these environments for us and then r10k removes them causing a "change" on the next puppet run.

mmoll:
Maybe this can get wrapped into an extra parameter?
Also, this needs a rebase :)

mmoll:
what's the status here?

mcanevet:
@mmoll I think we can close this as we don't use this module anymore

mmoll:
@mcanevet sorry to hear that :cry:
@jlambert121 feel free to open a new PR about this

jlambert121:
#348

PR #287, opened 2015-07-17
puppetlabs-puppetdb 5.x is incompatible

mmoll:
We need to revistit that, once puppet(master) 4.x gets supported by us.


domcleal:
Thanks!

PR #288, opened 2015-07-19
support puppet master setup on FreeBSD

mmoll:


manifests/server/rack.pp
```diff
@@ -16,9 +16,13 @@
   $rack_arguments = $::puppet::server_rack_arguments,
   $user           = $::puppet::server_user,
   $vardir         = $::puppet::vardir,
+  $touch_cmd      = $::osfamily ? {
+    /^(FreeBSD|DragonFly)$/ => '/usr/bin/touch',
+    default                 => '/bin/touch'
```
mmoll: Not 100% sure about that, as I read that historically only RedHat/Fedora had this in /bin. However, Debian has a symlink by default, so this never got noticed.

manifests/server/install.pp
```diff
@@ -2,9 +2,10 @@
 class puppet::server::install {
 
   $server_package_default = $::puppet::server_implementation ? {
-    'master'       => $::operatingsystem ? {
-      /(Debian|Ubuntu)/ => ['puppetmaster-common','puppetmaster'],
-      default           => ['puppet-server'],
+    'master'       => $::osfamily ? {
+      /(Debian|Ubuntu)/       => ['puppetmaster-common','puppetmaster'],
```
ekohl: Debian should do if you're changing it to osfamily.

manifests/server/rack.pp
ekohl: I'm not a fan of this complex logic in class parameters. Why not use ```touch``` as command and set a ```path``` which includes those directories?

mmoll: indeed, changed that. :bow:

manifests/server/install.pp
mmoll: it does ;)

ekohl:
:+1:

.gitignore
```diff
@@ -34,3 +34,6 @@ spec/fixtures/
 ## Puppet module
 pkg/
 coverage/
+
+# JetBrains IDEs
```
domcleal: You'll need to put this in modulesync, it's a synced file.

domcleal:
:+1: otherwise.

.gitignore
mmoll: fixed.

ekohl:
Thanks!

PR #289, opened 2015-07-23
Document sharedir parameter

domcleal:
I think this is the reason the nightly RPM build failed: http://ci.theforeman.org/job/packaging_build_rpm/2082/console

Ah yes, installers are now failing at runtime with:

    [FATAL 2015-07-23 07:50:06 fatal] Unable to continue because of: undocumented parameters in puppet: sharedir

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -23,6 +23,8 @@
 #
 # $ssldir::                        Override where SSL certificates are kept.
 #
+# $sharedir::                      Override the system data directory.
+#
 # $package_provider::              The provider used to install the agent.
 #                                  Defaults to chocolatey on Windows
 #                                  Defaults to undef elsewhere


ekohl:
Thanks!

mmoll:
thanks @domcleal and sorry for the breakage :sob:

domcleal:
Perhaps my fault for not finishing with https://github.com/domcleal/puppet-lint-param-docs!

It works, but I wasn't sure where to begin having it apply to our main classes.  If we simply enabled it like other lint plugins then it would check all classes and defines, which isn't something we really want in current state.  Perhaps we'd need a custom rake task appended to the end of spec that would only scan init.pp with a different puppet-lint config file.

mmoll:
well, also my fault for not testing it with -installer and not having made more progress with FreeBSD support in all modules to use -installer also on my BSD systems. ;)

ekohl:
@domcleal I think you should have a setting on which files to apply it to and default to ```init.pp```. Maybe support globbing so we can use ```compute/*.pp``` in puppet-foreman.

domcleal:
Needed a bit of fiddling around, but now implemented.  https://github.com/theforeman/foreman-installer-modulesync/pull/14 puts support into our msync config so we can start enabling it on each module.

domcleal:
Now it won't ever happen again :) #292

PR #290, opened 2015-07-23
All-in-one package support

roidelapluie:
Merge of PR #282

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -45,6 +45,7 @@ class puppet::params {
       $dir_prefix = 'C:/ProgramData/PuppetLabs/puppet'
 
       $dir        = "${dir_prefix}/etc"
+      $codedir    = "${dir_prefix}/etc"
       $logdir     = "${dir_prefix}/var/log"
       $rundir     = "${dir_prefix}/var/run"
       $ssldir     = "${dir_prefix}/etc/ssl"
@@ -55,6 +56,7 @@ class puppet::params {
 
     /^(FreeBSD|DragonFly)$/ : {
       $dir        = '/usr/local/etc/puppet'
+      $codedir    = '/usr/local/etc/puppet'
       $logdir     = '/var/log/puppet'
       $rundir     = '/var/run/puppet'
       $ssldir     = '/var/puppet/ssl'
@@ -64,12 +66,23 @@ class puppet::params {
     }
 
     default : {
-      $dir        = '/etc/puppet'
-      $logdir     = '/var/log/puppet'
-      $rundir     = '/var/run/puppet'
-      $ssldir     = '/var/lib/puppet/ssl'
-      $vardir     = '/var/lib/puppet'
-      $sharedir   = '/usr/share/puppet'
+      if versioncmp($::puppetversion, '4.0') < 0 {
+        $dir        = '/etc/puppet'
+        $codedir    = '/etc/puppet'
+        $logdir     = '/var/log/puppet'
+        $rundir     = '/var/run/puppet'
+        $ssldir     = '/var/lib/puppet/ssl'
+        $vardir     = '/var/lib/puppet'
+        $sharedir   = '/usr/share/puppet'
+      } else {
+        $dir      = '/etc/puppetlabs/puppet'
+        $codedir  = '/etc/puppetlabs/code'
+        $logdir   = '/var/log/puppetlabs/puppet'
+        $rundir   = '/var/run/puppetlabs'
+        $ssldir   = '/etc/puppetlabs/puppet/ssl'
+        $vardir   = '/opt/puppetlabs/puppet/cache'
+        $sharedir = '/opt/puppetlabs/puppet/modules'
+      }
       $root_group = undef
     }
   }
@@ -154,11 +167,11 @@ class puppet::params {
   $server_environments_group   = $root_group
   $server_environments_mode    = '0755'
   # Where we store our puppet environments
-  $server_envs_dir             = "${dir}/environments"
+  $server_envs_dir             = "${codedir}/environments"
   # Where remains our manifests dir
-  $server_manifest_path        = "${dir}/manifests"
+  $server_manifest_path        = "${codedir}/manifests"
   # Modules in this directory would be shared across all environments
-  $server_common_modules_path  = ["${server_envs_dir}/common", "${dir}/modules", "${sharedir}/modules"]
+  $server_common_modules_path  = ["${server_envs_dir}/common", "${codedir}/modules", "${sharedir}/modules"]
 
   # Dynamic environments config, ignore if the git_repo is 'false'
   # Path to the repository
@@ -180,13 +193,16 @@ class puppet::params {
 
   # Passenger config
   $server_app_root = "${dir}/rack"
-  $server_ssl_dir  = "${vardir}/ssl"
+  $server_ssl_dir  = $ssldir
 
   $server_package     = undef
   $server_version     = undef
-  $client_package     = $::osfamily ? {
-    'Debian' => ['puppet-common','puppet'],
-    default  => ['puppet'],
+  $client_package     = versioncmp($::puppetversion, '4.0') ? {
+    '-1'       => $::osfamily ? {
+      'Debian' => ['puppet-common','puppet'],
+      default  => ['puppet'],
+    },
+    default => 'puppet-agent',
   }
 
   # Only use 'puppet cert' on versions where puppetca no longer exists
@@ -194,7 +210,7 @@ class puppet::params {
     $puppetca_path = '/usr/sbin'
     $puppetca_bin  = 'puppetca'
     $puppetrun_cmd = '/usr/sbin/puppetrun'
-  } else {
+  } elsif versioncmp($::puppetversion, '4.0') < 0 {
     $puppetca_path = $::osfamily ? {
       /^(FreeBSD|DragonFly)$/ => '/usr/local/bin',
       default                 => '/usr/bin'
@@ -204,6 +220,10 @@ class puppet::params {
       /^(FreeBSD|DragonFly)$/ => '/usr/local/bin/puppet kick',
       default                 => '/usr/bin/puppet kick'
     }
+  } else {
+    $puppetca_path = '/opt/puppetlabs/bin'
+    $puppetca_bin = 'puppet cert'
+    $puppetrun_cmd = '/opt/puppetlabs/bin/puppet kick'
   }
 
   $puppetca_cmd = "${puppetca_path}/${puppetca_bin}"
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -38,6 +38,13 @@ class puppet::params {
   $hiera_config        = '$confdir/hiera.yaml'
   $syslogfacility      = undef
   $environment         = $::environment
+  if versioncmp($::puppetversion, '4.0') < 0 {
+    $aio_package = false
+  } elsif $::osfamily == 'Windows' or $::rubysitedir =~ /\/opt\/puppetlabs\/puppet/ {
+    $aio_package = true
+  } else {
+    $aio_package = false
+  }
 
   case $::osfamily {
     'Windows' : {
@@ -66,7 +73,15 @@ class puppet::params {
     }
 
     default : {
-      if versioncmp($::puppetversion, '4.0') < 0 {
+      if $aio_package {
+        $dir      = '/etc/puppetlabs/puppet'
+        $codedir  = '/etc/puppetlabs/code'
+        $logdir   = '/var/log/puppetlabs/puppet'
+        $rundir   = '/var/run/puppetlabs'
+        $ssldir   = '/etc/puppetlabs/puppet/ssl'
+        $vardir   = '/opt/puppetlabs/puppet/cache'
+        $sharedir = '/opt/puppetlabs/puppet'
+      } else {
         $dir        = '/etc/puppet'
         $codedir    = '/etc/puppet'
         $logdir     = '/var/log/puppet'
@@ -74,24 +89,30 @@ class puppet::params {
         $ssldir     = '/var/lib/puppet/ssl'
         $vardir     = '/var/lib/puppet'
         $sharedir   = '/usr/share/puppet'
-      } else {
-        $dir      = '/etc/puppetlabs/puppet'
-        $codedir  = '/etc/puppetlabs/code'
-        $logdir   = '/var/log/puppetlabs/puppet'
-        $rundir   = '/var/run/puppetlabs'
-        $ssldir   = '/etc/puppetlabs/puppet/ssl'
-        $vardir   = '/opt/puppetlabs/puppet/cache'
-        $sharedir = '/opt/puppetlabs/puppet/modules'
       }
       $root_group = undef
     }
   }
 
   $manage_packages = true
+
+  if $aio_package and $::osfamily != 'Windows' {
+    $dir_owner = 'root'
+    $dir_group = $root_group
+  } elsif $::osfamily == 'Windows' {
+    $dir_owner = undef
+    $dir_group = undef
+  } else {
+    $dir_owner = $user
+    $dir_group = $group
+  }
+
   $package_provider = $::osfamily ? {
     'windows' => 'chocolatey',
     default   => undef,
   }
+
+  $package_source = undef
 
   # Need your own config templates? Specify here:
   $main_template   = 'puppet/puppet.conf.erb'
@@ -197,12 +218,12 @@ class puppet::params {
 
   $server_package     = undef
   $server_version     = undef
-  $client_package     = versioncmp($::puppetversion, '4.0') ? {
-    '-1'       => $::osfamily ? {
+  $client_package     = $aio_package ? {
+    true       => ['puppet-agent'],
+    default    => $::osfamily ? {
       'Debian' => ['puppet-common','puppet'],
       default  => ['puppet'],
     },
-    default => 'puppet-agent',
   }
 
   # Only use 'puppet cert' on versions where puppetca no longer exists
@@ -210,7 +231,11 @@ class puppet::params {
     $puppetca_path = '/usr/sbin'
     $puppetca_bin  = 'puppetca'
     $puppetrun_cmd = '/usr/sbin/puppetrun'
-  } elsif versioncmp($::puppetversion, '4.0') < 0 {
+  } elsif $aio_package {
+    $puppetca_path = '/opt/puppetlabs/bin'
+    $puppetca_bin = 'puppet cert'
+    $puppetrun_cmd = '/opt/puppetlabs/bin/puppet kick'
+  } else {
     $puppetca_path = $::osfamily ? {
       /^(FreeBSD|DragonFly)$/ => '/usr/local/bin',
       default                 => '/usr/bin'
@@ -220,10 +245,6 @@ class puppet::params {
       /^(FreeBSD|DragonFly)$/ => '/usr/local/bin/puppet kick',
       default                 => '/usr/bin/puppet kick'
     }
-  } else {
-    $puppetca_path = '/opt/puppetlabs/bin'
-    $puppetca_bin = 'puppet cert'
-    $puppetrun_cmd = '/opt/puppetlabs/bin/puppet kick'
   }
 
   $puppetca_cmd = "${puppetca_path}/${puppetca_bin}"
--- a/spec/classes/puppet_agent_install_spec.rb
+++ b/spec/classes/puppet_agent_install_spec.rb
@@ -2,45 +2,49 @@ require 'spec_helper'
 
 describe 'puppet::agent::install' do
 
+  if Puppet.version < '4.0'
+    additional_facts = {}
+  else
+    additional_facts = {:rubysitedir => '/opt/puppetlabs/puppet/lib/ruby/site_ruby/2.1.0'}
+  end
+
+  let :common_centos_facts do on_supported_os['centos-6-x86_64'].merge({
+    :concat_basedir => '/nonexistant',
+    :puppetversion  => Puppet.version,
+  }).merge(additional_facts) end
+
   describe 'with default parameters' do
     let :pre_condition do
       'include ::puppet'
     end
 
-    context "on a RedHat family OS" do
-      let :facts do {
-        :osfamily => 'RedHat',
-        :concat_basedir => '/foo/bar',
-        :operatingsystemrelease => '6.6',
-        :puppetversion => Puppet.version,
-      } end
+    if Puppet.version < '4.0'
+      client_package = 'puppet'
+    else
+      client_package = 'puppet-agent'
+    end
 
-      if Puppet.version < '4.0'
-        it 'should not define provider' do
-          should contain_package('puppet').without_provider(nil)
-        end
-      else
-        it 'should not define provider' do
-          should contain_package('puppet-agent').without_provider(nil)
-        end
+    context "on a RedHat family OS" do
+      let :facts do
+        common_centos_facts
+      end
+
+      it 'should not define provider' do
+        should contain_package(client_package).without_provider(nil)
       end
     end
 
     context "on a Windows family OS" do
-      let :facts do {
-        :osfamily => 'windows',
-        :concat_basedir => 'C:\Temp',
-        :puppetversion => Puppet.version,
-      } end
+      let :facts do
+        {
+          :osfamily => 'windows',
+          :concat_basedir => 'C:\Temp',
+          :puppetversion => Puppet.version,
+        }
+      end
 
-      if Puppet.version < '4.0'
-        it 'should define provider as chocolatey' do
-          should contain_package('puppet').with_provider('chocolatey')
-        end
-      else
-        it 'should define provider as chocolatey' do
-          should contain_package('puppet-agent').with_provider('chocolatey')
-        end
+      it 'should define provider as chocolatey' do
+        should contain_package(client_package).with_provider('chocolatey')
       end
     end
 
@@ -48,24 +52,56 @@ describe 'puppet::agent::install' do
 
   describe "when package_provider => 'msi'" do
 
+    if Puppet.version < '4.0'
+      client_package = 'puppet'
+    else
+      client_package = 'puppet-agent'
+    end
+
     let :pre_condition do
       "class { 'puppet': package_provider => 'msi', }"
     end
 
-    let :facts do {
-      :osfamily => 'windows',
-      :concat_basedir => 'C:\Temp',
-      :puppetversion => Puppet.version,
-    } end
+    let :facts do
+      {
+        :osfamily => 'windows',
+        :concat_basedir => 'C:\Temp',
+        :puppetversion => Puppet.version,
+      }
+    end
+
+    it 'should define provider as msi' do
+      should contain_package(client_package).with_provider('msi')
+    end
+
+  end
+
+  describe "when package_provider => 'windows' and source is defined" do
 
     if Puppet.version < '4.0'
-      it 'should define provider as msi' do
-        should contain_package('puppet').with_provider('msi')
-      end
+      client_package = 'puppet'
     else
-      it 'should define provider as msi' do
-        should contain_package('puppet-agent').with_provider('msi')
-      end
+      client_package = 'puppet-agent'
+    end
+
+    let :pre_condition do
+      "class { 'puppet': package_provider => 'windows', package_source => 'C:\\Temp\\puppet.exe' }"
+    end
+
+    let :facts do
+      {
+        :osfamily => 'windows',
+        :concat_basedir => 'C:\Temp',
+        :puppetversion => Puppet.version,
+      }
+    end
+
+    it 'should define provider as windows' do
+      should contain_package(client_package).with_provider('windows')
+    end
+
+    it 'should define source as C:\Temp\puppet.exe' do
+      should contain_package(client_package).with_source('C:\Temp\puppet.exe')
     end
 
   end
@@ -75,15 +111,18 @@ describe 'puppet::agent::install' do
       "class { 'puppet': manage_packages => false }"
     end
 
-    let :facts do {
-      :osfamily => 'RedHat',
-      :concat_basedir => '/foo/bar',
-      :operatingsystemrelease => '6.6',
-      :puppetversion => Puppet.version,
-    } end
+    context "on a RedHat family OS" do
+      let :facts do
+        common_centos_facts
+      end
 
-    it 'should not contain Package[puppet]' do
-      should_not contain_package('puppet')
+      it 'should not contain Package[puppet]' do
+        should_not contain_package('puppet')
+      end
+
+      it 'should not contain Package[puppet-agent]' do
+        should_not contain_package('puppet-agent')
+      end
     end
   end
 
@@ -92,15 +131,20 @@ describe 'puppet::agent::install' do
       "class { 'puppet': manage_packages => 'agent' }"
     end
 
-    let :facts do {
-      :osfamily => 'RedHat',
-      :concat_basedir => '/foo/bar',
-      :operatingsystemrelease => '6.6',
-      :puppetversion => Puppet.version,
-    } end
+    if Puppet.version < '4.0'
+      client_package = 'puppet'
+    else
+      client_package = 'puppet-agent'
+    end
 
-    it 'should contain Package[puppet]' do
-      should contain_package('puppet')
+    context "on a RedHat family OS" do
+      let :facts do
+        common_centos_facts
+      end
+
+      it 'should contain Package[puppet]' do
+        should contain_package(client_package)
+      end
     end
   end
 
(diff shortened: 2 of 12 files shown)


manifests/params.pp
```diff
@@ -179,13 +192,19 @@
 
   # Passenger config
   $server_app_root = "${dir}/rack"
-  $server_ssl_dir  = "${vardir}/ssl"
+  $server_ssl_dir  = versioncmp($::puppetversion, '4.0') ? {
```
ekohl: This can be as simple as ```$ssldir```.

roidelapluie:
not if we want to change it, right?

ekohl:
```$ssldir``` contains the correct value because it's set in a conditional itself.

**Comment by roidelapluie** [comment]
Fixed

File: manifests/server/config.pp [hunk 2]
```diff
@@ -117,7 +117,7 @@
     }
 
   }
-  elsif ! $puppet::server_dynamic_environments {
+  elsif ! $puppet::server_dynamic_environments and !$puppet::server_directory_environments {
```
ekohl: Also in the other PR: https://github.com/theforeman/puppet-puppet/pull/286 discusses this change. Unless there are some arguments in favour I don't think we want this change.

manifests/params.pp
```diff
@@ -179,13 +192,16 @@
...
   $server_version     = undef
-  $client_package     = $::osfamily ? {
-    'Debian' => ['puppet-common','puppet'],
-    default  => ['puppet'],
+  $client_package     = versioncmp($::puppetversion, '4.0') ? {
+    '-1'     => $::osfamily ? {
+      'debian' => ['puppet-common','puppet'],
+      default   => ['puppet'],
+    },
+    default => 'puppet-agent',
```
mmoll: There is no Puppet 4 package for FreeBSD, yet, but I'm not sure if it would be called "puppet-agent"... Anyway, I know this is not ideal, but to be able to also use non-AIO packages on Linux I'd like to have a variable like ```aio_package```, which could be ```true``` for 4.x by default, but setable, so native 4.x packages are not that problematic.

ekohl: Given there's no Puppet 4 package yet, maybe we should just leave it for now and worry about it when there is one?

mmoll: Fedora 22 already has a puppet 4 package non-AIO style and for FreeBSD I could make one quite quickly, but it would be limited to 4.1, as 4.2 requires the C++ rewrite of facter.

I'd just be happier to have parameters set by puppetversion, where there's a real difference because of the version, and another variable that's used to set things according to the packaging style used (AIO/PE (when needed)/native).

ekohl: I wonder if we should adapt the ```puppet::global``` pattern where you can override the AIO/PE style and then the OS checks come in. Otherwise you need to set parameters in init.pp if a user overrides them. Or would you just have a variable in params.pp for now?

mmoll: Ah, damn... that whole logic would indeed not find it's place in params.pp anymore... yes, let's go for a variable in params.pp then. :heavy_dollar_sign:

roidelapluie:
@mmol @ekohl I have push a compromised. What do you think?

manifests/params.pp
```diff
@@ -38,6 +38,13 @@
   $hiera_config        = '$confdir/hiera.yaml'
   $syslogfacility      = undef
   $environment         = $::environment
+  if versioncmp($::puppetversion, '4.0') < 0 {
+    $aio_package = false
+  } elsif $::rubysitedir =~ /\/opt\/puppetlabs\/puppet/ {
```
ekohl: Nice one

spec/classes/puppet_agent_install_spec.rb
```diff
@@ -7,28 +7,41 @@
...
+    end
+
     context "on a RedHat family OS" do
-      let :facts do {
-        :osfamily => 'RedHat',
-        :concat_basedir => '/foo/bar',
-        :operatingsystemrelease => '6.6',
-        :puppetversion => Puppet.version,
-      } end
+      let :facts do
```
ekohl: It may be better to use ```on_supported_os['redhat-6-x86_64']``` so you also get the ```rubysitedir``` fact, which is now needed and fails tests.

roidelapluie:
Nice trick!

spec/classes/puppet_agent_install_spec.rb
roidelapluie: Nice trick!

mmoll:
I'm just trying to install an AIO puppet-agent for the first time... the package doesn't bring a puppet user anymore?

roidelapluie:
Remember that the agent runs as root... no need for a puppet user

roidelapluie:
if you install puppet-server you will get the user

mmoll:
I see... changing the user and group to root got things going. However, this leads to a situation where the module doesn't work with default settings:
```
[root@f21 pup]# /opt/puppetlabs/bin/puppet apply /root/puppet/agent.pp --modulepath=/root/puppet/modules
Notice: Compiled catalog for f21.kvedulv.de in environment production in 1.16 seconds
Error: Could not find user puppet
Error: /Stage[main]/Puppet::Config/File[/etc/puppetlabs/puppet]/owner: change from root to puppet failed: Could not find user puppet
Error: Could not find group puppet
Error: /Stage[main]/Puppet::Config/File[/etc/puppetlabs/puppet]/group: change from root to puppet failed: Could not find group puppet
[...]
```
Tests with 3.x and 4.x non-AIO packages on Debian and FreeBSD were good, we're getting closer :smiley:

roidelapluie:
What should we do about this?

mmoll:
There's already a split between the user and the server_user. I guess we can set root/root as (agent) user when using AIO packages and change the logic to set puppet/puppet as server_user and then hope, everything is coming out correctly.

roidelapluie:
(I only tested with puppet-server packages ...)

roidelapluie:
ok

roidelapluie:
I have pushed some changes that might fix that.

roidelapluie:
Travis is not launching tests anymore. Will take a look tonight.

mmoll:
Tested, works for me.
:+1:, pending green tests.

roidelapluie:
It's green :)

manifests/params.pp
```diff
@@ -64,16 +73,35 @@
...
+        $ssldir   = '/etc/puppetlabs/puppet/ssl'
+        $vardir   = '/opt/puppetlabs/puppet/cache'
+        $sharedir = '/opt/puppetlabs/puppet'
+      }
       $root_group = undef
     }
   }
 
+  if ($aio_package){
+    $dir_owner = 'root'
```
ekohl: Won't this break on Windows?

roidelapluie: I have zéro idéal.

roidelapluie:
Zéro idea

spec/classes/puppet_server_spec.rb
```diff
@@ -2,16 +2,30 @@
 
 describe 'puppet::server' do
 
-  let :default_facts do {
+  let :common_facts do on_supported_os['centos-6-x86_64'].merge({
     :concat_basedir         => '/nonexistant',
     :clientcert             => 'puppetmaster.example.com',
     :fqdn                   => 'puppetmaster.example.com',
-    :operatingsystemrelease => '6.5',
     :osfamily               => 'RedHat',
```
ekohl: You can remove this and ```rubyversion``` since it's in the facts you're merging.

roidelapluie:
T think we should be fine now.

mmoll:
@genebean could you test this on an WIndows machine?

genebean:
I should be able to do so today.

manifests/params.pp
```diff
@@ -64,16 +73,35 @@
     }
 
     default : {
-      $dir        = '/etc/puppet'
-      $logdir     = '/var/log/puppet'
-      $rundir     = '/var/run/puppet'
-      $ssldir     = '/var/lib/puppet/ssl'
-      $vardir     = '/var/lib/puppet'
-      $sharedir   = '/usr/share/puppet'
+      if ($aio_package) {
```
ekohl: nitpick: we don't use braces for these simple if statements.

manifests/params.pp
```diff
@@ -64,16 +73,35 @@
...
+        $rundir     = '/var/run/puppet'
+        $ssldir     = '/var/lib/puppet/ssl'
+        $vardir     = '/var/lib/puppet'
+        $sharedir   = '/usr/share/puppet'
+      }
       $root_group = undef
     }
   }
 
+  if ($aio_package and $::osfamily != 'Windows'){
```
ekohl: nitpick: we don't use braces for these simple if statements.

manifests/params.pp
```diff
@@ -64,16 +73,35 @@
...
+    $dir_owner = 'root'
+    $dir_group = 'root'
```
ekohl: BSD doesn't have a root group (wheel I believe).

roidelapluie: it does not have AIO neither :)

roidelapluie: Fixed by using $root_group (2c17dfa6c5b464cd4d5a09a2f667de95dfed8070)

manifests/params.pp
roidelapluie: corrected

manifests/params.pp
roidelapluie: corrected

genebean:
today got away from me, will try and test in the am.

genebean:
FYI, looking into this now...

spec/classes/puppet_agent_install_spec.rb
```diff
@@ -21,14 +32,16 @@
...
+        {
+          :osfamily => 'windows',
+          :concat_basedir => 'C:\Temp',
+          :puppetversion => Puppet.version,
+        }.merge(additional_facts)
+      end
 
       it 'should define provider as chocolatey' do
-        should contain_package('puppet').with_provider('chocolatey')
+        should contain_package(client_package).with_provider('chocolatey')
```
genebean: As of this moment, a `$client_package` named `puppet-agent` is not available via chocolatey. I just sent a message to the maintainers over there and am waiting to hear back.

roidelapluie: but the puppet version will never be > 4.0 at this point:

https://chocolatey.org/packages/puppet

Can we merge this PR then?

genebean: If the provider is switched to MSI it can be > 4.0. I am working to test it with that provider now... will hopefully have a result in the next 2 hours.

manifests/params.pp
```diff
@@ -64,16 +73,35 @@
...
     }
   }
 
+  if $aio_package and $::osfamily != 'Windows' {
+    $dir_owner = 'root'
+    $dir_group = $root_group
+  } else {
+    $dir_owner = $user
+    $dir_group = $group
+  }
```
genebean: On Windows we do not want to set the owner explicitly and there is no user named puppet so, as it stands, this else block causes a failed run. If you make the following change it fixes it:

```puppet
if $aio_package and $::osfamily != 'Windows' {
  $dir_owner = 'root'
  $dir_group = $root_group
} else {
  $dir_owner = undef
  $dir_group = undef
}
```

spec/classes/puppet_agent_install_spec.rb
genebean: While testing the MSI-based install I found a couple of bugs not directly related to these changes. I do we want to fix them prior to merging this PR or included them in their own PR... I am thinking the latter.

genebean:
Just to sum things up, here's a git diff showing all the changes I made locally to address my inline comments. After doing these it seems to be working. As mentioned above, I think the stuff related to installing via MSI can be a different PR since the changes are really unrelated to this one.

```diff
diff --git a/manifests/agent/install.pp b/manifests/agent/install.pp
index 0554b6d..ea65b73 100644
--- a/manifests/agent/install.pp
+++ b/manifests/agent/install.pp
@@ -3,5 +3,6 @@ class puppet::agent::install {
   package { $puppet::client_package:
     ensure   => $::puppet::version,
     provider => $::puppet::package_provider,
+       source   => $::puppet::agent_package_source,
   }
 }
diff --git a/manifests/init.pp b/manifests/init.pp
index 02f98e6..745e074 100644
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -31,6 +31,12 @@
 #                                  Defaults to chocolatey on Windows
 #                                  Defaults to undef elsewhere
 #
+# $agent_package_source::          The location of the file to be used by the
+#                                  agent's package resource
+#                                  Defaults to undef. If 'windows' or 'msi' are
+#                                  used as the provider then this setting is
+#                                  required.
+#
 # $port::                          Override the port of the master we connect to.
 #                                  type:integer
 #
@@ -395,6 +401,7 @@ class puppet (
   $dir_owner                     = $puppet::params::dir_owner,
   $dir_group                     = $puppet::params::dir_group,
   $package_provider              = $puppet::params::package_provider,
+  $agent_package_source          = $puppet::params::agent_package_source,
   $port                          = $puppet::params::port,
   $listen                        = $puppet::params::listen,
   $listen_to                     = $puppet::params::listen_to,
diff --git a/manifests/params.pp b/manifests/params.pp
index 1b0b08c..25fa3cc 100644
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -98,14 +98,16 @@ class puppet::params {
     $dir_owner = 'root'
     $dir_group = $root_group
   } else {
-    $dir_owner = $user
-    $dir_group = $group
+    $dir_owner = undef
+    $dir_group = undef
   }

   $package_provider = $::osfamily ? {
     'windows' => 'chocolatey',
     default   => undef,
   }
+
+  $agent_package_source = undef

   # Need your own config templates? Specify here:
   $main_template   = 'puppet/puppet.conf.erb'
diff --git a/spec/classes/puppet_agent_install_spec.rb b/spec/classes/puppet_agent_install_spec.rb
index 32d3def..0f2b6cb 100644
--- a/spec/classes/puppet_agent_install_spec.rb
+++ b/spec/classes/puppet_agent_install_spec.rb
@@ -47,7 +47,7 @@ describe 'puppet::agent::install' do

   end

-  describe "when package_provider => 'msi'" do
+  describe "when package_provider => 'windows'" do

     if Puppet.version < '4.0'
       client_package = 'puppet'
@@ -56,7 +56,7 @@ describe 'puppet::agent::install' do
     end

     let :pre_condition do
-      "class { 'puppet': package_provider => 'msi', }"
+      "class { 'puppet': package_provider => 'windows', }"
     end

     let :facts do
@@ -67,8 +67,8 @@ describe 'puppet::agent::install' do
       }
     end

-    it 'should define provider as msi' do
-      should contain_package(client_package).with_provider('msi')
+    it 'should define provider as windows' do
+      should contain_package(client_package).with_provider('windows')
     end

   end
```

roidelapluie:
I have addressed your comments.

genebean:
Sorry, just saw this. Will test asap.

genebean:
Seems to be working. Below is the output from two runs. The first is on a system that already has 3.x on it. The second is post-install where it actually runs the new agent. Not sure how it relates, but please take note of the errors near the line `Error: Could not find any content at puppet:///modules/concat/concatfragments.rb` in case they mean more to you all.

## Run 1

```powershell
Windows PowerShell
Copyright (C) 2009 Microsoft Corporation. All rights reserved.

PS C:\programdata\PuppetLabs\puppet\etc\modules\puppet> puppet apply --modulepath=../ -e "class { 'puppet': package_prov
ider => 'windows', package_source => 'C:\Users\gliverma\Downloads\puppet-agent-x64-latest.msi'}
>>
PS C:\programdata\PuppetLabs\puppet\etc\modules\puppet> puppet apply --modulepath=../ -e "class { 'puppet': package_prov
ider => 'windows', package_source => 'C:\Users\gliverma\Downloads\puppet-agent-x64-latest.msi'}"
Error loading fact C:/ProgramData/PuppetLabs/puppet/var/lib/facter/facts.rb: cannot load such file -- util/sssd
Error loading fact C:/ProgramData/PuppetLabs/puppet/var/lib/facter/facts.rb: cannot load such file -- util/sssd
Warning: Config file C:/ProgramData/PuppetLabs/puppet/etc/hiera.yaml not found, using Hiera defaults
Notice: Compiled catalog for win7-puppet.uwg.westga.edu in environment production in 51.90 seconds
Notice: /Stage[main]/Concat::Setup/File[C:/ProgramData/PuppetLabs/puppet/var/concat]/ensure: created
Notice: /Stage[main]/Concat::Setup/File[C:/ProgramData/PuppetLabs/puppet/var/concat/bin]/ensure: created
Warning: Copying owner/mode/group from the source file on Windows is deprecated; use source_permissions => ignore.
   (at C:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet/type/file/source.rb:133:in `block in copy_source_values')
Error: Could not find any content at puppet:///modules/concat/concatfragments.rb
Error: /Stage[main]/Concat::Setup/File[C:/ProgramData/PuppetLabs/puppet/var/concat/bin/concatfragments.rb]/ensure: chang
e from absent to file failed: Could not find any content at puppet:///modules/concat/concatfragments.rb
Notice: /Stage[main]/Puppet::Agent::Install/Package[puppet]/ensure: created
Notice: /Stage[main]/Puppet::Config/Concat[C:/ProgramData/PuppetLabs/puppet/etc/puppet.conf]/File[C:/ProgramData/PuppetL
abs/puppet/var/concat/C__ProgramData_PuppetLabs_puppet_etc_puppet.conf]/ensure: created
Notice: /Stage[main]/Puppet::Config/Concat[C:/ProgramData/PuppetLabs/puppet/etc/puppet.conf]/File[C:/ProgramData/PuppetL
abs/puppet/var/concat/C__ProgramData_PuppetLabs_puppet_etc_puppet.conf/fragments]/ensure: created
Notice: /Stage[main]/Puppet::Agent::Config/Concat::Fragment[puppet.conf+20-agent]/File[C:/ProgramData/PuppetLabs/puppet/
var/concat/C__ProgramData_PuppetLabs_puppet_etc_puppet.conf/fragments/20_puppet.conf+20-agent]/ensure: defined content a
s '{md5}740e51efe8f7c8ed90397c2005e4464a'
Notice: /Stage[main]/Puppet::Config/Concat[C:/ProgramData/PuppetLabs/puppet/etc/puppet.conf]/File[C:/ProgramData/PuppetL
abs/puppet/var/concat/C__ProgramData_PuppetLabs_puppet_etc_puppet.conf/fragments.concat.out]/ensure: created
Notice: /Stage[main]/Puppet::Config/Concat[C:/ProgramData/PuppetLabs/puppet/etc/puppet.conf]/File[C:/ProgramData/PuppetL
abs/puppet/var/concat/C__ProgramData_PuppetLabs_puppet_etc_puppet.conf/fragments.concat]/ensure: created
Notice: /Stage[main]/Puppet::Config/Concat::Fragment[puppet.conf+10-main]/File[C:/ProgramData/PuppetLabs/puppet/var/conc
at/C__ProgramData_PuppetLabs_puppet_etc_puppet.conf/fragments/10_puppet.conf+10-main]/ensure: defined content as '{md5}5
b04ba5894daaac16e837ccadf9b71c8'
Notice: /Stage[main]/Puppet::Config/Concat[C:/ProgramData/PuppetLabs/puppet/etc/puppet.conf]/Exec[concat_C:/ProgramData/
PuppetLabs/puppet/etc/puppet.conf]/returns: ruby.exe: No such file or directory -- C:/ProgramData/PuppetLabs/puppet/var/
concat/bin/concatfragments.rb (LoadError)
Error: ruby.exe 'C:/ProgramData/PuppetLabs/puppet/var/concat/bin/concatfragments.rb' -o "C:/ProgramData/PuppetLabs/puppe
t/var/concat/C__ProgramData_PuppetLabs_puppet_etc_puppet.conf/fragments.concat.out" -d "C:/ProgramData/PuppetLabs/puppet
/var/concat/C__ProgramData_PuppetLabs_puppet_etc_puppet.conf" returned 1 instead of one of [0]
Error: /Stage[main]/Puppet::Config/Concat[C:/ProgramData/PuppetLabs/puppet/etc/puppet.conf]/Exec[concat_C:/ProgramData/P
uppetLabs/puppet/etc/puppet.conf]/returns: change from notrun to 0 failed: ruby.exe 'C:/ProgramData/PuppetLabs/puppet/va
r/concat/bin/concatfragments.rb' -o "C:/ProgramData/PuppetLabs/puppet/var/concat/C__ProgramData_PuppetLabs_puppet_etc_pu
ppet.conf/fragments.concat.out" -d "C:/ProgramData/PuppetLabs/puppet/var/concat/C__ProgramData_PuppetLabs_puppet_etc_pup
pet.conf" returned 1 instead of one of [0]
Notice: /Stage[main]/Puppet::Config/Concat[C:/ProgramData/PuppetLabs/puppet/etc/puppet.conf]/Exec[concat_C:/ProgramData/
PuppetLabs/puppet/etc/puppet.conf]/returns: ruby.exe: No such file or directory -- C:/ProgramData/PuppetLabs/puppet/var/
concat/bin/concatfragments.rb (LoadError)
Error: /Stage[main]/Puppet::Config/Concat[C:/ProgramData/PuppetLabs/puppet/etc/puppet.conf]/Exec[concat_C:/ProgramData/P
uppetLabs/puppet/etc/puppet.conf]: Failed to call refresh: ruby.exe 'C:/ProgramData/PuppetLabs/puppet/var/concat/bin/con
catfragments.rb' -o "C:/ProgramData/PuppetLabs/puppet/var/concat/C__ProgramData_PuppetLabs_puppet_etc_puppet.conf/fragme
nts.concat.out" -d "C:/ProgramData/PuppetLabs/puppet/var/concat/C__ProgramData_PuppetLabs_puppet_etc_puppet.conf" return
ed 1 instead of one of [0]
Error: /Stage[main]/Puppet::Config/Concat[C:/ProgramData/PuppetLabs/puppet/etc/puppet.conf]/Exec[concat_C:/ProgramData/P
uppetLabs/puppet/etc/puppet.conf]: ruby.exe 'C:/ProgramData/PuppetLabs/puppet/var/concat/bin/concatfragments.rb' -o "C:/
ProgramData/PuppetLabs/puppet/var/concat/C__ProgramData_PuppetLabs_puppet_etc_puppet.conf/fragments.concat.out" -d "C:/P
rogramData/PuppetLabs/puppet/var/concat/C__ProgramData_PuppetLabs_puppet_etc_puppet.conf" returned 1 instead of one of [
0]
Notice: /Stage[main]/Puppet::Config/Concat[C:/ProgramData/PuppetLabs/puppet/etc/puppet.conf]/File[C:/ProgramData/PuppetL
abs/puppet/etc/puppet.conf]: Dependency Exec[concat_C:/ProgramData/PuppetLabs/puppet/etc/puppet.conf] has failures: true

Warning: /Stage[main]/Puppet::Config/Concat[C:/ProgramData/PuppetLabs/puppet/etc/puppet.conf]/File[C:/ProgramData/Puppet
Labs/puppet/etc/puppet.conf]: Skipping because of failed dependencies
Notice: /Stage[main]/Puppet::Config/File[C:/ProgramData/PuppetLabs/puppet/etc/auth.conf]: Dependency Exec[concat_C:/Prog
ramData/PuppetLabs/puppet/etc/puppet.conf] has failures: true
Warning: /Stage[main]/Puppet::Config/File[C:/ProgramData/PuppetLabs/puppet/etc/auth.conf]: Skipping because of failed de
pendencies
Notice: /Stage[main]/Puppet::Agent::Service/Service[puppet]: Dependency Exec[concat_C:/ProgramData/PuppetLabs/puppet/etc
/puppet.conf] has failures: true
Warning: /Stage[main]/Puppet::Agent::Service/Service[puppet]: Skipping because of failed dependencies
Notice: /Stage[main]/Puppet::Agent::Service/Service[puppet]: Triggered 'refresh' from 1 events
Notice: Finished catalog run in 137.27 seconds
```

## Run 2

```powershell
PS C:\programdata\PuppetLabs\puppet\etc\modules\puppet> puppet apply --modulepath=../ -e "class { 'puppet': package_prov
ider => 'windows', package_source => 'C:\Users\gliverma\Downloads\puppet-agent-x64-latest.msi'}"
Warning: Setting configtimeout is deprecated.
   (at C:/Program Files/Puppet Labs/Puppet/puppet/lib/puppet/settings.rb:1142:in `issue_deprecation_warning')
Error: Facter: error while resolving custom facts in C:/ProgramData\PuppetLabs\puppet\var\lib\facter\facts.rb: cannot lo
ad such file -- util/sssd
Error: Facter: error while resolving custom facts in C:/ProgramData\PuppetLabs\puppet\var\lib\facter\portage.rb: cannot
load such file -- facter/util/resolution
Warning: Config file C:/ProgramData/PuppetLabs/puppet/etc/hiera.yaml not found, using Hiera defaults
Notice: Compiled catalog for win7-puppet.uwg.westga.edu in environment production in 3.93 seconds
Notice: /Stage[main]/Puppet::Agent::Install/Package[puppet-agent]/ensure: created
Notice: /Stage[main]/Concat::Setup/File[C:/ProgramData/PuppetLabs/puppet/var/concat/bin/concatfragments.rb]/ensure: defi
ned content as '{md5}b684db0eac243553a6a79365119a363d'
Notice: /Stage[main]/Puppet::Config/Concat::Fragment[puppet.conf+10-main]/File[C:/ProgramData/PuppetLabs/puppet/var/conc
at/C__ProgramData_PuppetLabs_puppet_etc_puppet.conf/fragments/10_puppet.conf+10-main]/content: content changed '{md5}5b0
4ba5894daaac16e837ccadf9b71c8' to '{md5}61606ec8cd8af1b7496beb079165b278'
Notice: /Stage[main]/Puppet::Agent::Config/Concat::Fragment[puppet.conf+20-agent]/File[C:/ProgramData/PuppetLabs/puppet/
var/concat/C__ProgramData_PuppetLabs_puppet_etc_puppet.conf/fragments/20_puppet.conf+20-agent]/content: content changed
'{md5}740e51efe8f7c8ed90397c2005e4464a' to '{md5}1f06b9ccbee5ae7c66985b9a96b18a52'
Notice: /Stage[main]/Puppet::Config/Concat[C:/ProgramData/PuppetLabs/puppet/etc/puppet.conf]/Exec[concat_C:/ProgramData/
PuppetLabs/puppet/etc/puppet.conf]/returns: executed successfully
Notice: /Stage[main]/Puppet::Config/Concat[C:/ProgramData/PuppetLabs/puppet/etc/puppet.conf]/Exec[concat_C:/ProgramData/
PuppetLabs/puppet/etc/puppet.conf]: Triggered 'refresh' from 2 events
Notice: /Stage[main]/Puppet::Config/Concat[C:/ProgramData/PuppetLabs/puppet/etc/puppet.conf]/File[C:/ProgramData/PuppetL
abs/puppet/etc/puppet.conf]/content: content changed '{md5}01118058cead7a8411246cb98d1c6c08' to '{md5}8a19b5b37ea171f830
965fef43b138f1'
Notice: /Stage[main]/Puppet::Config/Concat[C:/ProgramData/PuppetLabs/puppet/etc/puppet.conf]/File[C:/ProgramData/PuppetL
abs/puppet/etc/puppet.conf]/mode: mode changed '2000770' to '0644'
Notice: /Stage[main]/Puppet::Config/File[C:/ProgramData/PuppetLabs/puppet/etc/auth.conf]/content: content changed '{md5}
6a6cfd614f082c23e4f75a74f1004f45' to '{md5}cc93aebac017ef93e73f6b1ca066f435'
Notice: /Stage[main]/Puppet::Agent::Service/Service[puppet]: Triggered 'refresh' from 1 events
Notice: Applied catalog in 19.44 seconds
PS C:\programdata\PuppetLabs\puppet\etc\modules\puppet>
                                                               
```

roidelapluie:
Guys, can all this windows migration from 3 to 4 be tackled on separated PR?

mmoll:
I would be OK with that, it sounds like only the 3 -> 4 update on Windows is running into problem and I'm not sure we would support that anyway...

genebean:
I have no issues with merging. I posted the errors I saw just in case they are something that we should be concerned about but don't see it as a reason to hold this up.

mmoll:
:+1:

roidelapluie:
:+1:

ekohl:
I plan to test this tomorrow.

genebean:
@mmoll & @ekohl - not to hold this one up, but do we / I / you all need to do anything related to those errors I posted? I am not quite sure what the root cause is but would be happy to help if given a little direction.

ekohl:
I tried upgrading a foreman server on EL7. It was a bit painful since you need to manually copy/move some files to the new puppetlabs locations. @genebean could it be that some directory has moved from puppet 3 to puppet 4 so it can no longer find it?

We also need to fix puppet-foreman_proxy which does some incorrect SSL-directory things in params.pp. Other than that I think it looks sane.

@domcleal mind giving this a look as well?

roidelapluie:
Can we maybe target an upgrade path in another pull request?

I suspect that some ruby migration magic will be needed to go from puppet 3 to puppet 4...

mmoll:
As said, I'm unsure if we want to tackle the whole 3 -> 4 migration in this module at all (if yes, we might want to bowrrorow code or even just use puppetlabs-puppet_agent.

My tests were also positive, but normally we want to have at least two positive voices and for big changes like this one ideally three, so after me and @ekohl now @domcleal will review and if he's positive, merge.

I already want to thank @mcanevet and @roidelapluie for tackling that, as this is indeed a big step forward to Puppet 4 support for us!

ekohl:
I agree we don't have to support automatic upgrades in this PR, but like @mmoll said, this is a big change which I want to carefully review.

genebean:
@ekohl From what I understand there have not been any directory changes on the Windows-side as part of the transition from Puppet 3 to 4 as it was already an AIO package. @ferventcoder - that's correct, isn't it?

genebean:
@mmoll - I think just using puppetlabs-puppet_agent would be a great idea since it is the official way to migrate... assuming nothing in it conflicts with this module (which it might). It's ideas are certainly worth a look.

ferventcoder:
>From what I understand there have not been any directory changes on the Windows-side as part of the transition from Puppet 3 to 4 as it was already an AIO package. ferventcoder - that's correct, isn't it?

@genebean That's correct. There are some folders added, but other than that, there is not much of a transition on Windows. However, FOSS now has MCO as part of the package as well with Windows as part of v4.

Most of the changes are documented here - https://github.com/puppetlabs/puppet-specifications/blob/master/file_paths.md#puppet-agent-windows

> I think just using puppetlabs-puppet_agent would be a great idea since it is the official way to migrate... assuming nothing in it conflicts with this module (which it might). It's ideas are certainly worth a look.

I'm assuming you've already looked at the requirements for this? It's assumed you are on at least 3.8 and you have switched on future parser.

roidelapluie:
This pr will probably trigger a major version of this module, won't it? In this case it would make sense to first release a minor version with the pending bugfixes (if any) before merging.

ekohl:
> This pr will probably trigger a major version of this module, won't it? In this case it would make sense to first release a minor version with the pending bugfixes (if any) before merging.

I agree that makes sense.

genebean:
Thanks @ferventcoder for the info

> I'm assuming you've already looked at the requirements for this? It's assumed you are on at least 3.8 and you have switched on future parser.

I had seen this but had forgotten and that will certainly make a difference here.

domcleal:
Looks fine to me too, thanks for your efforts on this!

I don't think we need a major bump as this can be released in a minor version, but I'd be happy to cut a new minor release with what's in master now so we can merge & stabilise this new feature before cutting another minor.

It needs rebasing before it can be merged.

roidelapluie:
@domcleal rebased. Waiting for green tests.

Can you maybe review the commit message as well?

mmoll:
Tests fail for Puppet 4, the fail for Puppet 2.7 is expected (problem with current stdlib master)

Regarding the commit message, I'd like to have something like "Fix directory ownership without puppet user" and "Adjust tests", as there are no real regressions. ;)

Per semver no new major version is required, so I'd go for a minor bump, or are there deprecated parameters waiting for removal?

roidelapluie:
I have already pushed a change to address failing tests

I think more than a minor dump is needed because wi kind of break the API: package['puppet'] is no longer guarantee to be present. People relying on that might be surprised, what do you think? I know it is an internal change but it is probably used in 3rd parties modules/roles/profiles?

domcleal:
Yeah, I'm not terribly sympathetic to the Package[puppet] argument, as it's an internal matter - and in the existing supported uses, it's going to remain the same anyway.  It'll only change when using AIO, which is a change anyway.

There still seems to be one test failure on Puppet 4.0.

roidelapluie:
Let's see now

roidelapluie:
I have improved the tests

roidelapluie:
:+1:

mmoll:
:+1:

domcleal:
Merged as 168ba94, 00073c0.

Thanks to everybody their help on this topic, especially @mcanevet, @roidelapluie and @genebean!

PR #291, opened 2015-07-23
Set explicit permissions and ownership

ekohl:
This matches the puppet default and it reduces differences if you set a
global default in site.pp.

### DIFF
--- a/manifests/server/config.pp
+++ b/manifests/server/config.pp
@@ -34,16 +34,21 @@ class puppet::server::config inherits puppet::config {
   file { "${puppet::vardir}/reports":
     ensure => directory,
     owner  => $puppet::server_user,
+    group  => $puppet::server_group,
+    mode   => '0750',
   }
 
   ## SSL and CA configuration
   # Open read permissions to private keys to puppet group for foreman, proxy etc.
   file { "${puppet::server_ssl_dir}/private_keys":
-    group => $puppet::server_group,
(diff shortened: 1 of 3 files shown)


manifests/server/config.pp
```diff
@@ -33,17 +33,21 @@
...
+    mode   => '0750',
     owner  => $puppet::server_user,
   }
 
   ## SSL and CA configuration
   # Open read permissions to private keys to puppet group for foreman, proxy etc.
   file { "${puppet::server_ssl_dir}/private_keys":
-    group => $puppet::server_group,
-    mode  => '0750',
+    ensure => directory,
```
mmoll: nice one :)
:file_folder:

mmoll:
:+1:

manifests/server/config.pp
```diff
@@ -33,17 +33,21 @@
 
   file { "${puppet::vardir}/reports":
     ensure => directory,
+    mode   => '0750',
     owner  => $puppet::server_user,
```
ekohl: I wonder if this also needs a group. Puppet doesn't enforce it I think, but can't hurt.

mmoll: yeah, I'd also say add it :+1:

ekohl: Done & rebased

mmoll:
bedankt @ekohl!

PR #292, opened 2015-07-24
Remove verify_*, enable param docs linting

domcleal:


.sync.yml
```diff
@@ -1,4 +1 @@
 ---
```
ekohl: Remove the entire file?

domcleal: Sorry, I forgot to add the new lint pattern here.  Fixed!

ekohl:
:+1: on green tests.

mmoll:
thanks @domcleal!

PR #294, opened 2015-07-31
relay passenger_max_pool to apache::mod::passenger class

michakrause:


ekohl:
I'd rather set this on the vhost, but maybe we need some custom templating for this. (I also know that I introduced this parameter a long time ago and then later broke it myself when I rewrote the module to puppetlabs-apache)

mmoll:

michakrause:
I don't want to spend any more time on this, as we are also planing to upgrade to Puppet 4. I have no problems using our fork in the meantime.

PR #296, opened 2015-08-01
Added $manage_packages to control if module install packages

hickey:
This allows one to use the puppet module but not have it install any of the puppet modules. I am championing this as we like the module, but can not handle having packages installed from an OS source as our puppet components are newer versions installed from gems. 

The default behavior out of the box is still to allow the module to manage package installation. This way existing functionality is not broken and still handles according to any existing documentation.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -25,6 +25,11 @@
 #
 # $sharedir::                      Override the system data directory.
 #
+# $manage_packages::               Should this module install packages or not.
+#                                  Can also install only server packages with value
+#                                  of 'server' or only agent packages with 'agent'.
+#                                  Defaults to true
+#
 # $package_provider::              The provider used to install the agent.
 #                                  Defaults to chocolatey on Windows
 #                                  Defaults to undef elsewhere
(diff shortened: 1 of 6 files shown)


manifests/init.pp
```diff
@@ -537,11 +543,15 @@
 
   validate_re($server_implementation, '^(master|puppetserver)$')
   validate_re($server_parser, '^(current|future)$')
-
+
```
ekohl: Nitpick: trailing whitespace

ekohl:
I think it's a good change. Please fix the whitespace nitpick, fix the tests and squash the commits.

hickey:
OK. Fixed the problem I had. Actually was not the tests, but at least the tests exposed the problem so it could be corrected before it got released. Yea!!! 

Also fixed the extra whitespace in init.pp and squashed the commits into a single commit. 

Thanks.

manifests/init.pp
```diff
@@ -537,11 +543,15 @@
...
-
+  
   if $server_environment_timeout {
     validate_re($server_environment_timeout, '^(unlimited|0|[0-9]+[smh]{1})$')
   }
 
+  if $manage_packages != true and $manage_packages != false {
+    validate_re($manage_packages, '^(server|agent)$')
+  }
+
```
ekohl: Nitpick: trailing whitespace.

ekohl:
:+1: the test failure is unrelated (stdlib master appears to have broken 2.7 compatibility). It'd be great if you could remove the two instances of trailing whitespace but I can also cherry-pick the commit and fix it there.

mmoll:
thanks @hickey!

PR #297, opened 2015-08-02
Correct $puppetrun_cmd on Linux

ekohl:
Regression introduced in f74e45d1dc8a96929f22733f39052a798aeff52f.


mmoll:
I'm leaning towards merging, even if tests on Puppet 2.7 are failing. I suppose the current stdlib master is not 2.7 compatible anymore?

ekohl:
I agree. We can pin stdlib in a separate PR. Most likely we need to do the same in foreman-installer.

mmoll:
bedankt @ekohl!

ekohl:
Geen dank :)

PR #298, opened 2015-08-03
use stdlib 4.7 in tests

mmoll:
for Puppet 2.7 compatibility


domcleal:
I'm a little hesitant with pinning things too much, but I think it's probably good to do in the short term until MODULES-2316 is fixed and then revert it again so we're aware of issues in master.  Sound sensible?

ekohl:
I share your hesitance. Do we know if puppetlabs intends to keep 2.7 support in stdlib?

domcleal:
I believe it was going to be dropped with 5.0, but I don't know when that will happen.  I'm surprised it hasn't already, to be honest.

Ideally we'd use the latest 4.x branch with 2.7 and master for everything else, but I don't know a way of doing that without some hacking around in spec_helper.

PR #300, opened 2015-08-05
Fix rspec-expectations should vs expect deprecation warning

vStone:
Fixes a deprecation warning:

```
Using `should` from rspec-expectations' old `:should` syntax without explicitly enabling the syntax is deprecated. Use the new `:expect` syntax or explicitly enable `:should` instead. Called from /tmp/puppet/modules/upstream/puppet/spec/spec_helper.rb:30:in `verify_concat_fragment_exact_contents'.
```

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -22,10 +22,10 @@ end
 
 def verify_concat_fragment_contents(subject, title, expected_lines)
   content = subject.resource('concat::fragment', title).send(:parameters)[:content]
-    (content.split("\n") & expected_lines).should == expected_lines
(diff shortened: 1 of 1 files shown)


domcleal:
Would you mind making the change over here instead?  https://github.com/theforeman/foreman-installer-modulesync/blob/master/moduleroot/spec/spec_helper.rb

ekohl:
I noticed this too, but puppetlabs_spec_helper still has the same deprecated syntax so I couldn't get rid of the warnings. Shouldn't this also require at least rspec 3.x?

vStone:
No longer valid since theforeman/foreman-installer-modulesync#16 got merged?

PR #305, opened 2015-08-27
allow autosign to be a boolean and add tests

mmckinst:
see pull request #301

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -71,11 +71,14 @@
 #                                  This setting can be a time interval in seconds
 #                                  (30 or 30s), minutes (30m), hours (6h), days (2d),
 #                                  or years (5y).
-# $autosign::                      Change location of autosign.conf or autosign
-#                                  script. If this is set to a script, make sure
-#                                  that script considers the content of
-#                                  autosign.conf, as it might break Foreman
-#                                  functionality if it doesn't.
+#
+# $autosign::                      Enable or disable autosign or change location
+#                                  of autosign.conf or autosign script.  If this
+#                                  is set to a script, make sure that script
+#                                  considers the content of autosign.conf, as it
+#                                  might break Foreman functionality if it
+#                                  doesn't. If this is set to a boolean, it can
+#                                  be cast as a string or a boolean.
 #                                  type:string
 #
 # $runinterval::                   Set up the interval (in seconds) to run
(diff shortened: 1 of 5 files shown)


manifests/init.pp
```diff
@@ -72,6 +72,14 @@
 #                                  (30 or 30s), minutes (30m), hours (6h), days (2d),
 #                                  or years (5y).
 #
+# $autosign::                      Eanble or disable autosign or change location
+#                                  of autosign.conf or autosign script.  If this
+#                                  is set to a script, make sure that script
+#                                  considers the content of autosign.conf, as it
+#                                  might break Foreman functionality if it
+#                                  doesn't.
+#                                  type:string or boolean
```
mmoll: this doesn't work with kafo:
`/usr/lib/ruby/vendor_ruby/kafo/param_builder.rb:83:in 'const_defined?': wrong constant name String or boolean (NameError)`
@ares what's the best to do in such a case?

mmoll:
:+1: pending kafo functionality

manifests/init.pp
ares: I suppose this should be a new type (stroolean?) but IIRC we tried to avoid having parameters like this, my preference would be adding new parameter or in the worst case keeping it as string and treat 'true' and 'false' strings explicitely

manifests/init.pp
```diff
@@ -72,6 +72,14 @@
 #                                  (30 or 30s), minutes (30m), hours (6h), days (2d),
 #                                  or years (5y).
 #
+# $autosign::                      Eanble or disable autosign or change location
```
ares: s/Eanble/Enable/

manifests/init.pp
ekohl: I think we can just keep it ```type:string``` since you can enter the string ```true``` or ```false``` and it will work.

ares: and since string is default, you could omit the type completely

mmoll:
@mmckinst ping?

mmckinst:
I didn't realize this was waiting on me.

If kafo can't do multiple types, I'm fine with it just being a string, but maybe it should be documented you can put a true/falase as the string? 

Other than that, pull request #301 does everything this does. I can cherry pick out my tests and send them as a different pull request if you want.

mmoll:
Yes, please fix that typo in the comment and delete the type-hint (and squash that into your last commit), then this is ready to go, IMHO.

mmckinst:
@mmoll should be good to go

mmoll:
Merged, minus the type hint, as string is default anyway.
Thanks @ofalk and @mmckinst!

PR #306, opened 2015-08-28
splay needs to be set in [main] section of puppet.conf, not in [agent]

mmckinst:
If you set splay and splaylimit in the [agent] section of puppet.conf, it will not work. It needs to be set in the [main] section of puppet.conf. See below for proof and also see 

On a server where splay is set in agent, it does not work:
```
[root@localhost ~]# cat puppet-splay-agent.conf
[agent]
  splay = true
  splaylimit = 600s
[root@localhost ~]# 
[root@localhost ~]# /opt/puppetlabs/bin/puppet config --config ~/puppet-splay-agent.conf  print  | grep splay
config = /root/puppet-splay-agent.conf
splaylimit = 1800
splay = false
[root@localhost ~]# 
```

On a server where splay is set in main, it works.
```
[root@localhost ~]# cat puppet-splay-main.conf
[main]
  splay = true
  splaylimit = 600s
[root@localhost ~]# 
[root@localhost ~]# /opt/puppetlabs/bin/puppet config --config ~/puppet-splay-main.conf  print  | grep splay
config = /root/puppet-splay-main.conf
splaylimit = 600
splay = true
[root@localhost ~]# 
```

I tested the above on puppet 4.1.0 and 3.3.0.

### DIFF
--- a/spec/classes/puppet_config_spec.rb
+++ b/spec/classes/puppet_config_spec.rb
@@ -89,6 +89,20 @@ describe 'puppet::config' do
       end
     end
 
+    context 'when splay = true and splaylimit = 60s' do
+      let :pre_condition do
+        'class {"::puppet": splay => true, splaylimit => \'60s\'}'
+      end
+
+      it 'should contain puppet.conf [main] with splay and splaylimit' do
+        verify_concat_fragment_contents(catalogue, 'puppet.conf+10-main', [
+          '[main]',
+          '    splay      = true',
+          '    splaylimit = 60s',
+        ])
+      end
+    end
+
     context "when dns_alt_names => ['foo','bar']" do
       let :pre_condition do
         "class { 'puppet': dns_alt_names => ['foo','bar'] }"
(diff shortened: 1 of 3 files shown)


mmckinst:
Actually I was wrong on this. ```puppet config``` by default prints out the [main] section per the man page of puppet-config. When puppet agent is actually run, it will pick up the setting in the [agent] section. If you want to see the config the agent would use you need to add the ```--section agent``` parameter.

See below for confirmation, using same files as before.

```
[root@localhost ~]$ puppet config print --config puppet-splay-agent.conf --section agent | grep splay
config = /home/mmckinst/puppet-splay-agent.conf
splaylimit = 600
splay = true
[root@localhost ~]$ 
[root@localhost ~]$ puppet config print --config puppet-splay-main.conf --section agent | grep splay
config = /home/mmckinst/puppet-splay-main.conf
splaylimit = 600
splay = true
[root@localhost ~]$ 
```

Issue #307, opened 2015-08-29
Missing puppet user/group when installing JVM-based puppetserver on CentOS 7 vbox


westbywest:
I tried running this manifest in standalone on a XXX virtualbox guest:
```
  class { '::puppet': 
    runmode => 'service',
    server => true,
    server_jvm_min_heap_size => '2G', 
    server_implementation => 'puppetserver',
    server_foreman        => false,
    server_reports        => 'store',
    server_external_nodes => '',
    server_environments   => [],
  }
```
Execution fails on missing puppet user/group:
```
Error: Could not find user puppet
Error: /Stage[main]/Puppet::Config/File[/etc/puppetlabs/puppet]/owner: change from root to puppet failed: Could not find user puppet
Error: Could not find group puppet
Error: /Stage[main]/Puppet::Config/File[/etc/puppetlabs/puppet]/group: change from root to puppet failed: Could not find group puppet
```
This is the puppet release rpm and agent I was using:
```
puppetlabs-release-pc1-0.9.2-1.el7.noarch
puppet-agent-1.2.2-1.el7.x86_64
```
The problem doesn't occur if I create a dummy puppet user/group before running the manifest:
```
useradd -d /tmp/puppet -s /bin/nologin puppet
```

mmoll:
When the PR for Puppet AIO packages was done, only the agent side was in scope, so only puppetserver 1.x (not 2.x from the PC1 repo) is tested at all and for some configurations there are still problems like #275.

However, while testing your manifest on a fresh CentOS 7.1 VM the puppetserver package (which is bringing the puppet user) was installed on the first run and the second run did suceed then (even the agent worked with the new master!), so I guess it's sufficient to change the resource order in this module to fix this specific problem (without losing compatibility with old packages of course).

@westbywest would you be willing to provide a PR for that?

Issue #302, opened 2015-08-14
Changing owner


roidelapluie:
Notice: /Stage[main]/Puppet::Config/File[/etc/puppetlabs/puppet]/owner: owner changed 'root' to 'puppet'
Notice: /Stage[main]/Puppet::Config/File[/etc/puppetlabs/puppet]/group: group changed 'root' to 'puppet'

With the new AIO package.

Should we do this or keep the package default?

ekohl:
I would prefer to have as little changes as possible, so I think this is something we should fix.

mmoll:
should be fixed with #308 also

PR #308, opened 2015-08-31
Declare puppet user/group resource dependencies.

westbywest:
Declare dependencies for a puppet user/group to satisfy ownership parameters for $puppet_dir.  Specifically, ensure user and group resources are in place (if defined) before trying to set ownership on the path $puppet_dir.  Seeks to address #307.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -25,19 +25,10 @@ class puppet::config(
     order   => '10',
   }
 
-  $dir_owner = $::puppet::server ? {
-    true    => $::puppet::user,
-    default => $::puppet::dir_owner,
-  }
-  $dir_group = $::puppet::server ? {
-    true    => $::puppet::group,
-    default => $::puppet::dir_group,
-  }
-
   file { $puppet_dir:
     ensure => directory,
-    owner  => $dir_owner,
-    group  => $dir_group,
+    owner  => $::puppet::dir_owner,
+    group  => $::puppet::dir_group,
   } ->
   case $::osfamily {
     'Windows': {


mmoll:
This module never did manage the puppet user, as the puppet packages brought it. The only thing that's managed is it's shell when using git integration (manifests/server/install.pp).

The thing is that with puppet 4 AIO packages, the puppet user is created _only_ when installing the puppetserver package, so the run order has to get changed somehow to have the server package installed before doing anything with the puppet user when server is enabled.

westbywest:
There is already the standing issue #302 about changing ownership of $puppet_dir to values different than what the v4 AIO puppetserver package deploys.  If the module kept $puppet_dir ownership as is, the dependency error in this issue also wouldn't occur (at least with the AIO packages).  Would gating the ownership parameters for the $puppet_dir path based on the value of $aio_package be a more elegant approach?  I.e., if $aio_package = true, then set user and group ownership of $puppet_dir to undef?  This would avoid having the module manage the puppet/user group with explicit resource declarations.

mmoll:
I wonder if just using $user/$group is also OK for that dir in all versions? The problem with setting it to `undef` is that probably some defaults come implicitely into play, setting it also to `root` for UNIX systems.

mmoll:
Could that work?
```diff
diff --git a/manifests/config.pp b/manifests/config.pp
index dd436c8..e13069d 100644
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -25,19 +25,10 @@ class puppet::config(
     order   => '10',
   }
 
-  $dir_owner = $::puppet::server ? {
-    true    => $::puppet::user,
-    default => $::puppet::dir_owner,
-  }
-  $dir_group = $::puppet::server ? {
-    true    => $::puppet::group,
-    default => $::puppet::dir_group,
-  }
-
   file { $puppet_dir:
     ensure => directory,
-    owner  => $dir_owner,
-    group  => $dir_group,
+    owner  => $::puppet::dir_owner,
+    group  => $::puppet::dir_group,
   } ->
   case $::osfamily {
     'Windows': {
```

westbywest:
Changing the owner/group params as recommended triggers build failures under FreeBSD, with the cause not being clear.  Works fine under RHEL.

mmoll:
what kind of build failures?

westbywest:
Quoting from travis CI:
```
  on a FreeBSD family OS
    with default parameters
      should contain auth.conf (FAILED - 1)
      should contain puppet.conf [main] (FAILED - 2)
```
FreeBSD is the only failing OS.

mmoll:
thanks, checking...

mmoll:
seems like fallout from rspec-puppet-facts switching to facterdb...

manifests/config.pp
```diff
@@ -36,8 +36,8 @@
```
mmoll: please remove the then uneeded blocks of $dir_owner and $dir_group above

mmoll:
If you rebase onto current master, FreeBSD tests are fixed. Please also squash your commits together into one.

westbywest:
Rebased and squashed.  Tests underway (or possibly hung?).

mmoll:
Plase see my line comment to remove the block
```
   $dir_owner = $::puppet::server ? {
     true    => $::puppet::user,
     default => $::puppet::dir_owner,
   }
   $dir_group = $::puppet::server ? {
     true    => $::puppet::group,
     default => $::puppet::dir_group,
   }
```

westbywest:
Rebased and resquashed,

domcleal:
So the only change in behaviour here is when using an AIO package, plus configuring the server, the /etc/puppetlabn/puppet directory will now be owned by root rather than puppet?

mmoll:
AFAICT, yes.

domcleal:
Thanks, I re-read #307 and it made more sense now.

Merged as 64c3fd0 with a new commit message to explain.  Thanks for the patch @westbywest!

PR #310, opened 2015-09-10
Fixing Passenger max_pool_size, adding more Passenger variables

sepulworld:
This change allows for more tuning options for Passenger setups.  The defaults for these new variables match what Passenger sets by default. 

The server_passenger_max_pool variable changes to match what Apache module expects, server_passenger_max_pool_size.

This change updates the Passenger tests to now validate the Apache passenger.conf

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -308,83 +308,104 @@
 #
 # === Advanced server parameters:
 #
-# $server_httpd_service::          Apache/httpd service name to notify
-#                                  on configuration changes. Defaults
-#                                  to 'httpd' based on the default
-#                                  apache module included with foreman-installer.
+# $server_httpd_service::              Apache/httpd service name to notify
+#                                      on configuration changes. Defaults
+#                                      to 'httpd' based on the default
+#                                      apache module included with foreman-installer.
 #
-# $server_service_fallback::       If passenger is not used, do we want to fallback
-#                                  to using the puppetmaster service? Set to false
-#                                  if you disabled passenger and you do NOT want to
-#                                  use the puppetmaster service. Defaults to true.
-#                                  type:boolean
+# $server_service_fallback::           If passenger is not used, do we want to fallback
+#                                      to using the puppetmaster service? Set to false
+#                                      if you disabled passenger and you do NOT want to
+#                                      use the puppetmaster service. Defaults to true.
+#                                      type:boolean
 #
-# $server_passenger_max_pool::     The PassengerMaxPoolSize parameter. If your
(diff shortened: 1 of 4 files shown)


sepulworld:
Some tests failing:

https://travis-ci.org/theforeman/puppet-puppet/jobs/79602162
```
Failure/Error: is_expected.to contain_file('passenger.conf').with({
       expected that the catalogue would contain File[passenger.conf] with path set to "/etc/apache2/mods-available/passenger.conf" but it is set to "/etc/httpd/conf.d/passenger_extra.conf"
```

https://travis-ci.org/theforeman/puppet-puppet/jobs/79602164
```
Failure/Error: should contain_apache__mod('passenger')
     Puppet::PreformattedError:
       Evaluation Error: Error while evaluating a Function Call, pick(): must receive at least one non empty value at /home/travis/build/theforeman/puppet-puppet/spec/fixtures/modules/puppet/manifests/server/install.pp:13:23 on node testing-worker-linux-docker-bf314f56-3216-linux-4.prod.travis-ci.org
```
Ill take a look...

mmoll:
This is similar to #294, please see there for a general comment. This module is used at least in foreman-installer together with our foreman puppet module, which is also configuring a passenger vhost.

sepulworld:
Thanks for the feedback and reference.  Should we move discussion to #294?  I reviewed puppet-foreman and foreman-installer and I see the dependencies now. It appears foreman-puppet manages Passenger configs in a custom Vhost, while in this module we just rely on the inclusion of apache::mod::passenger.  The server_passenger_max_pool variable defined in this module doesn't do anything presently.

I believe we should isolate the Passenger options to one module, and use something like the following in the the module where we don't expose tuning options.

```
unless defined('apache::mod::passenger') {
  include ::apache::mod::passenger
}
```

The use case for exposing more Passenger tuning options is as follows:
1. I have a large set of modules with multiple environments.  Running the 'classes import' api function times out easily if I don't have the right Passenger settings in place. I narrowed down the issue to Passenger timing out requests. With the right Passenger settings this issue is avoided.
2. Catalog compilation times improve with Passenger tuned properly.  Benching with 2000+ system environment. 

Or I could focus on getting Puppet Server (4.x) running instead :)

mmoll:
I don't have a clear opinion on that, I think it will be impossible to really solve this without some ugly Puppet tricks or using custom templates.

If you want to help with Puppet 4, that would be much appreciated, see i.e. theforeman/puppet-foreman#332 and theforeman/smart-proxy#317 - foreman-installer/kafo is still an unsolved topic in the whole equation.

mmoll:

PR #312, opened 2015-09-23
Adds 4 new class parameters to puppet main class

jlyheden:
This pull allows setting the default_manifest configurable in puppet.conf as well as optionally managing the file and its contents. It also allows ssldir to not be set in the puppet.conf master section which is necessary when provisioning masters without the CA role (https://docs.puppetlabs.com/guides/scaling_multiple_masters.html#before-running-puppet-agent-or-puppet-master)

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -4,395 +4,411 @@
 #
 # === Parameters:
 #
-# $version::                       Specify a specific version of a package to
-#                                  install. The version should be the exact
-#                                  match for your distro.
-#                                  You can also use certain values like 'latest'.
+# $version::                          Specify a specific version of a package to
+#                                     install. The version should be the exact
+#                                     match for your distro.
+#                                     You can also use certain values like 'latest'.
 #
-# $user::                          Override the name of the puppet user.
+# $user::                             Override the name of the puppet user.
 #
-# $group::                         Override the name of the puppet group.
+# $group::                            Override the name of the puppet group.
 #
-# $dir::                           Override the puppet directory.
+# $dir::                              Override the puppet directory.
 #
-# $codedir::                       Override the puppet code directory.
+# $codedir::                          Override the puppet code directory.
 #
-# $vardir::                        Override the puppet var directory.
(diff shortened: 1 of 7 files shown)


mmoll:
I didn't test this but it's looking correct after reading the Puppet docs about the config entries.

Please adjust the indentation of all the other vars in `init.pp` and squash your commits together.

mmoll:
please squash all your commits together into one (if it's too complicated, we can do this for you when merging), :+1: otherwise

jlyheden:
My git-fu failed me so I ended up just creating a unified diff of all my changes, cloned upstream again, applied the patch and then force push it to my forked repo. Hope it is good enough

manifests/init.pp
```diff
@@ -4,395 +4,411 @@
...
+#                                     type:string
+#
+# $server_default_manifest_content::  A string to set the content of the default_manifest
+#                                     If set to '' it will not manage the file
+#                                     type:string
+#
+# $server_ssl_dir_manage::            Toggle if ssl_dir should be added to the [master]
+#                                     configuration section. This is necessary to
+#                                     disable in case CA is delegated to a separate instance
+#                                     type:boolean
```
ekohl: Please validate this with ```validate_boolean```.

manifests/init.pp
```diff
@@ -4,395 +4,411 @@
...
-#                                  default to ['$1']
-#                                  type:array
+# $auth_allowed::                     An array of authenticated nodes allowed to
+#                                     access all catalog and node endpoints.
+#                                     default to ['$1']
+#                                     type:array
+#
+# $server_default_manifest::          Toggle if default_manifest setting should
+#                                     be added to the [main] section
+#                                     type:boolean
```
ekohl: Please validate this with ```validate_boolean```.

jlyheden:
Let me know if there is anything else missing that needs to be fixed, would be great to have this merged so we can start using the upstream module rather than our fork

ekohl:
Thanks!

PR #315, opened 2015-09-29
workaround for very slow rspec run

pall-valmundsson:
It seems that a large `operatingsystem_support` section in metadata.json takes a really long time to get parsed by rspec-puppet-facts.  Since tests are only run on one instance of a supported OS for most of the tests this passes an opts hash to `on_supported_os` which limits parsing and speeds up the rspec run a lot.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -13,7 +13,9 @@ end
 
 # Workaround for slow rspec-puppet-facts
 def supported_os_opts
-    { :supported_os => [ { "operatingsystem" => "CentOS", "operatingsystemrelease" => [ "6" ] } ] }
+    { :supported_os => [ { "operatingsystem" => "CentOS", "operatingsystemrelease" => [ "6" ] },
+                         { "operatingsystem" => "FreeBSD", "operatingsystemrelease" => [ "10" ] }
+    ] }
 end
 
 def get_content(subject, title)
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -11,6 +11,11 @@ class Undef
   def inspect; 'undef'; end
 end
 
+# Workaround for slow rspec-puppet-facts
+def supported_os_opts
+    { :supported_os => [ { "operatingsystem" => "CentOS", "operatingsystemrelease" => [ "6" ] } ] }
+end
+
 def get_content(subject, title)
   content = subject.resource('file', title).send(:parameters)[:content]
   content.split(/\n/).reject { |line| line =~ /(^#|^$|^\s+#)/ }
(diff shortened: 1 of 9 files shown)


pall-valmundsson:
Comparison travis runs; [without workaround](https://travis-ci.org/theforeman/puppet-puppet/builds/82775369) and [with workaround](https://travis-ci.org/theforeman/puppet-puppet/builds/82818081) (workaround run not finished when I'm posting this :).

My desktop went from me exiting the unfinished test run after around 40 minutes down to 100 seconds.

ekohl:
This kind of goes against the spirit of full coverage. It is an interesting thought to have a quick-test option which for every osfamily tests the latest supported OS. We could then run travis in fail fast mode where it stops if once of the tests fails. Then you get much faster results for failing tests and yet get full coverage where they pass.

pall-valmundsson:
Yes, I agree with regards to the full coverage. But notice that I'm not reducing the test matrix, it was already only running for a specific os release.

ekohl:
Oh right, that is interesting. Awesome idea!

Maybe that indicates an inefficiency in rspec-puppet-facts. Maybe that should cache the on_supported_os in some immutable datastructure so it doesn't need to parse it every time. Another option would be to have the option to retrieve once specific OS easily. @mcanevet any idea insights?

spec/spec_helper.rb
```diff
@@ -11,6 +11,13 @@ class Undef
   def inspect; 'undef'; end
 end
 
+# Workaround for slow rspec-puppet-facts
+def supported_os_opts
```
ekohl: The name threw me off here. Maybe a better name would be ```common_supported_os_opts``` with a note describing that this is for quick testing of a single OS.

pall-valmundsson:
This module is on the large end w.r.t. OS support so others may not have hit this issue. I tried profiling the test run (first time profiling ruby) and it looks like the issue is because of very many calls to JGrep inside facterdb (inside rspec-puppet-facts) which is amplified by having a large matrix of supported OSes in metadata.json.

spec/spec_helper.rb
pall-valmundsson: Or even `on_supported_os_opts` as it's the options hash for the `on_supported_os` method.

This should not be a permanent workaround, the underlying issue should be fixed somehow.

pall-valmundsson:
Maybe a cleaner way would be to make the helper method return the cached result of calling `on_supported_os` which could be used inside all the specs. But that's almost like something that should be implemented in rspec-puppet-facts itself, transparently for end users.

ekohl:
I agree that the better fix is in rspec-puppet-facts itself, but we can have a workaround in our tests until then. I'm also wondering what @domcleal thinks about this.

mcanevet:
@pall-valmundsson @ekohl very interesting. I know that facterdb is not efficient at all. Could you please try with this patch https://github.com/camptocamp/facterdb/pull/12 as it should reduce the number of jgrep calls.

pall-valmundsson:
@mcanevet I'm getting the same error as the test run for that pull:
`undefined local variable or method ```new_filter' for FacterDB:Module`

pall-valmundsson:
...fixed it. It isn't any faster though. I'm starting to think the time is actually spent inside JGrep while parsing the filter.

PR #317, opened 2015-10-02
rspec speedup demo

pall-valmundsson:
`let`s are not shared between examples so `on_supported_os` is re-evaluated for each example, which is inefficient. By changing the statement into an `.each` iterator the method is only evaluated once per run. An added bonus is "full coverage" but that can be trimmed down to current functionality to skip (or run only) selected OSes.

I hacked an example together that shows the decreased runtime; when testing the whole matrix a lot of tests fail on various platforms so I only enabled the currently tested platforms. The example also has incorrect indentation and code style as I didn't wan't to mess with whitespace changes, it's unmerge-able as is.

I can work up a merge-able commit if you're interested. I would begin by only picking the platforms that are currently being tested, as the tests actually fail for some others, but going forward they can be enabled and the tests and/or manifests can be fixed. There are also inconsistencies in how the test files are organized that could also be cleaned up.

Are you interested?

I'm closing #315 in favor of this one.

### DIFF
--- a/spec/classes/puppet_agent_install_spec.rb
+++ b/spec/classes/puppet_agent_install_spec.rb
@@ -8,7 +8,12 @@ describe 'puppet::agent::install' do
     additional_facts = {:rubysitedir => '/opt/puppetlabs/puppet/lib/ruby/site_ruby/2.1.0'}
   end
 
-  let :common_centos_facts do on_supported_os['centos-6-x86_64'].merge({
+on_supported_os.each do |os, os_facts|
+  next if not os == "centos-6-x86_64"
+  context "on #{os}" do
+
+  let (:common_centos_facts) do
+    os_facts.merge({
     :concat_basedir => '/nonexistant',
     :puppetversion  => Puppet.version,
   }).merge(additional_facts) end
@@ -167,4 +172,6 @@ describe 'puppet::agent::install' do
       end
     end
   end
+end
+end
 end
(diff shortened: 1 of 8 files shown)


ekohl:
This looks promising. I really appreciate any work to speed up tests and any cleanups possible.

mmoll:
what @ekohl says :100:

spec/classes/puppet_agent_install_spec.rb
```diff
@@ -8,7 +8,12 @@
     additional_facts = {:rubysitedir => '/opt/puppetlabs/puppet/lib/ruby/site_ruby/2.1.0'}
   end
 
-  let :common_centos_facts do on_supported_os['centos-6-x86_64'].merge({
+on_supported_os.each do |os, os_facts|
+  next if not os == "centos-6-x86_64"
```
ekohl: I'm wondering if we can generalize this. Maybe a variable or helper function in ```spec_helper``` to only test on the reduced set? The set can contain just 1 element for now, but I can see us testing this on CentOS latest and Debian latest to get coverage for both OS families.

ekohl:
This is a dramatic improvement and I'd be very interested in getting this merged. If you need anything, please ask. Other than that I'd be interested in a blog post describing how you got here. puppet-foreman tests take a long time as well and if we can get those down ...

spec/classes/puppet_agent_install_spec.rb
pall-valmundsson: Yes. I'm adding a helper that evaluates an environment variable that can be set to reduce the set of platforms tested.

mmoll:
I guess everything already landed in master?

mmoll:
Closing this, thanks again for bringing that in!

PR #318, opened 2015-10-06
refactor rspec tests for speed and coverage

pall-valmundsson:
Before this change most tests would only run on centos-6-x86_64 and take a long time to run since on_supported_os would be evaluated for each example.

This change reduces the time to run tests as on_supported_os is only evaluated once per file as well as testing all OSes that on_supported_os returns.  When enabling testing other platforms various failures emerged, mostly for FreeBSD, that were fixed by adjusting variables where they were either missing or needed to be adjusted for e.g. FreeBSD. Redundant examples were removed but since Windows is not currently supported by on_supported_os all Windows specific tests are
left untouched.

This increases the test examples from 431 to 2276 while lowering the current total execution time of the 431 examples. But since the execution time of the whole suite across the whole matrix of platforms still takes "too long" an option was introduced to reduce the set of tested platforms to a comma seperated list. The option can be set with the environment variable LIMIT_OS, e.g. `LIMIT_OS=centos-7-x86_64,ubuntu-14-x86_64.`

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -11,12 +11,21 @@ class Undef
   def inspect; 'undef'; end
 end
 
-# Running tests with the LIMIT_OS environment variable set
+# Running tests with the ONLY_OS environment variable set
 # limits the tested platforms to the specified values.
-# Example: LIMIT_OS=centos-7-x86_64,ubuntu-14-x86_64
-def limit_test_os
-  if ENV.key?('LIMIT_OS')
-    ENV['LIMIT_OS'].split(',')
+# Example: ONLY_OS=centos-7-x86_64,ubuntu-14-x86_64
+def only_test_os
+  if ENV.key?('ONLY_OS')
+    ENV['ONLY_OS'].split(',')
+  end
+end
+
+# Running tests with the EXCLUDE_OS environment variable set
+# limits the tested platforms to all but the specified values.
+# Example: EXCLUDE_OS=centos-7-x86_64,ubuntu-14-x86_64
+def exclude_test_os
+  if ENV.key?('EXCLUDE_OS')
+    ENV['EXCLUDE_OS'].split(',')
   end
 end
 
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -11,6 +11,15 @@ class Undef
   def inspect; 'undef'; end
 end
 
+# Running tests with the LIMIT_OS environment variable set
+# limits the tested platforms to the specified values.
+# Example: LIMIT_OS=centos-7-x86_64,ubuntu-14-x86_64
+def limit_test_os
+  if ENV.key?('LIMIT_OS')
+    ENV['LIMIT_OS'].split(',')
+  end
+end
+
 def get_content(subject, title)
   content = subject.resource('file', title).send(:parameters)[:content]
   content.split(/\n/).reject { |line| line =~ /(^#|^$|^\s+#)/ }
@@ -28,4 +37,20 @@ end
 def verify_concat_fragment_exact_contents(subject, title, expected_lines)
   content = subject.resource('concat::fragment', title).send(:parameters)[:content]
     content.split(/\n/).reject { |line| line =~ /(^#|^$|^\s+#)/ }.should == expected_lines
+end
+
+# See https://github.com/rodjek/rspec-puppet/issues/215
+# Without this patch the @@cache variable grows huge and causes memory usage issues
+# with a large amount of examples.
+module RSpec::Puppet
+  module Support
+    def build_catalog(*args)
+      if @@cache.has_key?(args)
+        @@cache[args]
+      else
+        @@cache = {}
+        @@cache[args] ||= self.build_catalog_without_cache(*args)
+      end
+    end
+  end
 end
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,10 +8,10 @@ rvm:
   - 1.8.7
 env:
   - PUPPET_VERSION=4.0 ONLY_OS=centos-7-x86_64,ubuntu-14-x86_64,freebsd-10-amd64
-  - PUPPET_VERSION=3.5 FUTURE_PARSER=yes TRUSTED_NODE_DATA=yes ONLY_OS=centos-7-x86_64,ubuntu-14-x86_64,freebsd-10-amd64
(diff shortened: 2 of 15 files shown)


spec/spec_helper.rb
```diff
@@ -29,3 +38,19 @@ def verify_concat_fragment_exact_contents(subject, title, expected_lines)
   content = subject.resource('concat::fragment', title).send(:parameters)[:content]
     content.split(/\n/).reject { |line| line =~ /(^#|^$|^\s+#)/ }.should == expected_lines
 end
+
+# See https://github.com/rodjek/rspec-puppet/issues/215
```
ekohl: Is this also the cause of our killed travis tests? If so, we should include this in our modulesync config.

spec/spec_helper.rb
```diff
@@ -11,6 +11,15 @@ class Undef
   def inspect; 'undef'; end
 end
 
+# Running tests with the LIMIT_OS environment variable set
+# limits the tested platforms to the specified values.
+# Example: LIMIT_OS=centos-7-x86_64,ubuntu-14-x86_64
```
ekohl: I think this would be good to document in the README.

ekohl:
Could you have a look at the travis failure?

Speaking of travis, would it be a good idea to do a fast config and add the fail fast option?

It's a lot to review, but I think it's a good change. I'll try to do a more thorough review soon.

spec/spec_helper.rb
pall-valmundsson: Not initially no. But when the example count increased the tests started to take a lot of RAM, around 3 GB.

pall-valmundsson:
You can review with this URL: https://github.com/theforeman/puppet-puppet/pull/318/files?w=1 which ignores whitespace changes.

And yes, I think some version of fail fast should be implemented. My guess is that it would be best to test on RHEL family, Debian family and then FreeBSD.

pall-valmundsson:
I'll look at the failures as well. It's interesting to see how much faster newer versions of Puppet and Ruby are.

pall-valmundsson:
Ok. The failures are currently:
* puppet master (not puppetserver) installation on freebsd with ruby 1.8.7. It complains about pick needing a value, the supplied value is `[]`. Which works on newer ruby for some reason.
* some failures on puppet 2.7. I have to jump through some hoops to get a local test environment for 2.7 to track this down.

What is your stance on continuing to support Puppet 2.7 it was EOLd a year ago?

I'm not sure of the best way to address the FreeBSD issue.

ekohl:
Since EPEL6 still ships puppet 2.7, we do support it. There have been discussions on moving to a newer puppet through SCL, but that's not an easy task. It will become harder with time so at some point we might have to require 3+ but we try to hold that off while we can.

That said, it isn't useful to support 2.7 + ruby 1.8.7 on FreeBSD. I wonder if we could remove that from our testing matrix somehow.

pall-valmundsson:
2.7+1.8.7+FreeBSD isn't being tested in my .travis.yml version. Only ubuntu-12-x86_64 and centos-6-x86_64 are tested.

The FreeBSD issue seems to be stdlib + 1.8.7 + `pick(undef, [])`.
Is the old puppet master supported somehow for FreeBSD? The code in question is lines 8 and 13 [here](https://github.com/theforeman/puppet-puppet/blob/master/manifests/server/install.pp#L4-L19).

I understand the EPEL issue being exactly that admin myself but quite a while ago migrated to the PL repos. Most other modules have moved on from 2.7 and it might be a valid argument that if you're still on 2.7 you use a version of the module that supports/supported 2.7. Stdlib, for example, has a support matrix: https://github.com/puppetlabs/puppetlabs-stdlib#version-compatibility.

But the 2.7 issues can be fixed, it's not a major issue. More of a question about your stance in the matter.

mmoll:
I guess FreeBSD can just get filtered out when `rubyversion` is 1.8.7

mmoll:
FreeBSD support for this module came in quite late, when Ruby 2.0 already the default and way after Puppet 3.x came out, so some problems might not be visible at all in real world setups...

pall-valmundsson:
That's what I was wondering, is there any FreeBSD 9 or 10 host out there with ruby 1.8.7? And if so, are newer ruby packages a simple whatever `yum update` is on FreeBSD?

mmoll:
Per freebsd/freebsd-ports@2bd6529: any FreeBSD system that would still be on 1.8 will be really old and bad managed, so that's nothing we have to care about, and yes, the update would be one command.

pall-valmundsson:
The last failures were caused by a) directory environment config being tested on 2.7, which is wrong and b) different rack config.ru templates are applied to 2.X.

This should now pass the tests.

pall-valmundsson:
Travis seems to be flipping... could you rerun/restart the tests sometime?

mmoll:
restarted the tests

pall-valmundsson:
Ah, I had old fixtures in my clone. https://github.com/theforeman/puppet-foreman/commit/6576e9efc39f632d1347211feea515307f513f56 removed functions that fail the tests.

Two solutions:
a) depend on extlib and update .fixtures.yml
b) version the fixtures

mmoll:
please rebase onto current master, containing 870eeb2

pall-valmundsson:
One job has the message:
```
There was an error with the connection to the VM.
Your job will be requeued shortly.
```

But that was 20 hours ago...

mmoll:
rerunning that job now

mmoll:
This is looking fine to me :+1:

ekohl:
I agree it looks good. The spec_helper changes should be added to https://github.com/theforeman/foreman-installer-modulesync as well because they will be overwritten when we update.

pall-valmundsson:
Ok. Should I create a PR for the modulesync repo?

I've been thinking whether the include/exclude filters should be proposed for inclusion in puppet-rspec-facts itself. That would remove a bit of boilerplate here.

Then the monkey patching of rspec-puppet.... I don't really like it, but the current upstream implementation has bad side-effects for hundreds of different catalogs.

ekohl:
I think a PR to modulesync for the modulesynced files would be great.

pall-valmundsson:
Ok. PR submitted.

domcleal:
Looks fine to me.  I've not been through in great detail, but it looks like it makes sense.

mmoll:
merged, takk @pall-valmundsson!

PR #319, opened 2015-10-06
update example to use correct parameter names

brandonweeks:


mmoll:
merged, thanks @brandonweeks!

PR #322, opened 2015-10-13
reload the agent by sending a SIGHUP

martinpfeifer:
Puppet does recognize this on its own and reloads the config.
```
puppet-agent[25329]: Config file /etc/puppet/puppet.conf changed;
triggering re-parse of all config files.
```

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -272,8 +272,7 @@ class puppet::params {
       $osreleasemajor = regsubst($::operatingsystemrelease, '^(\d+)\..*$', '\1') # workaround for the possibly missing operatingsystemmajrelease
       $agent_restart_command = $osreleasemajor ? {
         '6'     => "/sbin/service ${service_name} reload",
-        #'7'     => "/usr/bin/systemctl reload-or-restart ${service_name}",
-        '7'     => "/bin/kill -HUP $(pgrep -f '/usr/bin/ruby /usr/bin/puppet agent')",
+        '7'     => "/usr/bin/systemctl kill --signal SIGHUP --kill-who=main ${service_name}",
         default => undef,
       }
       $unavailable_runmodes = $osreleasemajor ? {
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -272,7 +272,8 @@ class puppet::params {
       $osreleasemajor = regsubst($::operatingsystemrelease, '^(\d+)\..*$', '\1') # workaround for the possibly missing operatingsystemmajrelease
       $agent_restart_command = $osreleasemajor ? {
         '6'     => "/sbin/service ${service_name} reload",
-        '7'     => "/usr/bin/systemctl reload-or-restart ${service_name}",
+        #'7'     => "/usr/bin/systemctl reload-or-restart ${service_name}",
+        '7'     => "/bin/kill -HUP $(pgrep -f '/usr/bin/ruby /usr/bin/puppet agent')",
         default => undef,
       }
       $unavailable_runmodes = $osreleasemajor ? {
--- a/manifests/agent/service.pp
+++ b/manifests/agent/service.pp
@@ -34,19 +34,19 @@ class puppet::agent::service {
   anchor { 'puppet::agent::service_start': }
   anchor { 'puppet::agent::service_end': }
 
-  Anchor['puppet::agent::service_start'] ->
+  Anchor['puppet::agent::service_start'] ~>
   class { '::puppet::agent::service::daemon':
     enabled => $service_enabled,
   } ->
   Anchor['puppet::agent::service_end']
 
-  Anchor['puppet::agent::service_start'] ->
+  Anchor['puppet::agent::service_start'] ~>
   class { '::puppet::agent::service::systemd':
     enabled => $systemd_enabled,
   } ->
   Anchor['puppet::agent::service_end']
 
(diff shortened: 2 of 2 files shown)


timogoebel:
:thumbsup:

martinpfeifer:
Note: we did this to prevent puppet killing itself with the service restart when running in agent mode.

domcleal:
Reload normally works, but I'm pretty sure there are some settings that it can't reload.  I think this is present for the `listen` parameter, which enables the old Puppet run listening mode - this needs a full restart of the agent to start.

mmoll:
On which specific OS, Distro, Version are you seeing problems?

martinpfeifer:
It did occur the most on RHEL 6.6 Systems with Puppet version 3.3.2. We did see it on RHEL 7 Systems with Puppet version 3.7.4 as well. Since we disabled the self-restart we had no further problems.
Note: We deploy our systems using Foreman and Kickstart and we noticed that Puppet kills itself on the first puppetrun and then the systems setup were not complete.

I could implement this as a parameter so everyone can decide for themselves whether they want puppet to restart itself or not.

mmoll:
See #265, normally by reloading this shouldn't do a hard restart and mitigate the impact of this action. If that's turning out be be a problem now, I'm unsure what's be best solution:
* strip out the reload thing and restart always (doesn't feel like a good idea)
* don't care about restarting at all because Puppet is handling this itself (is it?)
* have a parameter (what would that control? boolean or some specific action?)

mmoll:
any new insights here? :eyeglasses:

timogoebel:
I think, I'd go for the parameter approach.
@martinpfeifer : Let's update this pr next week, shall we?

martinpfeifer:
I would like to do some further tests to see what is happening exactly. The problem seems not as straight forward as is first thought.

mmoll:
what's the status here?

martinpfeifer:
Sorry for the delay. Didn't have time to work on this lately :(
I'll try to summarize what i've learned.

The puppet agent supports the SIGHUP signal to restart itself. This is used on SysV Init systems like RHEL6.6 to reload the agent. On RHEL7, which is using systemd, the unit file shipped by puppetlabs does not specify a reload command (ExecReload=) so systemd just stops and starts the service on a reload request by sending a SIGTERM followed by a normal startup. I don't know if this is intended by Puppetlabs.
When the agent receives a SIGHUP the current run will finish and the agent will restart afterwards keeping its PID. If the above mentioned listen feature ist enabled this restart will create a socket.
When the agent receives a SIGTERM the current run stops immediately (maybe unclean, of that i'm not sure) and the agent halts. Afterwards the init system performs a normal startup of the puppet agent as if it were never started. If that fails for any reason there's no running puppet agent and the systems configuration may be incomplete.

I'll open an issue over at Puppetlabs to try make them ship "ExecReload=/bin/kill -HUP $MAINPID" in their systemd unit file. Meanwhile we could set the restart command to "/bin/kill -HUP $(cat /var/run/puppet/agent.pid)" on RHEL7 systems as a workaround.

mmoll:
Perfect, thank you! Could you please let us know the PuppetLabs issue number here?

martinpfeifer:
Here you go: https://tickets.puppetlabs.com/browse/PUP-5637

mmoll:
I see. Let's hope that fix is getting backported from 4.x...

martinpfeifer:
I've updated the pull request to restart the agent by sending it a SIGHUP signal.

manifests/params.pp
```diff
@@ -272,7 +272,8 @@
       $osreleasemajor = regsubst($::operatingsystemrelease, '^(\d+)\..*$', '\1') # workaround for the possibly missing operatingsystemmajrelease
       $agent_restart_command = $osreleasemajor ? {
         '6'     => "/sbin/service ${service_name} reload",
-        '7'     => "/usr/bin/systemctl reload-or-restart ${service_name}",
+        #'7'     => "/usr/bin/systemctl reload-or-restart ${service_name}",
+        '7'     => "/bin/kill -HUP $(pgrep -f '/usr/bin/ruby /usr/bin/puppet agent')",
```
ekohl: I'm not sure if it works and I don't have a CentOS 7 machine with puppet ready, but how about ```systemctl kill puppet --signal HUP```?

manifests/params.pp
```diff
@@ -272,7 +272,8 @@
       $osreleasemajor = regsubst($::operatingsystemrelease, '^(\d+)\..*$', '\1') # workaround for the possibly missing operatingsystemmajrelease
       $agent_restart_command = $osreleasemajor ? {
         '6'     => "/sbin/service ${service_name} reload",
-        '7'     => "/usr/bin/systemctl reload-or-restart ${service_name}",
+        #'7'     => "/usr/bin/systemctl reload-or-restart ${service_name}",
```
ekohl: Mind removing this line?

roidelapluie:
https://github.com/puppetlabs/puppet/pull/4576

mmoll:
As @roidelapluie's PR will be part of the next Puppet 3 release, I'd like to see this issue closed. Anybody any objections?

timogoebel:
@mmoll : @martinpfeifer is off for the week. I talked to him last week about the issue and I concur. Let's close this down. Thanks, guys!

mmoll:
ok, danke!

PR #323, opened 2015-10-28
Start working on systemd timer support

GregSutcliffe:
This is a start, just to see if upstream want such a patch - I'm using it to good effect on my local systems.

For background, systemd can replicate the functionality of cron, which has some pros and some cons. Further, Archlinux is no longer shipping a cron daemon by default, and moving over to systemd timers. This patch adds support for a scheduled run of the agent via systemd timers.

This is a PoC. If there's upstream interest, then I still need to:

* correctly use `$puppet@::cron_cmd
* correctly clean up files/services depending on which option is used
* Add tests
* [optional] Split each of the three options to it's own subclass

Happy to do so if this is the right direction. Thoughts?

### DIFF
--- a/templates/agent/systemd.puppetcron.timer.erb
+++ b/templates/agent/systemd.puppetcron.timer.erb
@@ -1,0 +1,14 @@
+<%# ip_to_cron can return strings or arrays depending on the request -%>
+<%# so we make them always be strings, in the right format -%>
+<% @times[0] = @times[0].join(",") if @times[0].kind_of? Array -%>
+<% @times[1] = @times[1].join(",") if @times[1].kind_of? Array -%>
+[Unit]
+Description=Systemd Timer for Puppet Agent
+
+[Timer]
+OnActiveSec=0m
+OnCalendar=<%= @times[0] %>:<%= @times[1] %>
+Persistent=true
+
+[Install]
+WantedBy=timers.target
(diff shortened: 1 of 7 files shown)


mmoll:
While not being a fan of systemd myself: if there are distros which make use of systemd's "base system functionality" (FreeBSD terminology) and stop installing packages that were just there by default it's great to have support in here, especially as this trend might also catch on in Debian or Fedora.

mmoll:
It might be a bit of a gun to big to fire here, but is there a puppet module for this kind of systemd configs that could be used?

GregSutcliffe:
@mmoll there's almost certianly some we can use to handle the writing files and reloading systemd. I'll look around.

GregSutcliffe:
@mmoll sadly I don't see anything that wraps up the plaintext files and the reload as a defined type / provider, but https://forge.puppetlabs.com/justin8/systemd#usage seems the best of the reload helpers. Not sure it's worth including just for that though...

templates/agent/systemd.puppetcron.timer.erb
```diff
@@ -0,0 +1,12 @@
+<% times[0] = times[0].join(",") if times[0].kind_of? Array %>
```
ekohl: Nitpick: ```-%>```

ekohl:
In general I think this could a nice addition. I do wonder about the daemon reload and if it won't clash with other exec names. Slightly leaning to just accepting the risk at first and dealing with it when it shows up.

PR #324, opened 2015-11-02
Add aio_package to be a parameter of the base class

jyaworski:
Hello:

I noticed that whenever I went to use this on my puppet 4 server, this happened:

````
[root@puppet-infra puppetserver]# puppet agent -t --environment puppetserver --server `hostname --fqdn` --noop
Info: Retrieving pluginfacts
Info: Retrieving plugin
Info: Loading facts
Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Evaluation Error: Left match operand must result in a String value. Got an Undef Value. at /etc/puppetlabs/code/environments/puppetserver/modules/puppet/manifests/params.pp:227:12 on node puppet-infra.domain.com
````
While this doesn't appear to be related directly, I noticed that it got to the elsif in the first place because aio_package was false. This allows the user to define it.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -145,6 +145,9 @@
 # $additional_settings::              A hash of additional main settings.
 #                                     type:hash
 #
+# $aio_package::                      Use the aio package. Defaults to false on puppet < 4
+#                                     type:boolean
+#
 # == puppet::agent parameters
 #
 # $agent::                            Should a puppet agent be installed
@@ -445,6 +448,7 @@ class puppet (
   $dir_group                       = $puppet::params::dir_group,
   $package_provider                = $puppet::params::package_provider,
   $package_source                  = $puppet::params::package_source,
+  $aio_package                     = $puppet::params::aio_package,
   $port                            = $puppet::params::port,
   $listen                          = $puppet::params::listen,
   $listen_to                       = $puppet::params::listen_to,
@@ -576,6 +580,7 @@ class puppet (
   validate_bool($server_puppetdb_swf)
   validate_bool($server_default_manifest)
   validate_bool($server_ssl_dir_manage)
+  validate_bool($aio_package)
 
   validate_hash($additional_settings)
   validate_hash($agent_additional_settings)
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -39,7 +39,9 @@ class puppet::params {
   $hiera_config        = '$confdir/hiera.yaml'
   $syslogfacility      = undef
   $environment         = $::environment
-  if versioncmp($::puppetversion, '4.0') < 0 {
+  if versioncmp($::puppetversion, '4.0') >= 0 {
+    $aio_package = true
+  } elsif versioncmp($::puppetversion, '4.0') < 0 {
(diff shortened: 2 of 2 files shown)


jyaworski:
Actually, it looks like a significant amount of logic is in the params.pp file to set variables. I'm not sure if the above commit does anything except set it to be true on puppet4....

Any ideas?

jyaworski:
Actually, I just sort of answered my own question, but it's dirty. A better solution would be to move this logic out of params.pp, but that's a lot more effort.

mmoll:
As explained in IRC, please pastebin the output of `facter -p` and the versions used. Having this in params.pp should normally lead to a situation where all the params are automatically set correctly so a paramter in init.pp should't be needed (IMHO that one would be even harmful). I can image that module not being facter 3.x ready, although I did already use it with AIO packages, so we would really need to reproduce this to know what's the root cause.

jyaworski:
@mmoll here's the paste:

http://pastebin.com/mnL0eGgT

ekohl:
@jyaworski I don't see anything like ```puppetversion``` so I'm wondering if ```facter -p``` doesn't load puppet specific values or it's that's the cause of the problem.

jyaworski:
@ekohl I noticed that, too. There's some talk on the docs about how it's a puppet variable, and not a fact. I've been trying to confirm that we can access it the same way.

jyaworski:
Looks like this is also being hit by #330

mmoll:
Could you please update to the most recent puppet-agent package including Puppet 4.3.1 with this commit: puppetlabs/puppet@7728edb and check if $::puppetversion is there again?

domcleal:
> Looks like this is also being hit by #330

master has now been updated with the fixes from #330, so if you'd like to rebase the PR then you'll find it works much more reliably.  You will see a failure on Puppet 2.7, but we're dropping this shortly [subject to feedback](https://groups.google.com/forum/#!topic/foreman-users/YQNfxUzG_As) so ignore it for now.

ekohl:
:-1: as it is now. I know at least Fedora is shipping puppet 4 as non-AIO so this would break. Adding the parameter to init.pp is also useless since it's only used in params.pp. If you do want to make this useful, then you either need to move a lot of logic to init.pp (bad idea) or introduce a globals.pp (like puppetlabs-postgresql does) where you can override parameters to params.pp.

I'd hope that this is just an issue with puppet 4.3.0 and this is fixed with 4.3.1.

mmoll:
It seems that facter 3.1.3 is the solution with this fix: puppetlabs/facter@c399786, verified with facter 3.1.2/3.1.3 on FreeBSD (non-AIO package).

mmoll:
@jyaworski I did cancel that test run to safe resources, could you please test with the newer Puppet/Facter and report back?

mmoll:
I guess this can get closed? :grinning:

jyaworski:
@mmoll yes. Closed. I'll see if I can revisit this at a later time.

mmoll:
I hoped it would be resolved with the new Facter/Puppet versions.

PR #325, opened 2015-11-03
Allow both server and use_srv_records to be defined for compatibility with <3.0

kmcfate:
This provides proper operation of older versions of puppet that do not support srv records without interfering with the operation of srv records.They are not mutually exclusive options. Understandably <=2.x is not supported but this avoids other various hacks required to maintain archaic infrastructure until it is upgraded. Upgrading to 3.x is not an option as the dependency chain falls to facter and virt-what. Virt-what is not available for me currently.


mmoll:
Is this change working with 3.x and 4.x agents, too? also, tests fail (as expected ;))

mmoll:
@kmcfate ping?

mmoll:
As there was no reaction and we dropped Puppet 2.7 tests today,  closing this.

PR #327, opened 2015-11-04
Fix #311 by moving reports parameter in puppet.conf

### LINKED ISSUE #311: "reports" parameter needs to go into [main] section of puppet.conf
I was having problems with an instance of puppetserver deployed by theforeman-puppet displaying this error on agent run, when integrated with puppetDB:
```
[puppet-server] Puppet Attempt to assign to a reserved variable name: 'trusted' on node xmaster.vagrant.vm
```
I found the module is placing the "reports" parameter in the [master] section of the puppetserver's puppet.conf, when it appears to belong in the [main] section instead.
https://tickets.puppetlabs.com/browse/PDB-949?focusedCommentId=198934&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-198934

Also, acknowledging Puppet's own docs about this are vague:
https://docs.puppetlabs.com/guides/reporting.html

jyaworski:
Tag @mmoll and @westbywest

### DIFF
--- a/spec/classes/puppet_config_spec.rb
+++ b/spec/classes/puppet_config_spec.rb
@@ -64,6 +64,10 @@ describe 'puppet::config' do
           should_not contain_file('/etc/puppet/manifests/default_manifest.pp')
         end
 
+        it 'should_not contain reports setting in puppet.conf' do
+          should_not contain_concat__fragment('puppet.conf+10-main').with_content(/\s+reports = .*/)
+        end
+
         it 'should contain puppet.conf [main]' do
           concat_fragment_content = [
             '[main]',
--- a/spec/classes/puppet_server_config_spec.rb
+++ b/spec/classes/puppet_server_config_spec.rb
@@ -10,7 +10,6 @@ describe 'puppet::server::config' do
           :clientcert             => 'puppetmaster.example.com',
           :concat_basedir         => '/nonexistant',
           :fqdn                   => 'puppetmaster.example.com',
-#          :rubyversion            => '1.9.3',
           :puppetversion          => Puppet.version,
       }) end
 
@@ -146,6 +145,7 @@ describe 'puppet::server::config' do
             with_content(/^\s+logdir\s+= #{logdir}$/).
             with_content(/^\s+rundir\s+= #{rundir}$/).
             with_content(/^\s+ssldir\s+= #{ssldir}$/).
+            with_content(/^\s+reports\s+= foreman$/).
             with_content(/^\s+privatekeydir\s+= \$ssldir\/private_keys { group = service }$/).
             with_content(/^\s+hostprivkey\s+= \$privatekeydir\/\$certname.pem { mode = 640 }$/).
             with_content(/^\s+autosign\s+= \$confdir\/autosign.conf { mode = 664 }$/).
@@ -156,7 +156,6 @@ describe 'puppet::server::config' do
             with({}) # So we can use a trailing dot on each with_content line
 
           should contain_concat__fragment('puppet.conf+30-master').
-            with_content(/^\s+reports\s+= foreman$/).
             with_content(/^\s+external_nodes\s+= #{nodepath}$/).
             with_content(/^\s+node_terminus\s+= exec$/).
             with_content(/^\s+ca\s+= true$/).
@@ -196,10 +195,6 @@ describe 'puppet::server::config' do
               server_reports        => 'store',
               server_external_nodes => '',
            }"
-        end
-
-        it 'should store reports' do
-          should contain_concat__fragment('puppet.conf+30-master').with_content(/^\s+reports\s+= store$/)
         end
 
         it 'should contain an empty external_nodes' do
--- a/templates/puppet.conf.erb
+++ b/templates/puppet.conf.erb
@@ -24,6 +24,9 @@
     autosign       = <%= scope.lookupvar('::puppet::autosign') %>
 
     show_diff     = <%= scope.lookupvar('::puppet::show_diff') %>
+<% if scope.lookupvar('::puppet::server') -%>
+    reports     = <%= scope.lookupvar('::puppet::server_reports') %>
+<% end -%>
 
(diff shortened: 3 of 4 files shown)


mmoll:
Is this change compatible with 2.7 and 3.x? also, tests fail (as expected ;))

mmoll:
@jyaworski ping?

jyaworski:
@mmoll pong. I've been away on other tasks for a bit. I'll take another look at this.

mmoll:
I tested this with a 2.7 and 3.8 client (with only the client installed, without any master) and with `puppet apply` this is leading to:
```
Warning: No report named 'foreman'
```
I can image this to happen even with puppet 4, so there has to get some logic in to set that parameter only when configuring a puppet server.

jyaworski:
@mmoll I find it odd that this PR triggered that. I'll look into ways around it.

mmoll:
I find it quite logical, actually. When in the main section, also the non-master install wants to send reports via foreman.rb, which is not installed on it, leading to that mesage. I'm a bit surprised PL did move the parameter, but I could imagine that this is on the other help helpful for masterless installs.

spec/classes/puppet_config_spec.rb
```diff
@@ -11,6 +11,7 @@
           :domain                 => 'example.org',
           :fqdn                   => 'host.example.com',
           :puppetversion          => Puppet.version,
+          :server                 => true,
```
mmoll: what fact is this? :)

spec/classes/puppet_config_spec.rb
```diff
@@ -282,6 +284,7 @@
       :osfamily => 'windows',
       :domain   => 'example.org',
       :puppetversion => Puppet.version,
+      :server => true,
```
mmoll: see above

spec/classes/puppet_server_config_spec.rb
```diff
@@ -199,7 +199,7 @@
         end
 
         it 'should store reports' do
-          should contain_concat__fragment('puppet.conf+30-master').with_content(/^\s+reports\s+= store$/)
+          should contain_concat__fragment('puppet.conf+10-main').with_content(/^\s+reports\s+= store$/)
```
mmoll: this test can be deleted then, IMHO, as it's already tested in line 149.

spec/classes/puppet_config_spec.rb
```diff
@@ -75,6 +76,7 @@
             '    hostprivkey = $privatekeydir/$certname.pem { mode = 640 }',
             '    autosign       = $confdir/autosign.conf { mode = 664 }',
             '    show_diff     = false',
+            '    reports     = foreman',
```
mmoll: this should not be there on agent only installs, correct?

spec/classes/puppet_config_spec.rb
```diff
@@ -312,6 +315,7 @@
           '    hostprivkey = $privatekeydir/$certname.pem { mode = 640 }',
           '    autosign       = $confdir/autosign.conf { mode = 664 }',
           '    show_diff     = false',
+          '    reports     = foreman',
```
mmoll: see above

jyaworski:
@mmoll derp. Fixed.

spec/classes/puppet_server_config_spec.rb
```diff
@@ -10,9 +10,11 @@
           :clientcert             => 'puppetmaster.example.com',
           :concat_basedir         => '/nonexistant',
           :fqdn                   => 'puppetmaster.example.com',
-#          :rubyversion            => '1.9.3',
           :puppetversion          => Puppet.version,
       }) end
+      let (:params) {{
+          :server                 => true,
```
mmoll: that should be superflous, see line 68

mmoll:
Tested with puppet 2.7, everything is still fine, so I think we're ready to go here. Please also add this test to make sure it's also working for the agent install:
```diff
diff --git a/spec/classes/puppet_config_spec.rb b/spec/classes/puppet_config_spec.rb
index a27b277..5e1bb47 100644
--- a/spec/classes/puppet_config_spec.rb
+++ b/spec/classes/puppet_config_spec.rb
@@ -64,6 +64,10 @@ describe 'puppet::config' do
           should_not contain_file('/etc/puppet/manifests/default_manifest.pp')
         end
 
+        it 'should_not contain reports setting in puppet.conf' do
+          should_not contain_concat__fragment('puppet.conf+10-main').with_content(/\s+reports = .*/)
+        end
+
         it 'should contain puppet.conf [main]' do
           concat_fragment_content = [
             '[main]',
```
Also squash the commits together into one and rebase onto current master.

jyaworski:
@mmoll done.

mmoll:
@jyaworski dzięki!
@domcleal @ekohl I think this is good to go. :+1:

jyaworski:
@mmoll Proszę :)

jyaworski:
So this works, but it looks like the tests timed out. We may want to consider addressing that project-wide.

domcleal:
@jyaworski yes, we're tracking and slowly fixing individual causes under #330.

spec/classes/puppet_server_config_spec.rb
```diff
@@ -10,7 +10,6 @@
           :clientcert             => 'puppetmaster.example.com',
           :concat_basedir         => '/nonexistant',
           :fqdn                   => 'puppetmaster.example.com',
-#          :rubyversion            => '1.9.3',
```
ekohl: I'd prefer to either remove this line or keep it in, but not comment it out.

mmoll: it's removed in that diff? :)

ekohl: And a good morning to me too ...

ekohl:
The code looks good :+1:

mmoll:
merged, jeszcze raz dzięki @jyaworski!

PR #328, opened 2015-11-09
Add systemd run mode

roidelapluie:


GregSutcliffe:
Looks remarkably similar to #323 - but you got the tests finished before me :)

I did think about refactoring it to use a class per run-type - see https://github.com/GregSutcliffe/puppet-puppet/commit/21ff4e191c2e5835216d2d62a89bfd8faf5fad26. It does tidy up the execution somehwat, I think...

roidelapluie:
I like it. Will merge both PR's

GregSutcliffe:
Awesome, thanks! I was going to rewrite the tests the same way (since then you only have to test the enable/disable cases for each class) but ran out of time.

Also, the systemd-enable case might need to drop the disable-cron section in my version. Since my Arch systems have no suitable cron provider, they error out trying to disable the cron resource. I see you don't have it anyway, I assume you hit the same thing? Can we make it conditional?

PR #333, opened 2015-11-24
Full interface for apache::mod::passenger

dmichelotto:
Added full interface for apache::mod::passenger, in order to permit passenger fine tuning

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -144,25 +144,34 @@ class puppet::params {
   $server_additional_settings = {}
 
   # Will this host be a puppetmaster?
-  $server                     = false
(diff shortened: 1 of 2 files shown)


ekohl:
We try not to manage ```apache::mod::passenger```. The preferred way is to configure it yourself and do that before including puppet-puppet.

mmoll:
Please see #294 and #310 for further information.

mmoll:

ekohl:
I agree with this. As I stated before, we try not to touch apache::mod::apache and leave that up to the user (either in a profile module or through data bindings). Puppet moving away from passenger to puppet server even makes this more logical.

Issue #334, opened 2015-11-25
Service[puppet-run.timer]: Provider systemd is not functional on this host


domcleal:
On EL6, which doesn't have systemd, the puppet::agent::service::systemd class is included with a service resource that tries to use the systemd service provider, which fails:

    # [ WARN 2015-11-25 07:35:06 verbose] /Stage[main]/Puppet::Agent::Service::Daemon/Service[puppet]/ensure: ensure changed 'stopped' to 'running'
    # [ERROR 2015-11-25 07:38:22 verbose] /Stage[main]/Puppet::Agent::Service::Systemd/Service[puppet-run.timer]: Provider systemd is not functional on this host

(http://ci.theforeman.org/job/systest_foreman/8036/tapResults/)

Perhaps the default value of $::puppet::unavailable_runmodes should take into account which OSes are systemd-enabled?

PR #335, opened 2015-11-25
Remove management of systemd timer unit on non-systemd OSes

domcleal:
Check specific versions of Debian and EL OSes to stop managing the
systemd timer, as a service resource with the systemd provider will
fail on older OS releases.

Fixes GH-334


ekohl:
:+1: on green tests.

mmoll:
merged, thanks @domcleal!

PR #338, opened 2015-12-04
Expose v1 /status endpoint

brandonweeks:
This is a departure from the default config in Puppet 3.x, however since Puppet 4 exposes the status endpoint by default I think this provides a more sane default.


mmoll:
@ekohl any opinion about this? I'm not an expert for these Puppet API things, but after reading the docs I think it's not causing any harm.

ekohl:
Sorry for the slow response, but :+1:

mmoll:
merged, thanks @brandonweeks!

PR #342, opened 2016-01-20
Fix default systemd and cron commands with AIO package

### LINKED ISSUE #340: systemd runmode does not work by default with Puppet 4 AIO
We use /urs/bin/env for cron and systemd run modes.

Which gives 

```
Dec 16 08:40:01 foo.bar[21958]: /usr/bin/env: puppet: No such file or directory
```

with the AIO package.

roidelapluie:
This commit simplifies the way we manage the puppet commands,
especially the puppet cert command.

This commit removes Puppet 2.7 support as well -- it was already not
supported by the tests anymore.

Closes #340.


roidelapluie:
@GregSutcliffe

roidelapluie:
using an absolute path is also meaningful from a security point of view. Using /usr/bin/env in a cron does not sound secure.

ekohl:
At first glance it looks alright and we already dropped puppet 2.7 support. Might as well make it official.

roidelapluie:
@ekohl that was not visible in metadata.json

metadata.json
```diff
@@ -31,7 +31,7 @@
   "requirements": [
     {
       "name": "puppet",
-      "version_requirement": ">= 2.7.0 < 5.0.0"
```
roidelapluie: @ekohl

ekohl:
@roidelapluie before this it could technically work, but we didn't test it. Now it's complete :)

roidelapluie:
@ekohl @GregSutcliffe Travis is green

ekohl:
:+1:

GregSutcliffe:
Works great here. It's been a long time since I committed anything to the installer... but if @ekohl gives it the thumbs up, it must be ok ;)

GregSutcliffe:
Merged as f4ea1e0, thanks @roidelapluie!

ekohl:
@GregSutcliffe we'll drag you back in ;)

PR #343, opened 2016-01-26
Add puppetserver configs

jyaworski:


ekohl:
@jyaworski Need any help finishing this?

manifests/server/puppetserver.pp
```diff
@@ -41,13 +41,22 @@
   $jvm_max_heap_size = $::puppet::server_jvm_max_heap_size,
   $jvm_extra_args    = $::puppet::server_jvm_extra_args,
 ) {
+  
+  File {
+    notify => Class['puppet::server::service'],
+  }
+
+  $puppetserver_config_dir = versioncmp($::puppetversion, '4.0.0') >= 0 ? {
```
ekohl: This line is causing the tests to fail, possibly because the fact is not defined in our tests and is causing the tests to fail.

jyaworski: Any ideas for a solution?

jyaworski:
@ekohl reviews would be good. I'm still progressing on this but it has been on the backburner to some other PRs I'm working on. I still very much need to do this, though.

brandonweeks:
@jyaworski this was on my todo list as well, so I'm very ready to test it. :)

manifests/server/puppetserver.pp
brandonweeks: `$puppetserver_config_dir = num2bool(versioncmp($::puppetversion, '4.0.0') >= 0) ? {`

That fixed the test for me.

ekohl: @brandonweeks that reads like it shouldn't work. I think it's not needed in params.pp because there it's an if statement and not a selector. In any case, I think we should move this to params.pp and possibly reuse ```$aio_package```.

ekohl:
I was poked by @zipkid at a meeting between puppet users, puppetlabs and foreman devs about puppet 4 that this is on his wishlist as well, so this is in demand.

jyaworski:
I'll see if I can work on this by the end of this week. I'll ping when it's comment-ready again.

ekohl:
@jyaworski awesome!

zipkid:
Hi @ekohl, @jyaworski,

Thanks for poking me on this.
As mentioned yesterday i also had to add management of conf.d/webserver.conf to set
- ssl-cert
- ssl-key
- ssl-ca-cert
which i don't find in the PR.

Also, in bootstrap.cfg, when '::puppet::server_ca' evaluates to false the following line should be added.
'puppetlabs.services.ca.certificate-authority-disabled-service/certificate-authority-disabled-service'

Regards,

Stefan.

jyaworski:
@zipkid thanks for the feedback. I'll make those changes. @ekohl and @brandonweeks I've already addressed some of your comments.

jyaworski:
@zipkid made a few changes. How's it look?

zipkid:
That is indeed what i meant. 
Thank you.

jyaworski:
Great. I believe this PR is mostly done, except that I have to make the gem-home configurable:

https://github.com/theforeman/puppet-puppet/pull/343/files#diff-133e2a0f1f52427763456aee8fba2a7aR5

I'm unaware of a way to get the puppetserver gem home without shelling out (is it worth a puppetserver_gem_home fact? I seriously hope not). If anyone has any ideas while I'm thinking on it, that would be great.

manifests/init.pp
```diff
@@ -4,431 +4,557 @@
...
-#                                     type:string
+# $server_enable_ruby_profiler::         Should the puppetserver ruby profiler be enabled?
+#                                        Defaults to false
+#                                        type:boolean
 #
-# $server_ssl_dir_manage::            Toggle if ssl_dir should be added to the [master]
-#                                     configuration section. This is necessary to
-#                                     disable in case CA is delegated to a separate instance
-#                                     type:boolean
+# $server_ca_authorization_required::    Should the puppetserver ruby profiler be enabled?
```
brandonweeks: This is missing from puppet::params and the description is wrong.

jyaworski: Good find. I did way too much copy-pasting during that part.

mmoll:
@jyaworski thanks for working on this, that's awesome :star2:
tests are failing at the moment...

templates/server/puppetserver/conf.d/webserver.conf.erb
```diff
@@ -0,0 +1,15 @@
+webserver: {
+    access-log-config = /etc/puppetlabs/puppetserver/request-logging.xml
```
brandonweeks: This shouldn't be hardcoded.

templates/server/puppetserver/conf.d/ca.conf.erb
```diff
@@ -0,0 +1,22 @@
...
+<%- if scope.lookupvar('puppet::server_ca_authorization_required') == false -%>
+        authorization-required: false
+<%- else -%>
+        authorization-required: true
+        client-whitelist: [
+<%- scope.lookupvar('puppet::server_ca_client_whitelist').each do |client| -%>
+            <%= client %>,
+    <%- end -%>
+<%- end -%>
+        ]
```
brandonweeks: This needs to be moved up a line or it break with `puppet::server_ca_authorization_required` set to False.

jyaworski: Good catch.

brandonweeks:
On Ubuntu 14.04 I'm getting `/usr/local/lib/site_ruby/1.9.1 for `$::rubysitedir`, which isn't an working entry for ruby-load-path.

templates/server/puppetserver/conf.d/webserver.conf.erb
brandonweeks: `access-log-config = <%= scope.lookupvar('puppet::server_puppetserver_dir') %>/request-logging.xml`

jyaworski:
Do you have an idea? I don't have the facility to test Ubuntu at the moment?

templates/server/puppetserver/conf.d/webserver.conf.erb
jyaworski: Noted.

templates/server/puppetserver/conf.d/puppetserver.conf.erb
```diff
@@ -0,0 +1,60 @@
...
+profiler: {
+    # enable or disable profiling for the Ruby code; defaults to 'false'.
+    enabled: <%= scope.lookupvar('::puppet::server_enable_ruby_profiler') %>
+}
+
+# Settings related to the puppet-admin HTTP API
+puppet-admin: {
+    client-whitelist: [
+    <%- scope.lookupvar('::puppet::server_admin_api_whitelist').each do |certname| -%>
+        <%= certname %>,
```
brandonweeks: >Exception in thread "main" com.typesafe.config.ConfigException$Parse: /etc/puppetserver/conf.d/puppetserver.conf: 58: in value for key 'client-whitelist': List should have had new element after a comma, instead had token: ':' (if you want the comma or ':' to be part of a string value, then double-quote it)

I get this error when `::1` is in the list with puppetserver 1.1.3.

brandonweeks:
@jyaworski I confirmed it with a clean install of Ubuntu, I don't think you'll be able to rely on `$::rubysitedir` for Ubuntu. 😢 @mmoll do you have any insight into why this fact points to an empty directory on Ubuntu?

templates/server/puppetserver/conf.d/web-routes.conf.erb
```diff
@@ -0,0 +1,10 @@
+web-router-service: {
```
brandonweeks: https://gist.github.com/brandonweeks/bc72082ab65c17b11396

Replacement template with puppetserver 1.x and 2.x support :)

mmoll:
@brandonweeks no idea, but since facter 2.x is probably unmaintained now, this won't get resolved.

templates/server/puppetserver/conf.d/puppetserver.conf.erb
jyaworski: Weird. Fixed.

templates/server/puppetserver/conf.d/web-routes.conf.erb
jyaworski: Dzięki!

jyaworski:
@mmoll facter 2.4.x is in maintenance mode, just like 3.8.x is for puppet. If there's no fact we can rely on, that's unfortunate. If we add a dependency on [puppetlabs-ruby](https://github.com/puppetlabs/puppetlabs-ruby), we can get access to the <code>$::gemhome</code> fact, which appears to be equivalent enough at least on a RHEL6 system:

```shell
[root@dagon ~]# facter -p gemhome rubysitedir
gemhome => /usr/lib/ruby/gems/1.8
rubysitedir => /usr/lib/ruby/site_ruby/1.8
```

We could always do directory magic or gsub...

jyaworski:
Actually, looking into it, os-settings should essentially never be changed. I'm suggesting we just let whatever the packager puts in there exist, and we don't manage the file. I've removed that part of the code. I can't find references to it in the docs for newer versions.

https://docs.puppetlabs.com/puppetserver/2.2/configuration.html
http://docs.puppetlabs.com/puppetserver/2.1/configuration.html
http://docs.puppetlabs.com/puppetserver/1.1/configuration.html

jyaworski:
Oh, sorry for the spam.... but it looks like I spoke too soon:

```shell
jruby-puppet: {
    # Where the puppet-agent dependency places puppet, facter, etc...
    # Puppet server expects to load Puppet from this location
    ruby-load-path: [/opt/puppetlabs/puppet/lib/ruby/vendor_ruby]

    # This setting determines where JRuby will look for gems.  It is also
    # used by the `puppetserver gem` command line tool.
    gem-home: /opt/puppetlabs/server/data/puppetserver/jruby-gems
```

In 2.x they moved it into puppetserver.conf. I think we can let that value stand (if we can get a puppetserver_version working), and leave os-settings.conf alone. I'm also willing to set gem-home on the same principle.

jyaworski:
OK, I think I have a working solution now (without rspec tests). I'll add the tests once I get confirmation this works on a wide variety of setups.

As an aside, I saw [puppetlabs-hocon](https://github.com/puppetlabs/puppetlabs-hocon) if we ever decide to move away from templates for this kind of thing.

templates/server/puppetserver/conf.d/puppetserver.conf.erb
brandonweeks: 9af8daec579c9db6fd37da289507c78dd21dda82 fixed it for me, puppetserver 1.x expects quotes in the conf file itself.

jyaworski: Committed.

manifests/params.pp
```diff
@@ -310,13 +316,23 @@
...
+  $server_ssl_dir_manage               = true
+  $server_default_manifest             = false
+  $server_default_manifest_path        = '/etc/puppet/manifests/default_manifest.pp'
+  $server_default_manifest_content     = '' # lint:ignore:empty_string_assignment
+  $server_max_active_instances         = $::processorcount
+  $server_idle_timeout_milliseconds    = 1200000
+  $server_connect_timeout_milliseconds = 120000
+  $server_enable_ruby_profiler         = false
+  $server_ca_authorization_required    = true
+  $server_admin_api_whitelist          = [ "127.0.0.1", "::1", $::ipaddress ] # lint:ignore:double_quoted_strings
```
daenney: Why not simply switch to single quotes and remove the lint exception?

jyaworski: :+1:

manifests/init.pp
```diff
@@ -4,431 +4,560 @@
...
+# $server_puppetserver_dir::             The path of the puppetserver config dir
+#                                        type:string
 #
-# $server_jvm_config::                Specify the puppetserver jvm configuration file.
+# $server_max_active_instances::         Max number of active jruby instances. Defaults to
+#                                        processor count
+#                                        type:integer
 #
-# $server_jvm_min_heap_size::         Specify the minimum jvm heap space.
+# $server_idle_timeout_milliseconds::    How long the server will wait for a response
```
mmoll: please cut off the `_milliseconds` and rather document it in the comment, that way the whole shitespace-shift will hopefully go back, making reviewing easier.

Also note that `type:sting` is default, so it's ommited in that case.

manifests/init.pp
```diff
@@ -4,431 +4,560 @@
...
+#                                        type:string
+#                                        type:integer
```
mmoll: That line slipped in I presume? kafo doesn't support multiple types anyway (iirc)

manifests/params.pp
```diff
@@ -310,13 +316,23 @@
   }
 
   $server_jvm_java_bin = '/usr/bin/java'
-  $server_jvm_min_heap_size = '2G'
-  $server_jvm_max_heap_size = '2G'
+  $server_jvm_min_heap_size = '4G'
+  $server_jvm_max_heap_size = '4G'
```
mmoll: is this needed for puppetserver 2? If not, please leave the 2G, I was already stumped when I switched to puppetserver 1 for module testing and my VM got maxed out.

templates/server/puppetserver/bootstrap.cfg.erb
```diff
@@ -0,0 +1,15 @@
+puppetlabs.services.request-handler.request-handler-service/request-handler-service
+puppetlabs.services.jruby.jruby-puppet-service/jruby-puppet-pooled-service
+puppetlabs.services.puppet-profiler.puppet-profiler-service/puppet-profiler-service
+puppetlabs.trapperkeeper.services.webserver.jetty9-service/jetty9-service
+puppetlabs.trapperkeeper.services.webrouting.webrouting-service/webrouting-service
+puppetlabs.services.config.puppet-server-config-service/puppet-server-config-service
+puppetlabs.services.master.master-service/master-service
+puppetlabs.services.puppet-admin.puppet-admin-service/puppet-admin-service
+puppetlabs.services.version.version-check-service/version-check-service
+
```
mmoll: I hope this is and will stay compatible with all the versions :children_crossing:, but as a puppet mdoule, ther's probably no way to do this future safe.

templates/server/puppetserver/conf.d/webserver.conf.erb
```diff
@@ -0,0 +1,15 @@
...
+    ssl-host          = 0.0.0.0
+    ssl-port          = <%= scope.lookupvar('puppet::server_port') %>
+    ssl-cert          = <%= scope.lookupvar('puppet::server_ssl_dir') %>/certs/<%= @fqdn.downcase %>.pem
+    ssl-key           = <%= scope.lookupvar('puppet::server_ssl_dir') %>/private_keys/<%= @fqdn.downcase %>.pem
+<%- if scope.lookupvar('puppet::server_ca') -%>
+    ssl-ca-cert       = <%= scope.lookupvar('puppet::server_ssl_dir') %>/ca/ca_crt.pem
+    ssl-cert-chain    = <%= scope.lookupvar('puppet::server_ssl_dir') %>/ca/ca_crt.pem
+    ssl-crl-path      = <%= scope.lookupvar('puppet::server_ssl_dir') %>/ca/ca_crl.pem
+<%- else -%>
+    ssl-ca-cert       = <%= scope.lookupvar('puppet::server_ssl_dir') %>/certs/ca.pem
```
mmoll: please have a look into the existing templates, there are already seperate variables for the SSL files

mmoll:
@jyaworski some initial comments, I'll test the next iteration with puppetserver 1. If possible, please always squash your changes into the existing commit and force-push or otherwise add `[skip ci]` to the commit message. At the moment the whole test suite will run for every commit, blocking the whole travis time for all theforeman repos using travis.

manifests/init.pp
jyaworski: That's a completely reasonable proposal on the milliseconds.

I added type:string to be pedantic, and to make it explicit. Is this opposed?

manifests/init.pp
jyaworski: Absolutely.

manifests/params.pp
jyaworski: Puppetserver 1.1.3 won't even run with less than 4G of RAM, I believe. I recall trying to stand up a very tiny box, only to be slapped down by saying it had a minimum required memory.

templates/server/puppetserver/conf.d/webserver.conf.erb
jyaworski: I don't understand. I am doing lookups on those variables. Did you mean something else?

jyaworski:
Thanks for letting me know about [skip ci]. I didn't know that.

manifests/init.pp
mmoll: Not really opposed, I think in the last (way smaller) PRs it was ommitted for briefness. but anyay, explicit is probably also here better than implicit. @ares?

manifests/params.pp
mmoll: I had 1.1.something running with 1G of heap. What I want to avoid is to shift away from the situation that at the moment some VM with 1-2GB RAM (or even a RPi or Odroid C1 with a little swap) can be used to try out the usual all-in-one foreman+proxy+puppetmaster setup. When puppetmaster is allowed to already allocate some gigs of RAM only for startup, I'll walk away at that point from testing this on e.g. my laptop. It's quite clear that the CLJ/JVM approach will always need more RAM, but it would be ncie to have this as low as possible to work OK for small setups and document prominently ion the README how to raise this for bigger deployments.

templates/server/puppetserver/conf.d/webserver.conf.erb
mmoll: never mind :bow:, in the old style puppet.conf, things like `$certname` are used then. Instead of `fqdn.downcase`, `<%= scope.lookupvar("puppet::server_certname") %>` could be used, although I also see that `fqdn` is used for some things already in teh current code, so I think this can be left as-is for the moment and we can think about how to handle this after this PR is merged.

templates/server/puppetserver/bootstrap.cfg.erb
jyaworski: We could try to manage the CA line only by using file_line.

jyaworski:
OK, I've reset the spacing and stopped clobbering travis until reviews are done.

templates/server/puppetserver/bootstrap.cfg.erb
mmoll: I think this OK as-is if it works with 1.x and 2.x, Otherwise I'd recommend rather augeas, as it's a tad more predictable/idempotent than `file_line` in such cases, IMHO.

jyaworski: I'm researching how to exactly get augeas to work with this file...

jyaworski: I think I got it using augeas.

jyaworski:
@raphink alerted me to their trapperkeeper lens.

Should we use this instead of my simplelines approach?

https://github.com/camptocamp/puppet-puppetserver/blob/master/files/lenses/trapperkeeper.aug

mmoll:
@jyaworski the trapperkeeper lens seems to be crafted for these exact config files, so that should even be easier in the end.

templates/server/puppetserver/conf.d/webserver.conf.erb
ekohl: Can't we reuse what's defined in ```puppet::server```? See https://github.com/theforeman/puppet-puppet/blob/26e1213d0df7e2efc16329616b620c94494da251/manifests/server.pp#L6-L14

manifests/init.pp
ares: I'm ok with either, strings are used most frequently and to support existing puppet modules we consider it default value. But if you prefer to be explicit that's perfectly fine.

jyaworski:
I'm unable to get the trapperkeeper lens to actually function on the bootstrap.cfg file:

```shell
$ cat test.augeas                                                        :(
#!/usr/local/bin/augtool -f -n -L -A
set /augeas/load/Trapperkeeper/lens trapperkeeper.lns
set /augeas/load/Trapperkeeper/incl /Users/joseph.yaworski/github/puppet-puppet/templates/server/puppetserver/bootstrap.cfg
load
print /augeas//error
ls /files/Users/joseph.yaworski/github/puppet-puppet/templates/server/puppetserver/bootstrap.cfg
print /files/Users/joseph.yaworski/github/puppet-puppet/templates/server/puppetserver/bootstrap.cfg
rm /files/Users/joseph.yaworski/github/puppet-puppet/templates/server/puppetserver/bootstrap.cfg/puppetlabs.services.ca.certificate-authority-disabled-service/certificate-authority-disabled-service
save

$ augtool -L -A -f test.augeas                                           :(
/augeas/files/Users/joseph.yaworski/github/puppet-puppet/templates/server/puppetserver/bootstrap.cfg/error = "parse_failed"
/augeas/files/Users/joseph.yaworski/github/puppet-puppet/templates/server/puppetserver/bootstrap.cfg/error/pos = "47"
/augeas/files/Users/joseph.yaworski/github/puppet-puppet/templates/server/puppetserver/bootstrap.cfg/error/line = "1"
/augeas/files/Users/joseph.yaworski/github/puppet-puppet/templates/server/puppetserver/bootstrap.cfg/error/char = "47"
/augeas/files/Users/joseph.yaworski/github/puppet-puppet/templates/server/puppetserver/bootstrap.cfg/error/lens = "/usr/local/Cellar/augeas/1.4.0/share/augeas/lenses/dist/trapperkeeper.aug:116.10-.56:"
/augeas/files/Users/joseph.yaworski/github/puppet-puppet/templates/server/puppetserver/bootstrap.cfg/error/message = "Syntax error"
```

Should we just stay with the simplelines config?

jyaworski:
Nevermind; I got it. I had to modify the lens and I submitted a PR against the other repo. We're good to go once I write some tests.

raphink:
Yes, the Trapperkeeper lens was made for this need specifically. Camptocamp's module also has wrapping definitions for bootstrap and puppetserver configs.

Camptocamp is not using this module anymore (we're not using Puppet at all to manage the Puppet master stack anymore), so we're willing to donate the code if you want to merge it.

mmoll:
@raphink yes, that would be perfect, do you know what the licensing implications of LGPL v2+ vs. GPL v3+ would be?

manifests/params.pp
```diff
@@ -311,14 +317,23 @@
...
   $server_default_manifest_content = '' # lint:ignore:empty_string_assignment
+  $server_max_active_instances     = $::processorcount
+  $server_idle_timeout             = 1200000
+  $server_connect_timeout          = 120000
+  $server_enable_ruby_profiler     = false
+  $server_ca_auth_required         = true
+  $server_admin_api_whitelist      = [ '127.0.0.1', '::1', $::ipaddress ]
+  $server_ca_client_whitelist      = [ '127.0.0.1', '::1', $::ipaddress ]
+  $server_cipher_suites            = [ 'TLS_RSA_WITH_AES_256_CBC_SHA256', 'TLS_RSA_WITH_AES_256_CBC_SHA', 'TLS_RSA_WITH_AES_128_CBC_SHA256', 'TLS_RSA_WITH_AES_128_CBC_SHA', ]
+  $server_ssl_protocols            = [ 'TLSv1', 'TLSv1.1', 'TLSv1.2', ]
```
ekohl: I'd argue to only use TLSv1.2 because I'd expect all puppetserver implementations run on sufficiently new software versions, but I have no idea how realistic that expectation is.

manifests/server/puppetserver.pp
```diff
@@ -35,19 +49,34 @@
...
+  $server_ssl_protocols        = $::puppet::server_ssl_protocols,
+  $server_ca                   = $::puppet::server_ca,
+  $server_dir                  = $::puppet::server_dir,
+  $vardir                      = $::puppet::vardir,
+  $server_ca_client_whitelist  = $::puppet::server_ca_client_whitelist,
+  $server_admin_api_whitelist  = $::puppet::server_admin_api_whitelist,
 ) {
 
+  File {
+    notify => Class['puppet::server::service'],
```
ekohl: Everything in this class notifies the service. ```puppet::server::config``` includes this class and since ```puppet::server::config ~> puppet::server::service``` I believe this is redundant.

spec/classes/puppet_server_puppetserver_spec.rb
```diff
@@ -7,7 +7,9 @@
     context "on #{os}" do
       let (:default_facts) do
         os_facts.merge({
-          :puppetversion          => Puppet.version,
+          :puppetversion => Puppet.version,
+          :ipaddress     => '1.2.3.4',
+          :processorcount => 1,
```
ekohl: Odd alignment here.

spec/classes/puppet_server_puppetserver_spec.rb
```diff
@@ -7,7 +7,9 @@
     context "on #{os}" do
       let (:default_facts) do
         os_facts.merge({
-          :puppetversion          => Puppet.version,
+          :puppetversion => Puppet.version,
+          :ipaddress     => '1.2.3.4',
```
ekohl: I think it's good a good practice to use [RFC 5735 example IPs](https://tools.ietf.org/html/rfc5737) or [RFC 1918 NAT IPs](https://tools.ietf.org/html/rfc1918).

templates/server/puppetserver/conf.d/webserver.conf.erb
ekohl: Please address this.

jyaworski: @ekohl Iet me be sure I understand what you're asking. You want

```erb
 ssl-cert          = <%= scope.lookupvar('puppet::server_ssl_dir') %>/certs/<%= @fqdn.downcase %>.pem 
```

to instead be

```erb
ssl-cert          = <%= scope.lookupvar('puppet::server::ssl_ca_cert') %>
```

If that's the case, I can change all of those to do that lookup instead.

manifests/params.pp
jyaworski: @ekohl I'm OK with it as a default. I'll change it.

templates/server/puppetserver/conf.d/webserver.conf.erb
ekohl: That's pretty much it, but in your example you have ssl-cert = $ssl_ca_cert where you meant $ssl_cert (which I didn't highlight but is a bit below).

jyaworski:
OK, I've made all the changes that @ekohl and others have requested.

templates/server/puppetserver/conf.d/ca.conf.erb
```diff
@@ -0,0 +1,22 @@
...
+        # this setting contains a list of client certnames who are whitelisted to
+        # have access to the certificate_status endpoint. Any requests made to
+        # this endpoint that do not present a valid client cert mentioned in
+        # this list will be denied access.
+<%- if @server_ca_auth_required == false -%>
+        authorization-required: false
+<%- else -%>
+        authorization-required: true
+        client-whitelist: [
+<%- @server_ca_client_whitelist.each do |client| -%>
```
ekohl: Nitpick: odd indenting

ekohl:
I think the code is looking good, but not that familiar with puppetserver. If others agree I think this can be merged.

templates/server/puppetserver/conf.d/ca.conf.erb
jyaworski: Fixed.

mmoll:
Tegarding the JAVA defaults, I just installed a recent puppetserver and it has `JAVA_ARGS="-Xms2g -Xmx2g -XX:MaxPermSize=256m"` as defaults, so the 2G defaults in this modules should be kept. 

Otherwise I think this looks good (but I will test net week) and it's mainly about getting green test now. :green_heart:

@brandonweeks could you test this against a puppetserver 1 installation you currently have unter puppet-puppet control?

jyaworski:
I've reset the defaults to 2G.

manifests/server/puppetserver.pp
```diff
@@ -35,11 +49,22 @@
...
+  $server_jruby_gem_home       = $::puppet::server_jruby_gem_home,
+  $server_ruby_load_paths      = $::puppet::server_ruby_load_paths,
+  $server_cipher_suites        = $::puppet::server_cipher_suites,
+  $server_max_active_instances = $::puppet::server_max_active_instances,
+  $server_ssl_protocols        = $::puppet::server_ssl_protocols,
+  $server_ca                   = $::puppet::server_ca,
+  $server_dir                  = $::puppet::server_dir,
+  $vardir                      = $::puppet::vardir,
+  $server_ca_client_whitelist  = $::puppet::server_ca_client_whitelist,
+  $server_admin_api_whitelist  = $::puppet::server_admin_api_whitelist,
```
mmoll: There are some values missing, that will be needed in the template later. At least:
$server_idle_timeout
$server_connect_timeout
$server_enable_ruby_profiler

templates/server/puppetserver/conf.d/ca.conf.erb
```diff
@@ -0,0 +1,22 @@
...
+        # have access to the certificate_status endpoint. Any requests made to
+        # this endpoint that do not present a valid client cert mentioned in
+        # this list will be denied access.
+<%- if @server_ca_auth_required == false -%>
+        authorization-required: false
+<%- else -%>
+        authorization-required: true
+        client-whitelist: [
+    <%- @server_ca_client_whitelist.each do |client| -%>
+            <%= client %>,
```
mmoll: Not 100% sure if it has any negative consequences on IPv4 addresses, but at least IPv6 addresses need to get double-quoted here, otherwise `::1` will result into: `Exception in thread "main" com.typesafe.config.ConfigException$Parse: /etc/puppetlabs/puppetserver/conf.d/ca.conf: 14: in value for key 'client-whitelist': List should have ended with ] or had a comma, instead had token: ':' (if you want ':' to be part of a string value, then double-quote it)`

mmoll:
in the spec tests, it's still 4G ;)

mmoll:
BTW, also puppetserver 1.1.3 has `-XX:MaxPermSize=256m` in JAVA_ARGS per default, so that should IMHO get added here, too.

brandonweeks:
I ran into the two things you already found, looks good otherwise so far.

manifests/init.pp
```diff
@@ -8,39 +8,52 @@
...
+#                                     type:string
 #
 # $sharedir::                         Override the system data directory.
+#                                     type:string
 #
 # $manage_packages::                  Should this module install packages or not.
 #                                     Can also install only server packages with value
 #                                     of 'server' or only agent packages with 'agent'.
 #                                     Defaults to true
+#                                     type:boolean
```
mmoll: this is a "stroolean", so it has to be string

manifests/init.pp
```diff
@@ -118,6 +137,7 @@
 # $ca_server::                        Use a different ca server. Should be either
 #                                     a string with the location of the ca_server
 #                                     or 'false'.
+#                                     type:boolean
```
mmoll: string also here

manifests/init.pp
```diff
@@ -363,46 +430,63 @@
...
+#                                     type:string
 #
 # $server_jvm_max_heap_size::         Specify the maximum jvm heap space.
+#                                     type:string
 #
 # $server_jvm_extra_args::            Additional java options to pass through.
 #                                     This can be used for Java versions prior to
 #                                     Java 8 to specify the max perm space to use:
 #                                     For example: '-XX:MaxPermSpace=128m'.
+#                                     type:array
```
mmoll: string also here, although this could be changed, I'd leave it as string for now, this can get refactored later, if needed.

templates/server/puppetserver/conf.d/ca.conf.erb
jyaworski: Fixed.

jyaworski:
Implemented all recommendations.

mmoll:
@jyaworski thanks, please feel free to squash in mmoll/puppet-puppet@d6c921f to get the tests green by setting the JAVA options back to 2G also in the tests.

Furthermore I couldn't get this to work on a fresh system with `puppet apply`. Further investigation and a look onto cryptocamp-puppetserver shows that the augeas lens has to be deployed before using it, I have a POC there: mmoll/puppet-puppet@a6f9044

However, I think now that it's probably better to have the addition of the trapperkeeper lens in a seperate PR, because this is getting a bit too big here. :bomb: 

So the way to go would be:

1. get camptocamp/puppet-puppetserver#25 merged
2. open a PR to add the trapperkeeper lens into theforeman/puppet-puppet - ideally @raphink and @domcleal could do this as augeas experts, otherwise (depending on how much is to do regarding testing in addition to just copying it in) I can draft the change and give colab permissions to my puppet-puppet fork to whoever wants to work on this as I'm a bit short on time)
3. rebase this PR onto theforeman/puppet-puppet master after the trapperkeeper lens went in
4. get this PR merged :smiley:

Opinions on this?

jyaworski:
@mmoll the augeas lens should work if pluginsync is enabled.

http://docs.puppetlabs.com/guides/plugins_in_modules.html#module-structure

Are you seeing something else?

mmoll:
@jyaworski did you test this with `puppet apply` on a fresh box also?

jyaworski:
My mistake; you're talking about running locally with an apply, and I was testing in a client/server setup so pluginsync was working for me.

If that's the case, we'll have to ship this the way that @raphink originally did it (and yell at Puppetlabs to do this better)?

mmoll:
Yes, the thing is for foreman-installer this needs to work locally.

Even if PL would change the behaviour, we would need to wait for a new Puppet version and depend on specific versions, that wouldn't end well. :)

For my POC I took the original cryptocamp code and transformed it back to a way that IMHO works without requiring cryptocamp-augeas.

jyaworski:
Proposed solution: since we're just managing one file with augeas and not the entirety of trapperkeeper, we can bypass https://github.com/camptocamp/puppet-puppetserver/pull/25 until we figure out a way to somehow get them to talk. For now, we only care about its behavior with bootstrap.cfg. I've incorporated your changes.

mmoll:
My changes were unpolished, at least mmoll/puppet-puppet@645c8e5 is needed, also the added lines in manifests/params.pp should be better aligned.

The lens test is not needed for now, as we don't test anything, which is not desirable, that's what I meant with "depending on how much is to do regarding testing in addition to just copying it in".

I think this PR is fine for now and we need to wait for other people to have a look onto the lens topic. :mag:.

jyaworski:
Sure. This is good to merge minus looking at the lens later.

brandonweeks:
https://gist.github.com/brandonweeks/6ca65417399650c4f8b8

Running into this.

jyaworski:
@brandonweeks what's your environment? I'm not running into that.

brandonweeks:
Oh, I'm an idiot. 😅 This is working perfectly for me with puppetserver 1.x. :+1:

domcleal:
> The lens test is not needed for now, as we don't test anything, which is not desirable, that's what I meant with "depending on how much is to do regarding testing in addition to just copying it in".

I'd prefer to keep the lens test in the tree because it's customised from the camptocamp version.  It's very important with Augeas lenses to typecheck them to prevent bizarre runtime behaviour, and this lets us do it (manually with augparse, if nothing else).  Perhaps just remove/comment the broken test for now so the file passes.

jyaworski:
@domcleal the failing test is the one that says that '/' can be a part of keys. I commented that out; the last few lines in the test.

mmoll:
Regarding the lens see https://github.com/camptocamp/puppet-puppetserver/pull/25#issuecomment-188790028, we'll just pull that in as-is now.

mmoll:
Merged, dziękujemy bardzo @jyaworski!

jyaworski:
@mmoll proszę bardzo!

Issue #67, opened 2013-05-23
Fix puppet common module path properly


GregSutcliffe:
#66 is a workaround, we need to properly parse the path and deal with missing directory elements.

ekohl:
Given the age of this issue and a proper solution is hard to come up with I think we should close this.

PR #348, opened 2016-02-15
Don't configure defaults environments if directory_environment is set

jlambert121:


mmoll:
Thanks for resurrecting this! As stated in the original PR, the default behaviour should be unchanged, I think it would be best to introduce a new class paramter there and add a test for it.

mmoll:
@jlambert121 ping?

jlambert121:
It seems like a bug to me now that it isn't possible for this to happen.  If dynamic environments are enabled and you're using directory environments, what is the use case for creating the common environment directories?

mmoll:
@domcleal's argument was that there's also server_environments which also sets up directories, maybe this can get refactored together somehow? I don't use directory_environments myself, so I can't argue too much...

domcleal:
I'm afraid I don't quite understand which part of the code block is causing the problem here.  Is it the creation of one or more environments?  If so then you'd set `$server_environments` to `[]` and it won't create any.  You could then manage the environment directories using r10k or similar.

The common directories are different, but I don't think we should link them to whether directory environments are in use.  You can use a common directory for modules shared across all environments whether you're using directory or config environments - it should perhaps be controlled by a separate parameter or `$server_common_modules_path`?

jlambert121:
I just set `server_dynamic_environments = true` which makes this no longer needed for me, but this module already has a crazy number of parameters which makes it very difficult to understand what to set.  Adding another parameter I don't think is a good solution.

mmoll:
would it make sense to document that under https://github.com/theforeman/puppet-puppet/blob/master/README.md#environments-support ?

domcleal:
At the moment the `server_dynamic_environments` parameter is deprecated when using directory envs, in favour of using `server_environments = []`.  Or is it something else you were trying to disable (see my earlier reply)?

jlambert121:
If you just set `server_environments = []` the file block that is inside this conditional will then create the common dir in the shared dir environment which then restarts puppet and then r10k will remove again later.  If `server_dynamic_environments` is deprecated, something is going to need to wrap this block to prevent that from happening, hence this PR which makes the code do what I expected.

mmoll:
related to #67 then?

discreet:
I personally think the common directory in environments has a use case and should remain. I'm a directory environments user (will move to dynamic environments eventually) and the common directory is perfect for things such as stdlib or module to be shared across all directories as previously stated.

brandonweeks:
Even with `server_environments = []` set the common directory is created, which _is_ less than ideal for a dynamic environment where custom tooling might expect to own the environments directory. Here is the log from my puppetserver after I delete `/etc/puppetlabs/code/environments/common`.

https://gist.github.com/brandonweeks/a4ac15eaf186da3c81f6

domcleal:
@brandonweeks right, that's the behaviour I meant when I said:

> The common directories are different, but I don't think we should link them to whether directory environments are in use. You can use a common directory for modules shared across all environments whether you're using directory or config environments - it should perhaps be controlled by a separate parameter or $server_common_modules_path?

As @discreet said, it does have a use with directory environments, which is why I'd prefer not to tie it to whether `$server_directory_environments` is in use or not.

When I added dir envs, I deprecated `$server_dynamic_environments` as I only intended it to control behaviour when using config environments, but the way it's written it's still useful with directory envs for this case.

I think we can either:

1. Use `$server_dynamic_environments` with directory envs too (this PR), and undeprecate this parameter.
1. Add a new parameter, e.g. `$server_manage_common_modules` to control whether to create common dirs or not.
1. Allow the existing `$server_common_modules_path` to be set to an empty/undef value to do the same.

mmoll:
I'm slightly leaning towards 3

discreet:
I like 3 because it makes use of existing features, though I'm not a fan of having leaving values empty/undef that can impact something that might be critical for users. Even though 2 is adding a new parameter it would be a simple boolean value with a very clear use.

jlambert121:
Personally I'm least in favor of a new parameter.  I think this module has a high barrier to use because of the massive number of parameters.

mmckinst:
I also like option 3.

mmoll:
@jlambert121 could you implement option 3 in this PR?

mmckinst:
option number 3 implemented in PR #400

domcleal:
Thanks for bringing this up @jlambert121, I'm going to close this PR now in favour of the change merged in #400 which allows the common directories param to be unset.

PR #350, opened 2016-02-16
environment.conf permissions are incorrect for custom users or groups

mleklund:
When you set puppet::server_environments_group, and or puppet::server_environments_user, envrionments.conf ends up with differing permissions. This can cause spurrious reloads of puppetmaster or puppetserver. This is probably not right if running with git, though git users may not be messing with the user settings.


ekohl:
:+1:

mmoll:
merged, thanks @mleklund!

PR #353, opened 2016-02-29
Fix reporting when running under puppet AIO installs

ekohl:
This follows the discussion in https://github.com/theforeman/puppet-foreman/pull/398.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -305,7 +305,10 @@ class puppet::params {
   $lower_fqdn              = downcase($::fqdn)
   $server_foreman          = true
   $server_facts            = true
-  $server_puppet_basedir   = undef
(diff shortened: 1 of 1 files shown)


mmoll:
looking good, merged. bedankt @ekohl!

ekohl:
I just realized I might have pushed the wrong version. This resolves to ```/opt/puppetlabs/puppet/lib/ruby/site_ruby/2.1.0/puppet``` instead of ```/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet```.

mmoll:
true, but it still sends the report :)

ekohl:
I think it didn't for me and I needed to place it in vendor_ruby.

mmoll:
fixed in #354, in the location of that PR, a report was sent but without contents

PR #355, opened 2016-03-05
Add yamldir and mode

mhjacks:
This is at the request of @mmoll.  Please critique and I will be happy to revise it.


mmoll:
see theforeman/puppet-foreman#242 for background

mmoll:
OK, after asking @mhjacks in IRC I think there was a miscommunication regarding the yamldir. the yamldir itself always was fine, so this PR is not needed, as this is handled by Puppet anyway: https://github.com/puppetlabs/puppet/blob/3.7.3/lib/puppet/defaults.rb#L1198-L1204
I guess this PR can get closed and we can proceed reviewing the original PR?

mhjacks:
I have no objections if the project wishes to proceed that way.

mmoll:
thanks, closing

PR #356, opened 2016-03-08
set hiera_config correctly on puppet 4

mmckinst:
This was changed in puppet 3 vs 4:

puppet 3: https://docs.puppetlabs.com/puppet/3/reference/configuration.html#hieraconfig
puppet 4: https://docs.puppetlabs.com/puppet/latest/reference/configuration.html#hieraconfig

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -152,7 +152,6 @@
 #                                     type:string
 #
 # $hiera_config::                     The hiera configuration file.
-#                                     Defaults to '$confdir/hiera.yaml'.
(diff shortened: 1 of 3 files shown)


ekohl:
The tests are failing, other than that the change looks good.

mmckinst:
the failing tests appears to be due to ruby 1.8.7 and bundler no longer being compatible with it

ekohl:
:+1:

mmoll:
merged as e5f0423, thanks @mmckinst!

Issue #339, opened 2015-12-16
PuppetDB workaround


roidelapluie:
When using [PuppetDB](https://github.com/puppetlabs/puppetlabs-puppetdb) and this module, the following dependency is needed:

```puppet
Exec['puppet_server_config-generate_ca_cert'] -> Package['puppetdb']
```

Not sure what to do with that.

mmoll:
It's a bit of a cludge, but how about this?
```diff
diff --git a/manifests/server/config.pp b/manifests/server/config.pp
index b826d2f..84cbb99 100644
--- a/manifests/server/config.pp
+++ b/manifests/server/config.pp
@@ -185,6 +185,7 @@ class puppet::server::config inherits puppet::config {
 
   ## PuppetDB
   if $puppet::server_puppetdb_host {
+    Exec['puppet_server_config-generate_ca_cert'] ->
     class { '::puppetdb::master::config':
       puppetdb_server             => $puppet::server_puppetdb_host,
       puppetdb_port               => $puppet::server_puppetdb_port,
```

roidelapluie:
it is not before the config but really before the package. Package postinstall script expects SSL certificates to be present.

roidelapluie:
currently I have added the hack on my internal profile_puppet module.

mmoll:
Hrm, I'm just seeing... As we only manage the server config and not the whole PuppetDB stuff anyway, there's not more we can do here, IMHO.

ekohl:
@roidelapluie looking at the code it should be ```Package[$::puppetdb::puppetdb_package]```.

@mmoll I think we can add the dependency in our module at the place you mentioned.

roidelapluie:
that will only affect puppet masters, not puppetserver setups

mmoll:
I'm a bit unsure what to do here. theforeman-puppet only has the integration to configure a master and thus is using `::puppetdb::master::config` which installs the puppetdb-terminus package. Can we somehow detect if the puppetdb module is also used to set up the PuppetDB server itself?

roidelapluie:
That's why I created this ticket. Maybe just a note in the README is enough.

ekohl:
You could work with ```if defined()``` but that's ugly so a known issue in the README is the best course of action.

roidelapluie:
There is another hack needed:

```
 Class['puppetdb::master::puppetdb_conf'] ~> Class['puppet::server::service']
```

mmoll:
[...]

PR #357, opened 2016-03-10
let puppetdb_conf notify the puppetmaster service

mmoll:
closes GH-339

See the issue for details, I think the documentation part is already covered by the existing entires regarding PuppetDB in the README.

### DIFF
--- a/manifests/server/config.pp
+++ b/manifests/server/config.pp
@@ -195,5 +195,6 @@ class puppet::server::config inherits puppet::config {
       manage_storeconfigs         => false,
       restart_puppet              => false,
     }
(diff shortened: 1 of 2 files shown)


domcleal:
Looks fine, but could you extend the with PuppetDB test in spec/classes/puppet_server_config_spec.rb to also check the catalog compiles with dependencies (http://rspec-puppet.com/matchers/)?  This will prevent any change in the PuppetDB module silently breaking here.

mmoll:
sure, done :bowtie:

domcleal:
:+1: if tests pass.

domcleal:
Merged, thanks @mmoll.

PR #358, opened 2016-03-11
Swap hiera_config default values for Puppet 3.x and 4.x

domcleal:
`$codedir` should only be used on Puppet 4, and the Puppet 3 default
should be using `$confdir`.  This failed on Puppet 3 tests with:

    /Puppet::Server::Config/Exec[puppet_server_config-generate_ca_cert]/returns:
    Error converting value for param 'hiera_config': Could not find value for $codedir

Follows up #356.


ekohl:
:+1:

mmoll:
merged, thanks @domcleal! integrative testing FTW!

PR #359, opened 2016-03-11
Use 'production' environment when proxing CA

aleksey-novikov:
Puppet CA may not have specified environment so replace it by 'production' one


aleksey-novikov:
I fix tests and renew request soon.

PR #361, opened 2016-03-13
document puppetlabs/puppetdb specifics

mmoll:
Tested with Debian/wheezy and CentOS with and without PC1 repos. This enables usage of the PuppetDB integration also with Puppet 4. As the terminus packages are not available for Fedora, EL is suffcient. I know it's a bit ugly, but the puppetdb module 5.x is designed to respond to that version variable.

Let me hear your comments and if it's OK like this, I'll add tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,11 @@ PuppetDB instance, using the puppetlabs/puppetdb module.  Use its
 `puppetdb::server` class to install the PuppetDB server and this module to
 configure the Puppet master to connect to PuppetDB.
 
-Requires [puppetlabs/puppetdb](https://forge.puppetlabs.com/puppetlabs/puppetdb) <5.0.0.
+Requires [puppetlabs/puppetdb](https://forge.puppetlabs.com/puppetlabs/puppetdb)
+
+Please see the notes about using puppetlabs/puppetdb 5.x with older versions of Puppet (< 4.x) and PuppetDB (< 3.x) with
+newer releases of the module and set the values via hiera or an extra include of `puppetdb::globals` with
+`puppetdb_version` defined.
 
 # Installation
 


ekohl:
Looking at the module, I don't think this is needed. It defaults to present and it allows users to override this in the profile layer.

mmoll:
I don't quite follow... If theforeman-puppet would be the single "entry-point" regarding Puppet master/server and it's configuration, how would this be handled without causing unexpected behaviour?

manifests/server/config.pp
```diff
@@ -188,6 +188,19 @@
...
+
+    unless $puppet::params::puppetdb_version == undef {
+      $pkg_suffix = $::osfamily ? {
+        'Debian'  => '-1puppetlabs1',
+        'FreeBSD' => '',
+        'Redhat'  => "-1.el${$puppet::params::osreleasemajor}",
+      }
+
+      class { '::puppetdb::globals':
+        version => "${puppet::params::puppetdb_version}${pkg_suffix}",
```
jyaworski: I would argue against referencing ::params things outside of params and the init class. We should always let the user override.

mmoll: pulled into init.pp, although it's not making the inital problem nicer ;)

ekohl:
Maybe we need a installer preferences profile module. It could expose additional settings:
* [PuppetDB settings](https://github.com/theforeman/puppet-puppet/pull/361)
* [Apache status board](https://github.com/theforeman/puppet-foreman/pull/408)
* [PostgreSQL settings](https://github.com/theforeman/puppet-foreman/pull/410)

They all don't belong in our modules, but are useful in the installer.

mmoll:
@ekohl I guess that would somehow touch the "scenarios" topic... however, I think this PR here is a bit different, as the only problem here is the nonoptimal design of pl-puppetdb 5.x regarding Puppet < 4

ekohl:
@mmoll the code implies the version defaults to installed, so why is this needed?

mmoll:
https://github.com/puppetlabs/puppetlabs-puppetdb#upgrading-from-4x-to-5x - If `puppetdb::globals::version` is not set, the `puppetdb-termini` package is installed on the puppet master, which only works for PuppetDB 3.x. This will render a Puppet 3.x setup unusable which requires the `puppetdb-terminus` package to talk to PuppetDB 2.x. (From all I can say, I'm not a PuppetDB expert).

mmoll:
oh, hang on, there's a `$legacy_terminus` parameter. Will look into that. :book:

jyaworski:
@ekohl big fan of removing external settings from our modules, and putting them specifically into the installer.

mmoll:
ALso that `legacy_terminus` parameter doesn't help... I'm out of ideas here, how to solve this differently.

Using 4.0 of the puppetdb module does get all the things right for Puppet 3.x, using 5.0 does get it right for Puppet 4.x.

mmoll:
@ekohl @domcleal Ping again, I don't think this can be solved otherwise, so IMHO we either have to go with this PR or remove PuppetDB integration via puppetlabs-puppetdb. At least within the installer it wouldn't work as expected, there would be ways around the problem when using it as a normal module by installing puppetlabs-pupeptdb 4.x or 5.x specifically.

jyaworski:
Ping again, again.

mmoll:
As I'm not smart enough to build a matching hierarchy in hiera for usage with foreman-installer, I'd like to just update puppetlabs-puppetdb there now and for the direct users of our module I want to just update the docs.

mmoll:
any comment?

domcleal:
Merged, thanks @mmoll!

PR #363, opened 2016-03-15
set puppet_etcdir of foreman module

mmoll:
this will put the correct etc_dir into the foreman module config

to be used together with theforeman/puppet-foreman#413

### DIFF
--- a/manifests/server/config.pp
+++ b/manifests/server/config.pp
@@ -177,6 +177,7 @@ class puppet::server::config inherits puppet::config {
       receive_facts  => $puppet::server_facts,
       puppet_home    => $puppet::vardir,
       puppet_basedir => $puppet::server_puppet_basedir,
+      puppet_etcdir  => $puppet::dir,
       enc_api        => $puppet::server_enc_api,
       report_api     => $puppet::server_report_api,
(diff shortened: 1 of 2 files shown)


brandonweeks:
Tested with Puppet 4. :+1:

ekohl:
Thanks! We should remember to bump the required version of puppet-foreman when releasing.

domcleal:
There's no dependency on puppet-foreman here, the only thing we might want to add is that puppet-foreman must be >= 5.0.0 in the README.

ekohl:
That's a good point. Actually for AIO installs it needs https://github.com/theforeman/puppet-foreman/commit/e14c941a2d543f7ba4144ecda5bb6aec4335bb6d which is not yet in a released version.

@mmoll what happens on FreeBSD? I think it now places that config in /usr/local/puppet/etc/foreman.yaml but the script doesn't look for it in that location.

mmoll:
@ekohl it does, as this is the wanted location for it: https://github.com/theforeman/puppet-foreman/blob/master/files/foreman-report_v2.rb#L27-L29 :)

ekohl:
Right :)

PR #364, opened 2016-03-16
RFC: split the classes.

SimonPe:
This pr is still kind of work in progress.

The target is to make the module more versatile by splitting it into multiple independent classes, keeping init.pp for compatibility and the simpler cases.

I had 2 targets:
 * make it possible to have agent and server config separate (agent config is in my base profile, while server config is in my puppetmaster profile)
 * make it possible to use an exported resource to set the enc. This makes it easy to have puppetmaster auto-detect a foreman instance and set it as enc.

Is this a kind of change that is wanted?
Comments on the general approach?

The code itself will probably still change a little


SimonPe:
btw: the meat of the change is https://github.com/SimonPe/puppet-puppet/blob/class_split/manifests/config/entry.pp
The rest are just refactors on top of that.

ekohl:
This is a pretty big change and I'm having trouble fully seeing the impact of this change so let's split the problem. There are 3 changes grouped together.

https://github.com/theforeman/puppet-puppet/pull/364/commits/d78f03111a7b3fb5caa805e427261e44c1611465 looks good and I think it can be accepted as it is now. That reduces the size of the PR a bit.

Another is the explicit parameters on agent.pp and server.pp. I like it and it simplifies many variables. It's now late in the commits and I'd like to see that before the other changes because I think we could accept that, regardless of the other changes.

That leaves the big changes: rewriting from templates to many simple concat fragments. It does break compatibility, but I don't know how many users rely on the removed parameters. I'm open to it, but it needs to sink in to see the impact. Also very interested in what @mmoll and @domcleal think.

SimonPe:
@ekohl:

Thanks for the quick feedback

> https://github.com/theforeman/puppet-puppet/pull/364/commits/d78f03111a7b3fb5caa805e427261e44c1611465 looks good and I think it can be accepted as it is now. That reduces the size of the PR a bit.

I'll split that out then.

> Another is the explicit parameters on agent.pp and server.pp. I like it and it simplifies many variables. It's now late in the commits and I'd like to see that before the other changes because I think we could accept that, regardless of the other changes.

I'm unsure whether splitting the explicit parameters to agent.pp and server.pp will make much sense/work (due to the other settings being in config.pp, but we could do it already on the parameters that are used in agent.pp and server.pp and add the extra once which move from config.pp when we get there)

Will look into this.

> That leaves the big changes: rewriting from templates to many simple concat fragments. It does break compatibility, but I don't know how many users rely on the removed parameters. I'm open to it, but it needs to sink in to see the impact. Also very interested in what @mmoll and @domcleal think.

Yeah, I know this is a big change, And it does indeed break compatibility in some parts:
 * custom templates: since we no longer use templates, it doesn't make any sense
 * `*_additional_settings` (We could make a compat layer for this.)
users of any of these should switch to using `puppet::config::*` resources.

ekohl:
> *_additional_settings (We could make a compat layer for this.) users of any of these should switch to using puppet::config::* resources.

I wonder if these can be automatically mapped to puppet::config::resources with create_resources to maintain compatibility.

SimonPe:
> I wonder if these can be automatically mapped to puppet::config::resources with create_resources to maintain compatibility.

Which is what I meant with _We could make a compat layer for this_

ps: noted that i got a little formatting error, which made this look wierd, fixed it now.

mmoll:
please rebase :)

SimonPe:
@mmoll I'm reworking it a bit and try to keep the test's passing for each commit.
I hope to update it soon. (but probably not today)

SimonPe:
@mmoll @ekohl : I reworked the patchset now.

It's broken up in #365 #375 and this one.

Could you take a look at it so maybe it can be included in the next major?

mmoll:
I guess 535708c is something that should go in anyway and is independent?

SimonPe:
@mmoll yes, it could go in independent, the issue it fixes only occurs further in the patch series, so it's not really a bug.

(the issue was that the puppetserver tests included only ::puppet and ::puppet::server::puppetserver, but not ::puppet::server, which caused issues if ::puppet::server::puppetserver accessed variables from ::puppet::server such as introduced in 7b556a3 )

SimonPe:
I Just pushed an updated version to fix the puppet-lint issues.

mmoll:
@SimonPe please open a PR with that change, to get that reviewed and in independently. Even if the underlying bug is not uncovered with the current module structure, it might be in the future anyway.

SimonPe:
@mmoll #377

SimonPe:
in #375 @domcleal wrote:
> My concern about #364 was that IIRC, this module used to be laid out that way, and we went back to a single class as there were some parameters that should be shared between the two classes (agent/master), which was difficult to deal with. With the Puppet agent and server being slowly decoupled in 4 with the separation of packages, it'll increasingly make sense to separate the classes, but I'm a bit hesitant until that's entirely complete (and we stop supporting 3).

The way I see it, by using a separate resource for each setting, I think we can mostly mitigate those issues.
When a setting is needed for both the agent and the server, we could start by not set that setting if it isn't passed to the class (by using a default "undef" value). And if the setting is passed, we could use ensure_resource() to protect against contradicting settings.

For non optional settings I would assume It to be fair to enforce them to be set on either a "global" level, or even specific for the agent. As I would assume that if you want to set up a puppetised puppet master, that you also want to be able to run puppet agent.

Further, I don't really know whether there's a need for this, When moving some parameters from the common part to the server part, I discovered that there are quiet a couple of parameters that while they have no effect on puppet agent, they do affect puppet apply runs. To solve this I would suggest adding a 3rd option to $::puppet::server_implementation, being "apply" this would set up most of this stuff (hiera_config and the like) without really setting up a puppet server/master.

My main motivation in this class separation story is the following:
Internally we have some base classes that are included on every node, one of those classes is responsible for setting up the puppet agent.
In a completely different set of classes, we have a class responsible for setting up the puppet master, Only this class, not the previous one, has access to the correct parameters needed for the master. So we need a way to set up the agent without the needed parameters for the server, and then add the server onto it. currently we use an ugly hack which disables puppet agent configuration on puppet masters, but I want to avoid that in the future, and have stuff just work. (also that means that now the config for the agent is in 2 places, once the server, once for all other nodes

I also think, that by making the different classes more independent, the whole code will be better testable using real unit tests instead of always needing to include the main class in order to test some subclass.

SimonPe:
this was merged as part of #376

PR #365, opened 2016-03-16
move default manifest creation to server config

SimonPe:
This only makes sense on a server

split out from #364

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -52,16 +52,4 @@ class puppet::config(
       content => template($nsauth_template),
     }
   }
-
-  # only manage this file if we provide content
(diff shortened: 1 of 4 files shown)


ekohl:
:+1: on green tests.

mmoll:
tests fail, I canceled the other testruns in the matrix to speed things up...

SimonPe:
@mmoll yeah, i saw it, I'll have to move those tests from puppet_config_spec.rb to puppet_server_config_spec.rb (And it's actually my first time touching puppet tests :-) )

SimonPe:
I updated the tests to only check the default_manifest when server => true. they are passing now locally.

mmoll:
merged, bedankt @SimonPe!

Issue #362, opened 2016-03-15
autosign file is not created


szemlyanoy:
Hi all,

It seems file $confdir/autosign.conf is not created at all which ends up with unsigned cert requests 

I propose simple file resource with template content which just iterates over array of ACLs

Cheers,
Sergey

domcleal:
Yes, this should probably be in this module.

We currently manage it in for Foreman's purposes in another module: https://github.com/theforeman/puppet-foreman_proxy/blob/master/manifests/puppetca.pp, so we'll need to remove from there.

szemlyanoy:
File resource mentioned above misses few things:
1. content filling from array
2. usage of $autosign value (default is "$confdir/autosign.conf { mode = 664 }")  as file target

mmoll:
@szemlyanoy could you open a PR with your proposed changes? we would be happy to have a look. :)

szemlyanoy:
On it!
Can I get access to collaborators please?

mmoll:
@szemlyanoy, please fork this repo and open a pull request from your fork to this repo, thanks!

szemlyanoy:
Hi all,

build is failed due to warning:
manifests/config.pp:34:WARNING: string containing only a variable

But why is that treated as not appropriate style
 file { "${autosign_target}":   
?

domcleal:
It can simply be written as `file { $autosign_target:`

PR #366, opened 2016-03-17
Autosign handler. Issue #362

szemlyanoy:
Added handler of autosing.conf path, also if true passed then defaults are used based on $dir folder, if false passed then no white rules apply(effectively disable all autosigns).
Array of whiterules can be passed now to module

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,24 @@
 # Changelog
 
+## 4.3.0
+* New or changed parameters:
+    * Add server_git_repo_mode, group and user parameters for repo ownership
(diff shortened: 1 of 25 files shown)


mmoll:
I think as this is server specific, this should to into the `server` hierarchy of configurations and also use `puppet::server_dir`.

Also, our travis is a bit overloaded of all the puppet-* pull requests at the moment, so it would be nice if you could run the tests (`bundle exec rake` or similar) locally.

Please squash your commits together and force-push if you make changes.

It should also be noted, once this gets merged here, we have to remove autodign from puppet-foreman_proxy.

PR #368, opened 2016-03-18
use puppetserver by default with AIO packages

mmoll:


spec/classes/puppet_server_config_spec.rb
```diff
@@ -79,7 +79,7 @@
             :creates => "#{ssldir}/certs/puppetmaster.example.com.pem",
             :command => "#{puppetcacmd} --generate puppetmaster.example.com",
             :require => ["Concat[#{conf_file}]", "Exec[puppet_server_config-create_ssl_dir]"],
-          }).that_notifies('Service[httpd]')
```
domcleal: Could we have another test that checks this when the master implementation is used?  I wonder if it ought to do the same for Puppet server too.

mmoll: kind of touches #229

mmoll: updated.

manifests/server/config.pp
```diff
@@ -102,6 +102,10 @@
     Exec['puppet_server_config-generate_ca_cert'] ~> Service[$puppet::server_httpd_service]
   }
 
+  if $::puppet::server_implementation == 'puppetserver' and $::puppet::server_ca {
+    Exec['puppet_server_config-generate_ca_cert'] ~> Service['puppetserver']
```
domcleal: I think this may be superfluous, since puppet::server already notifies from puppet::server::config to puppet::server::service.  The httpd one is probably required because puppet::service::service contains only the puppetmaster/puppetserver service and not Apache.

It might be that the tests don't pick up the notify, but it should work as-is.

spec/classes/puppet_server_config_spec.rb
domcleal: > I wonder if it ought to do the same for Puppet server too.

Sorry, I think this was already the case, now I look at the code in more detail!

manifests/server/config.pp
mmoll: right... added the test and removed the superfluous addition

domcleal:
:+1: if green.

ekohl:
A bit unoriginal, but :+1: if green.

mmoll:
:green_heart:

roidelapluie:
noo :) you will break my next PR

mmoll:
@roidelapluie maybe, but thenyou just have to rebase. We lied to ourselves the whole time about AIO vs. Puppet 4 :)

roidelapluie:
Why do you check AIO instead of checking if puppetversion > 4 ?

mmoll:
I still can set up a rack based puppet master with 4.x. (at least I could with 4.2 and am not aware of that having been ripped out). and there are OSes (namely FreeBSD) that gained a puppetserver package only very recently and that's not yet supported by this module. I can see that this might be changed in the future, but not now.

mmoll:
Also, in quite some tests, Puppet 4 = AIO was just presumed always, while we support Fedora or FreeBSD where there are non-AIO Agent packages.

manifests/params.pp
```diff
@@ -184,6 +183,12 @@
   $server_http_port           = 8139
   $server_http_allow          = []
 
+  # use puppetserver (JVM) or puppet master (Ruby)?
+  $server_implementation = $aio_package ? {
+    true    => 'puppetserver',
+    default => 'master',
+  }
+
```
jyaworski: The issue I have here is that nothing can override $aio_package, right? Is there a ::globals thing like postgres has?

mmoll: I'm strongly against that `globals` pattern, it's making things for non-hiera setups much worse in some aspects, especially when using something like foreman-installer, see #361. `aio_package` in theforeman-puppet is also not overrideable and I think duplicating the pattern here is no problem unless somebody (probably me) starts implementing support for puppetserver on platforms without AIO (e.g. FreeBSD). `server_implementation` is overrideable anyway.

ekohl: While I don't have strong objections against the globals pattern, I'd prefer to avoid it if possible. In this case I agree with @mmoll that you can still override ```$server_implementation``` and that's the only place it's used.

jyaworski: If this is just controlling implementation, rather than a whole host of stuff, that's fine. I can just override the implementation.

mmoll:
rebased

ekohl:
Thanks!

roidelapluie:
That requires a major version bump

mmoll:
I don't think so.

roidelapluie:
This commit will break any puppetmaster setup with AIO. And I would not bet that noone has one of them.

jyaworski:
If we did do a major version bump, I would feel less bad about potentially dropping Ruby 1.8.7 support in https://github.com/voxpupuli/puppet-extlib/pull/53

mmoll:
@roidelapluie That's somewhat true... maybe we should run a poll on the mailinglist or so, although I guess the probability to have such a setup on a Linux system is next to zero.

@jyaworski That's quite bad as, at least for now, we still need 1.8.7 compatibility for RHEL 6

roidelapluie:
@mmol polls are not part of semver. We break the API, we bump the major version. I do not see any alternative really.

mmoll:
@roidelapluie I know you are correct, but still am trying to mangle this as new functionality which doesn't break anything. ;)

ekohl:
I agree with @roidelapluie that this is a major version bump.

mmoll:
#371

PR #369, opened 2016-03-18
Allow to specify the listen ip address of the puppetserver

roidelapluie:
Such an attempt was done in the past (#200) but was very complex due to
the use of apache::vhost, which does mot allow mixing IP and non-IP
based vhosts.

This PR only takes care of the Puppetserver side and issues a warning if
users try to change the IP address with the puppetmaster implementation.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -5,6 +5,7 @@ class puppet::params {
   $version             = 'present'
   $user                = 'puppet'
   $group               = 'puppet'
+  $ip                  = '0.0.0.0'
   $port                = 8140
   $listen              = false
   $listen_to           = []
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -251,6 +251,9 @@
 # $server_dir::                       Puppet configuration directory
 #                                     type:string
 #
+# $server_ip::                        Bind ip address of the puppetmaster
+#                                     type:string
+#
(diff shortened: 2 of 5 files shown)


brandonweeks:
:+1:

manifests/params.pp
```diff
@@ -5,6 +5,7 @@
   $version             = 'present'
   $user                = 'puppet'
   $group               = 'puppet'
+  $ip                  = '0.0.0.0'
```
mmoll: I don't know if IPv6 is enabled by default with PL packages, but this would limit it to IPv4 only, correct?

brandonweeks: `ssl-host = 0.0.0.0` is all I have by default.

mmoll: then it's perfectly fine to keep this

roidelapluie: It was the hardcoded value in the template

mmoll: indeed :see_no_evil:

ekohl:
:+1:

mmoll:
tests fail...

roidelapluie:
Yeah sorry, will fix tests asap

roidelapluie:
Hopefully fixed the tests

mmoll:
merged, merci @roidelapluie!

PR #370, opened 2016-03-20
Autosign config handler. Issue #362

szemlyanoy:


mmoll:
I'm bit uncomfortable, where this is going, so I'm trying to provide more background. This is related to Foreman (or rather foreman-proxy), not to Puppet, although this module is used to manage Puppet.

The code currently handling the autosign file is https://github.com/theforeman/puppet-foreman_proxy/blob/master/manifests/puppetca.pp which is only ensuring that the autosign file exists and is usable for the foreman-proxy (that's why it's belonging to the foreman-proxy user, the puppet group and mode 0644, which might look a bit odd for a pure Puppet setup).  Having this set up in that specific way is curcial for foreman-proxy (and the whole Puppet integration) to work for us.

Also, foreman-proxy will manage the contents of this file by itself outside of Puppet, that's why no contents are ensured, which is something that's not making sense for a pure Puppet setup and probably the initial problem you want to fix with this PR, however, having Puppet managing the contents will render the autosign file unusable for foreman-proxy.

So I propose to split this up in some smaller step-by-step pull requests, which will also make it easier to add/change the tests, which always need updating/adding when doing a code change to ensure correctness of all parts in most possible configurations of this module and also ensuring that there are no unintended cross effects in other places when brining in one change.

The first step would be to port over the existing autosign handling from puppet-foreman_proxy and once this is ready, propose the needed changes for puppet-foreman_proxy to work with puppet-puppet by passing the needed parameters. Note that the whole autosign stuff is in the `[master]` section of puppet.conf, mapping to the `server` hierarchy in this module, take the `parser` or `strict_variables` config entries as examples how the related variables are passed in this this module, although I think `$confdir/autosign.conf { mode = 664 }` is best to be left in place in the first iteration anyway.

Lastly, please check the GitHub (and further) documentation about branching, squashing, resetting, rebasing and force-pushing with git. This also means, that if you change something and want to roll all the changes into one commit, an existing pull request (like this one here) can (and should) be used instead of opening a new one, which looses all the discussion of the old one.

Our workflow might be a bit confusing to Git(Hub) newcomers, but ensures a clean history, which is very important when it's needed to understand past changes https://blog.spreedly.com/2014/06/24/merge-pull-request-considered-harmful/

Sorry for making you jump through all the hoops for such a small change, but the functionality touched there is really crucial for us. :weary:

mmoll:
Hm, just found f36d451

szemlyanoy:
To be honest never worked with Foreman, so I need some time to examine how it operates and coordinates with puppetserver. From what I got it's used to have control over puppetmaster provisioning and maintaining.
It effectively means any change in puppet-puppet module would have need reflection in puppet-foreman_proxy module which sounds like potential impediment, isn't it? 

Regarding PR processing thanks for heads up, will take into consideration.

mmoll:
@szemlyanoy correct, foreman talks to foreman-proxy for doing things with various sub-services, one of them being the Puppet master. Foreman-proxy then does the real work of interacting with these services, that's why some configuration values (in this case the path of autosign.conf) have to be the same for the puppet master and foreman-proxy.

As you can see in the params.pp of theforeman-foreman_proxy some default values are being pulled from `$puppet::params`, I wouldn't call this `impediment`, but yes, sometimes it can be hard to change things, because you either need a single source of truth or configure it twice to the same value. Regarding the autosiogn.conf in special, that case is already a bit problematic as-is. However, I guess there are just very few people who want to set a different path of the file and that's why I'm proposing as the first step to just pull in the `file` definition we have in the foreman_proxy module (maybe after theforeman/puppet-foreman_proxy#225 got merged) and refactor foreman_proxy to tap the path from the theforeman-puppet module for the setting in the config file, foreman-proxy needs and that's done via theforeman-proxy.

After that the added functionality you need can be implemented parametrized in theforeman-puppet and we will think about some validation in the foreman_proxy module to fail if the contents of the autosign.conf file are under Puppet control, but puppetca functionality is used in foreman-proxy.

szemlyanoy:
I see https://github.com/theforeman/puppet-foreman_proxy/pull/225 is merged so we are likely good to proceed ahead. 
Should we merge current PR? It will not affect autosign.conf path. Then we can add in foreman_proxy logic to check if defaults from  puppet module were overridden

mmoll:
Please see mmoll/puppet-puppet@9988aaf for how I would approach that first step mentioned above.

I don't know if the other maintainers would agree whith that way of solving to have the possibilty to override the autosign file path and mode while still being able to use the boolean type parameters (Puppet does support true/false for that setting) for autosign and also note that I didn't fix any tests, this is just to show how I would approach it.

As mentioned above, this PR as-is does break functionality we desperately need.

szemlyanoy:
Having dedicated $autosign_mode variable is not a big deal(I can just change it in my PR easily if needed), also I thought combining true/false/path switches in one variable is not suboptimal, basically you are doing the same just inside ERB so not sure what's the difference. 
And apologies but could you please point out me breaking of what kind of functionality you mentioned above? 

And yes of course I need to adjust tests accordingly - it's another question

mmoll:
The thing with having a dedicated path variable is that this var can be tapped in foreman_proxy for the path in the config file, other than that there's no real value, although maybe the "stroolean" (string or boolean) behaviour seemed easier to implement for me that way.

If I see that correctly, with the current PR contents the autosign.conf _contents_ would be always rendered by Puppet and thus the file would be overwritten at each run. This breaks foreman-proxy, which menages that file on its own, outside of Puppet.

mmoll:
Another note about the file mode, we have 0664 as default because the foreman-proxy user is part of the puppet group to have write access to that file.

szemlyanoy:
> This breaks foreman-proxy, which manages the contents of that file on its own, outside of Puppet.

Ok what about checking if autosign.conf is controlled by puppet module and if yes then take it into consideration in foreman-proxy?  

Seems not that straightforward to maintain puppet-module since lots of clashes appear down the road with foreman-proxy

mmoll:
> what about checking if autosign.conf is controlled by puppet module

It's not about the file resource itself, but the content. While it's possible to check if a resource (only the resource, no parameters like content!) exists in a puppet catalog this is considered bad practice and we don't do it in our modules (there would also be the possibility of using exported resources, which might also be abused in such a case, but PuppetDB would be needed for that, which we don't require). So this would complicate matters even more as this would require a further boolean parameter, but even then it's not perfect.

The current situation also has it's own problems, as the foreman_roxy module assumes /etc/puppet/autosign.conf as path, while one can override it here and I'd like to at least improve that situation (this is not solvable, as `autosign = true` is also already supported in this modules here).

To reduce complexity I proposed to bring in only the file resource without managing content as a first step and then see how managing content can be brought in.

> Seems not that straightforward to maintain puppet-module since lots of clashes appear down the road with foreman-proxy

foreman-proxy is the main integration point for different services (Puppet being one of them) into Foreman, have a look onto the foreman_proxy module how different modules (like tftp, dns, dhcp) are used for downstream services. Personally I don't find it terribly complicated and can't say that there are frequent clashes. The problem here is the complexitity of what one can do with that autosign setting in puppet.conf itself (boolean + path + mode) and the implications of having the data in autosign.conf or not.

ekohl:
I largely agree with @mmoll in that we primarily aim for a good experience for foreman users but I would like to cater to non-foreman users as much as possible. That means I'd prefer to have parameters that default to the foreman situation but allow for other uses as well.

mmoll:
BTW, the autosign file change itself and a larger refactoring also touching this (and parameter passing to ::puppet::server) are in master now, so the could be rebased and reworked.

mmoll:
@szemlyanoy what's the staus here?

mmoll:
no reaction, closing

PR #372, opened 2016-03-22
Add new 'server_use_legacy_auth_conf' parameter.  Defaults to false.

alexjfisher:
For puppetserver 2, ```/etc/puppetlabs/puppetserver/conf.d/auth.conf``` is
only used if ```use-legacy-auth-conf``` is set to false in puppetserver.conf

The second commit adjusts ```/etc/puppetlabs/puppetserver/conf.d/auth.conf``` to allow access to the [Resource Type API](https://docs.puppetlabs.com/puppet/latest/reference/http_api/http_resource_type.html) This will be needed by the foreman puppet smart-proxy if/when https://github.com/theforeman/smart-proxy/pull/376 is merged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,11 @@ The Puppet master is configured under Apache and Passenger by default, unless
 3.x based installation, `server_implementation` can be set to `puppetserver`
 to switch to the JVM-based Puppet Server.
 
+When using Puppet Server 2 (version 2.0 was the first version to support Puppet 4),
+the module supports and assumes you will be installing the latest version (currently 2.3.1).
+If you know you'll be installing an earlier version, you will need to override
+`server_puppetserver_version`.
+
 Many puppet.conf options for agents, masters and other are parameterized, with
 class documentation provided at the top of the manifests. In addition, there
 are hash parameters for each configuration section that can be used to supply
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -523,6 +523,12 @@
 # $server_puppetserver_dir::          The path of the puppetserver config dir
 #                                     type:string
 #
+# $server_puppetserver_version::      The version of puppetserver 2 installed (or being installed)
+#                                     Unfortunately, different versions of puppetserver need configuring differently,
+#                                     and there's no easy way of determining which version is being installed.
+#                                     Defaults to '2.3.1' but can be overriden if you're installing an older version.
+#                                     type:string
+#
 # $server_max_active_instances::      Max number of active jruby instances. Defaults to
 #                                     processor count
 #                                     type:integer
@@ -562,6 +568,10 @@
 #
 # $server_ca_auth_required::          Whether client certificates are needed to access the puppet-admin api
 #                                     Defaults to true
+#                                     type:boolean
+#
+# $server_use_legacy_auth_conf::      Should the puppetserver use the legacy puppet auth.conf?
+#                                     Defaults to false (the puppetserver will use its own conf.d/auth.conf)
 #                                     type:boolean
 #
 # === Usage:
@@ -662,6 +672,7 @@ class puppet (
   $server_implementation           = $puppet::params::server_implementation,
   $server_passenger                = $puppet::params::server_passenger,
   $server_puppetserver_dir         = $puppet::params::server_puppetserver_dir,
+  $server_puppetserver_version     = $puppet::params::server_puppetserver_version,
   $server_service_fallback         = $puppet::params::server_service_fallback,
   $server_passenger_max_pool       = $puppet::params::server_passenger_max_pool,
   $server_httpd_service            = $puppet::params::server_httpd_service,
@@ -727,6 +738,7 @@ class puppet (
   $server_jvm_extra_args           = $puppet::params::server_jvm_extra_args,
   $server_jruby_gem_home           = $puppet::params::server_jruby_gem_home,
   $server_max_active_instances     = $puppet::params::server_max_active_instances,
+  $server_use_legacy_auth_conf     = $puppet::params::server_use_legacy_auth_conf,
 ) inherits puppet::params {
 
   validate_bool($listen)
@@ -819,6 +831,8 @@ class puppet (
     validate_array($server_admin_api_whitelist)
     validate_bool($server_enable_ruby_profiler)
     validate_bool($server_ca_auth_required)
+    validate_bool($server_use_legacy_auth_conf)
+    validate_re($server_puppetserver_version, '^[\d]\.[\d]\.[\d]$')
   } else {
     if $server_ip != $puppet::params::ip {
       notify {
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -349,4 +349,10 @@ class puppet::params {
   $server_ca_client_whitelist      = [ '127.0.0.1', '::1', $::ipaddress ]
   $server_cipher_suites            = [ 'TLS_RSA_WITH_AES_256_CBC_SHA256', 'TLS_RSA_WITH_AES_256_CBC_SHA', 'TLS_RSA_WITH_AES_128_CBC_SHA256', 'TLS_RSA_WITH_AES_128_CBC_SHA', ]
   $server_ssl_protocols            = [ 'TLSv1.2', ]
+
+  # Puppetserver >= 2.2  Which auth.conf shall we use? https://docs.puppetlabs.com/puppetserver/latest/config_file_auth.html#aside-changes-to-authorization-in-puppet-server-220
+  $server_use_legacy_auth_conf     = false
+
+  # For puppetserver 2, certain configuration parameters are version specific.  We assume a particular version here.
+  $server_puppetserver_version     = '2.3.1'
 }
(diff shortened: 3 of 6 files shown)


templates/server/puppetserver/conf.d/auth.conf.erb
```diff
@@ -56,6 +56,16 @@ authorization: {
             name: "puppetlabs environments"
         },
         {
+            match-request: {
+                path: "/puppet/v3/resource_type"
```
mmoll: :+1: I ran into this when testing that proxy PR

manifests/params.pp
```diff
@@ -353,4 +353,7 @@
   $server_cipher_suites            = [ 'TLS_RSA_WITH_AES_256_CBC_SHA256', 'TLS_RSA_WITH_AES_256_CBC_SHA', 'TLS_RSA_WITH_AES_128_CBC_SHA256', 'TLS_RSA_WITH_AES_128_CBC_SHA', ]
   $server_ssl_protocols            = [ 'TLSv1.2', ]
 
+  # Puppetserver >= 2.2  Which auth.conf shall we use? https://docs.puppetlabs.com/puppetserver/latest/config_file_auth.html#aside-changes-to-authorization-in-puppet-server-220
```
mmoll: that URL should go into the commit message

mmoll:
If that's only working with Puppetserver 2 from 2.2 onwards, that should be documented somewhere in the readme...

alexjfisher:
Are you happy for puppetserver 2 support (which is new anyway?) to be >= 2.2 only?

mmoll:
I am. @brandonweeks how was that whole puppetserver 2.x situation when using Puppet 3.8 agents?

brandonweeks:
Nooooooo. We absolutely need to support puppetserver 2.1 to provide a migration path for our existing users from 3.x to 4.x.

https://docs.puppetlabs.com/puppetserver/2.2/release_notes.html

mmoll:
is there a fact with the puppetserver version?

mmoll:
On the other hand, `$server_use_legacy_auth_conf` is overridable, so it could be `false` by default and docs added that for 2.1 that setting is needed...

brandonweeks:
No fact I could find. Might be worth adding one?

mmoll:
I hope to avoid that... I'm not in for working around Puppet Labs product/version politics, and my priority is for now to get things working with recent versions and AIO/puppetserver2 installs on a clean, new box. Once that is dully working, we can have a look at furtheri mprovements and migration strategies. Having that paramter to set doesn't sound too bad for me.

alexjfisher:
To be clear, a puppetserver 2.2 can still serve puppet 3 *clients* though, can't it?

If a parameter is in the config file, but the version doesn't support the parameter at all, I *think* it gets upset?  At least when I made up a puppetserver.conf option earlier, I think I recall ```service puppetserver restart``` moaning (I can double check tomorrow).  I suppose we could make the parameter boolean or undef?  The docs could say to set it to undef for puppetserver 2.1 and I could adjust the template accordingly.

Also, I could tighten the version comparison in the template.  Puppet server 2.2 [depends on puppet-agent 1.3.0](https://docs.puppetlabs.com/puppetserver/2.2/release_notes.html#new-requirements) which comes with puppet 4.3.  It would improve the default behaviour, but since you can still have puppetserver 2.1 installed with puppet 4.3 (or 4.4) it wouldn't avoid the need for another fix.

A fact could definitely be useful, but presumably would be blank until the puppetserver was installed and we'd have to make an assumption based on something else for the first run?

brandonweeks:
As far as I understand it puppetserver 2.2 drops the "legacy routes" feature that is required to interact with puppet 3 clients.

alexjfisher:
> As far as I understand it puppetserver 2.2 drops the "legacy routes" feature that is required to interact with puppet 3 clients.

I'm probably misreading something but...

Puppetserver 2.3 documentation has a section about [puppet 3 backwards compatibility](https://docs.puppetlabs.com/puppetserver/2.3/compatibility_with_puppet_agent.html#puppet-server:-backward-compatibility-with-puppet-3-agents).

brandonweeks:
Nope, _I_ completely misread some release notes... :sweat:

brandonweeks:
Tested working against puppetserver 2.2. :+1:

README.md
```diff
@@ -13,6 +13,9 @@ The Puppet master is configured under Apache and Passenger by default, unless
 3.x based installation, `server_implementation` can be set to `puppetserver`
 to switch to the JVM-based Puppet Server.
 
+When using Puppet Server 2 (version 2.0 was the first version to support Puppet 4),
+Puppet Server 2.2 is the minimum version supported by this module.
```
ekohl: I think this should be updated now that there's a parameter for this.

manifests/params.pp
mmoll: ...and be deleted here ;)

mmoll:
BTW, we could go for 2.3 as minimum puppetserver version, ruling out #373?

ekohl:
I don't mind requiring a newer version but I don't run a puppetserver 2.x in production so I don't know how much of a problem it would be.

README.md
alexjfisher: @ekohl Sorry, I'm not sure I follow.  Which parameter? 'server_use_legacy_auth_conf'?

alexjfisher:
@mmoll I wouldn't mind 2.3 being the minimum supported version, but I don't think that means https://github.com/theforeman/puppet-puppet/issues/373 doesn't need addressing unless we're also only supporting clean installs (where puppetserver 2.2 hasn't been manually configured previously).

README.md
ekohl: Am I understanding the situation correctly that for puppet server >= 2.3 you need to set ```server_use_legacy_auth_conf``` to ```true``` and for >= 2.0 < 2.3 to ```false```?

mmoll:
@alexjfisher yes, that would also be a consequence of that.

README.md
alexjfisher: Nope.  For puppetserver >= 2.2 we set it to ```false``` to use the 'conf.d/auth.conf' file (which was already being installed, but never used).

For puppetserver 2.0 and 2.1, if you set ```server_use_legacy_auth_conf``` to anything, I believe you get an error when starting the service.  Error will be of the form...

```Exception in thread "main" clojure.lang.ExceptionInfo: Output of initialize-config does not match schema```
At least that's what happens in 2.3 if you try to set an unrecognised option in the same section of puppetserver.conf

Comment (ekohl) [inline README.md]: Right, now I get it.

> Nope. For puppetserver >= 2.2 we set it to false to use the 'conf.d/auth.conf' file (which was already being installed, but never used).

Did you mean 2.3 here?

**Comment by mmoll** [comment]
`server_puppetserver_version` (without the 2) would be better. Like Puppet 4, some major version of puppetserver might come out sooner than expected.

**Comment by alexjfisher** [comment]
@mmoll Good point.  I hadn't considered puppetserver 3, just that the parameter wasn't relevant for puppetserver 1.x  I'll rename it.  Happy with keeping the regex the same?

**Comment by mmoll** [comment]
I'd rather go for `^[\d]\.[\d]\.[\d]$`, as a puppetserver 3 might be in some way incompatible with some clients but still work the same config files and commands, https://docs.puppetlabs.com/puppetdb/latest/release_notes.html#breaking-changes as example.

**Comment by mmoll** [comment]
:+1:

File: manifests/init.pp [hunk 4]
```diff
@@ -563,6 +569,9 @@
 #                                     Defaults to true
 #                                     type:boolean
 #
+# $server_use_legacy_auth_conf::      Should the puppetserver use the legacy puppet auth.conf?
+#                                     Defaults to false (the puppetserver will use its own conf.d/auth.conf)
```
domcleal: This parameter needs `type:boolean` under it.

manifests/init.pp
```diff
@@ -522,6 +522,12 @@
 # $server_puppetserver_dir::          The path of the puppetserver config dir
 #                                     type:string
 #
+# $server_puppetserver_version::      The version of puppetserver 2 installed (or being installed)
+#                                     Unfortunately, different versions of puppetserver need configuring differently,
+#                                     and there's no easy way of determining which version is being installed.
+#                                     Defaults to '2.3.1' but can be overriden if you're installing an older version.
+#                                     type:string
```
domcleal: string's superfluous, but no harm

domcleal:
@mmoll aside from a small comment about parameter types, please feel free to merge if you're happy with it.  I'm not very familiar with the subject.

templates/server/puppetserver/conf.d/auth.conf.erb
alexjfisher: I've just created https://github.com/theforeman/puppet-puppet/pull/388 which should probably be applied before this PR (which may then need a rebase).

mmoll:
please rebase

alexjfisher:
Rebased.

mmoll:
merged, thanks @alexjfisher!

PR #375, opened 2016-03-25
add explicit parameters to puppet::server

SimonPe:
part split out of #364 

note that server_main_template will disappear again with the rest of the reworked #364

### DIFF
--- a/templates/server/puppet.conf.main.erb
+++ b/templates/server/puppet.conf.main.erb
@@ -1,15 +1,15 @@
 ## Server config
 
-    reports          = <%= scope.lookupvar('::puppet::server_reports') %>
+    reports          = <%= scope.lookupvar('::puppet::server::reports') %>
 
-<% if scope.lookupvar("puppet::server_directory_environments") -%>
-    environmentpath  = <%= scope.lookupvar("puppet::server_envs_dir") %>
-    basemodulepath   = <%= [scope.lookupvar("puppet::server_common_modules_path")].flatten.join(':') %>
+<% if scope.lookupvar("puppet::server::directory_environments") -%>
+    environmentpath  = <%= scope.lookupvar("puppet::server::envs_dir") %>
+    basemodulepath   = <%= [scope.lookupvar("puppet::server::common_modules_path")].flatten.join(':') %>
 <% end -%>
-<% if scope.lookupvar("puppet::server_default_manifest") -%>
-    default_manifest = <%= scope.lookupvar("puppet::server_default_manifest_path") %>
+<% if scope.lookupvar("puppet::server::default_manifest") -%>
+    default_manifest = <%= scope.lookupvar("puppet::server::default_manifest_path") %>
 <% end -%>
 
-<% if scope.lookupvar("puppet::hiera_config") and !scope.lookupvar("puppet::hiera_config").empty? -%>
-    hiera_config     = <%= scope.lookupvar("puppet::hiera_config") %>
+<% if scope.lookupvar("puppet::server::hiera_config") and !scope.lookupvar("puppet::server::hiera_config").empty? -%>
+    hiera_config     = <%= scope.lookupvar("puppet::server::hiera_config") %>
 <% end -%>
--- a/templates/server/puppet.conf.main.erb
+++ b/templates/server/puppet.conf.main.erb
@@ -1,0 +1,15 @@
+## Server config
+
+    reports          = <%= scope.lookupvar('::puppet::server_reports') %>
+
+<% if scope.lookupvar("puppet::server_directory_environments") -%>
+    environmentpath  = <%= scope.lookupvar("puppet::server_envs_dir") %>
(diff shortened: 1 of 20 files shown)


templates/server/puppet.conf.main.erb
```diff
@@ -0,0 +1,13 @@
+## Server config
+
+    # Puppet 3.0.x requires this in both [main] and [master] - harmless on agents
+    autosign       = <%= scope.lookupvar('::puppet::autosign') %>
+    reports     = <%= scope.lookupvar('::puppet::server::reports') %>
```
ekohl: Alignment is off. Also please leave an empty line in between. It now looks like the comment is also for reports.

templates/server/puppet.conf.main.erb
```diff
@@ -0,0 +1,13 @@
+## Server config
+
+    # Puppet 3.0.x requires this in both [main] and [master] - harmless on agents
+    autosign       = <%= scope.lookupvar('::puppet::autosign') %>
```
ekohl: Since it refers to puppet 3.0.x (which we no longer support), I wonder if this is still needed. Anyone knows if this was fixed in later versions?

SimonPe:
@ekohl Addressed your first comment  (alingnment and whitespace) now.
For the second comment I have no clue about which versions really require it.

templates/server/puppet.conf.main.erb
domcleal: I'm pretty certain it was fixed in 3.1.x, but I've not double checked.

mmoll: @GregSutcliffe the original change came from one of your commits, do you remember any PuppetLabs ticket or similar? This would also influence #381.

mmoll:
I didn't test this in my setup, but IMHO this is looking good :+1:

SimonPe:
I rebased it all again for the latest changes on master.
Tests are still passing locally.

templates/server/puppet.conf.main.erb
GregSutcliffe: Sorry for the lag, missed that notification. I'm likewise fairly certain that this is no longer required, but I haven't tested it either.

mmoll: I'd say, just keep it for now (in #381).

mmoll:
After #381 is merged, I'd like to see this rebased and then quickly merged, so we don't require another rebase.

mmoll:
@SimonPe please rebase (hopefully the last time :bow:)

mmoll:
IMHO this also could get squashed into one commit.

SimonPe:
@mmoll This IS the rebase you are looking for!

I did the removal of autosign from [main] in a commit before the actual patch series becaus it's one less setting to bother with :-)

I still kept the rest in separate commits since that's easier for me to track the changes (tests complete after each commit)

mmoll:
@ekohl @domcleal what do you think?
@SimonPe it would be super nice if you could test that autosign still works, preferable with Puppet 3.1.x.

SimonPe:
@mmoll any idea of a platform that comes with 3.1? then i can quickly spin up a vagrant env.

mmoll:
@SimonPe you'd need to pull that from the puppetlabs repo

mmoll:
unfortunately this needs another rebase after #393

domcleal:
> @ekohl @domcleal what do you think?

Sorry for being a bit quiet on this, but I was a bit overwhelmed.  I think this is probably a reasonable change, it's more maintainable to see all of the validations code contained within puppet::server, though it does cause duplication of the parameters between the main entry class and the server class.

My concern about #364 was that IIRC, this module used to be laid out that way, and we went back to a single class as there were some parameters that should be shared between the two classes (agent/master), which was difficult to deal with.  With the Puppet agent and server being slowly decoupled in 4 with the separation of packages, it'll increasingly make sense to separate the classes, but I'm a bit hesitant until that's entirely complete (and we stop supporting 3).

SimonPe:
@mmoll: no problem, will look into it this weekend (same for the testing for autosign on puppet 3.1)

@domcleal:
Thanks for weighing in, better late than never :-)
The parameter duplication is indeed something that's not easy to avoid, unless we manage to completely split sever and agent and deprecate the use of the main class for configuring the server.
For the second part of your comment, I'll take that to #364 and try to give some answers there (later this evening probably)

SimonPe:
@mmoll  I rebased it but didn't manage to test it with puppet 3.1, I was unable to set up a puppet 3.1 env that actually worked even without that change.

mmoll:
@SimonPe does it work with 3.8?

SimonPe:
@mmoll

mmoll:
:+1:

mmoll:
If there are no objections, I would merge this tomorrow evening.

ekohl:
:+1:

mmoll:
merged as:
```
b2e3764 move parameter documentation to puppet::server
bb27f52 move server parameters to puppet::server
9f3631e split out settings for server
4bb44cf remove autosign from [main]
```
bedankt @SimonPe!

PR #376, opened 2016-03-25
Put the enc config in a separate class 

SimonPe:
This makes it possible to "autodiscover" the enc using exported
resources.

This sits on top of the reworked #364


mmoll:
please rebase :)

SimonPe:
@mmoll here you are.
sorry it took so long, I was a bit busy with other parts of our infra.

SimonPe:
@mmoll ping?

mmoll:
Sorry, I don't have much time at the moment, @ekohl @domcleal, what do you think? It's looking fine for me in general, however, I don't know if adding all the parameters to everything is a big downside.

SimonPe:
@mmoll  @domcleal @ekohl any update on this?

mmoll:
I'd like to bring this in generally, reviewed each change some weeks ago and think if tests are green after a rebase this is good to go.

@ekohl @domcleal any objections?

Also see https://github.com/theforeman/puppet-puppet/pull/364#issuecomment-213613886

domcleal:
@mmoll I've not reviewed it in detail, but I think I'm OK with the principle of the change, if you did want to merge it. It looks more reliable to manage the file entries from the manifest instead of keeping logic in the template.

manifests/config/additional_settings.pp
```diff
@@ -0,0 +1,12 @@
+define puppet::config::additional_settings(
+  $hash,
+  $resource,
+  $key = $title,
+  $options = {},
+){
+  create_resources($resource, {
+      "${key}" => {
```
ekohl: Is this needed or can it just be ```$key => {```?

manifests/config/entry.pp
```diff
@@ -0,0 +1,37 @@
...
+
+  # note the spaces at he end of the 'order' parameters,
+  # they make sure that '1_main ' is ordered before '1_main_*'
+  ensure_resource('concat::fragment', "puppet.conf_${section}", {
+    target  => "${::puppet::dir}/puppet.conf",
+    content => "\n\n[${section}]",
+    order   => "${sectionorder}_${section} ",
+  })
+
+  if (!defined(Concat::Fragment["puppet.conf_${section}_${key}"])){
```
ekohl: This could use a comment that it appends the value when the value is used multiple times.

SimonPe: added a comment now.

SimonPe:
@ekohl @mmoll : thanks for your patience, i rebased it now and added the comment @ekohl asked for, the "${key}" is still in there since the old parser requires this.

mmoll:
This is looking good and survived testing in my (small) environment. @ekohl what's your opinion?

carroarmato0:
This would be lovely

ekohl:
I haven't tested it, but the code looks good. Given @mmoll did some testing and we all agree it's a good idea I think we should merge this.

mmoll:
cherry-picked as

commit  | message
------------ | -------------
06dfd8c | Put the enc config in a separate class
bfbc974 | Add backwards compatible additional_settings handling
51a3570 | use resources to manage puppet.conf
0d25365 | add puppet::config::* resource definitions


hartelijk dank aan @SimonPe!

ekohl:
Big thank you to @SimonPe, this was a big refactor and thank you for your patience to stick with it.

PR #377, opened 2016-03-25
Make sure puppet::server is included in puppet::server::puppetserver

SimonPe:
the issue was that the puppetserver tests included only ::puppet and ::puppet::server::puppetserver, but not ::puppet::server, which caused issues if ::puppet::server::puppetserver accessed variables from ::puppet::server such as in #364

Thanks to @roidelapluie


ekohl:
:+1: looks like it can't hurt.

mmoll:
merged, bedankt / merci!

PR #378, opened 2016-03-28
Add support for puppet agent certname

noskovao:
In some cases you need to set custom client cert name in agent section of puppet.conf file.
For example if you are using pre-generated or generic certificate

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -229,6 +229,10 @@
 # $remove_lock::                      Remove the agent lock when running.
 #                                     type:boolean
 #
+# $client_certname::                  The node’s certificate name, and the unique
+#                                     identifier it uses when requesting catalogs.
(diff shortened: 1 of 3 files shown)


ekohl:
Would using the ```dns_alt_names``` also suit your use case? I'd prefer to avoid too many parameters, but if it doesn't then :+1:

noskovao:
Hi!
Unfortunatelly, `dns_alt_names` not solving the situation with generic certificate.
If generic certificate was signed on master, but agent providing his own cert with `FQDN.pem`, master returns:
`Exiting; no certificate found and waitforcert is disabled`

mmoll:
I'm not overly fond of this, but do see more than one use case.

mmoll:
merged, спасибо @noskovao!

PR #379, opened 2016-03-29
refactor tests as facterdb gained Windows facts

mmoll:
I disabled some tests because of https://github.com/fessyfoo/rspec-puppet-windows-issue

### DIFF
--- a/spec/classes/puppet_agent_config_spec.rb
+++ b/spec/classes/puppet_agent_config_spec.rb
@@ -5,6 +5,7 @@ describe 'puppet::agent::config' do
   on_supported_os.each do |os, facts|
     next if only_test_os() and not only_test_os.include?(os)
     next if exclude_test_os() and exclude_test_os.include?(os)
+    next if facts[:osfamily] == 'windows' # TODO, see https://github.com/fessyfoo/rspec-puppet-windows-issue
     context "on #{os}" do
       let(:facts) do
         facts.merge({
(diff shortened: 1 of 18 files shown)


spec/classes/puppet_agent_config_spec.rb
```diff
@@ -13,66 +13,68 @@
         })
       end
 
-      context 'with default parameters' do
-        let :pre_condition do
-          'include ::puppet'
-        end
+      unless facts[:osfamily] == 'windows'
```
ekohl: Can also be written as ```context 'with default parameters', :if => (facts[:osfamily] != 'windows') do```. I think it's cleaner because you end up with fewer indents. Also results in a smaller diff :)

mmoll: I did something similar now, much better diff!

spec/classes/puppet_agent_config_spec.rb
```diff
@@ -5,6 +5,7 @@
   on_supported_os.each do |os, facts|
     next if only_test_os() and not only_test_os.include?(os)
     next if exclude_test_os() and exclude_test_os.include?(os)
+    next if facts[:osfamily] == 'windows'
```
ekohl: Isn't this class run on windows as well? In that case I'd at least add a TODO to add that coverage.

mmoll: done.

PR #380, opened 2016-03-30
Added RedHat 5 to $osreleasemajor conditionals

szynaka:


ekohl:
I don't have an EL5 install at hand, but it looks like it makes sense.

mmoll:
EL5 is unsupported, according by metadata.json and we don't test it in any way. While I'm not against merging this, the status of 5 is questionable.

szynaka:
Michael,
  I hadn't checked the metadata.json before submitting.  Feel free to
ignore it if you wish, as I know how bad it can be to try to support an OS
you don't have locally.  I originally fixed it on my end through Hiera data
but thought it was an easy enough fix I'd submit a pull.

Ed

mmoll:
@szynaka as said, I'm not against merging this... I suppose you're running this module with EL5 nodes happily?

ekohl:
Thanks!

PR #371, opened 2016-03-22
bump major version

mmoll:

PR #381, opened 2016-03-30
manage autosign file in puppet-puppet

mmoll:
Also to be used a base for #370. This breaks the API and has to go into #371 after merging. puppet-foreman_proxy also needs updating after this.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -90,13 +90,14 @@
 #                                     the puppet agent.
 #                                     type:integer
 #
-# $autosign::                         Enable or disable autosign or change location
-#                                     of autosign.conf or autosign script.  If this
-#                                     is set to a script, make sure that script
-#                                     considers the content of autosign.conf, as it
-#                                     might break Foreman functionality if it
-#                                     doesn't. If this is set to a boolean, it can
-#                                     be cast as a string or a boolean.
+# $autosign::                         If set to a boolean, autosign is enabled or disabled
+#                                     for all incoming requests. Otherwise this has to be
+#                                     set to the full file path of an autosign.conf file or
+#                                     an autosign script. If this is set to a script, make
+#                                     sure that script considers the content of autosign.conf
+#                                     as otherwise Foreman functionality might be broken.
+#
+# $autosign_mode::                    mode of the autosign file/script
 #
 # $usecacheonfailure::                Switch to enable use of cached catalog on
 #                                     failure of run.
@@ -621,6 +622,7 @@ class puppet (
   $splay                           = $puppet::params::splay,
   $splaylimit                      = $puppet::params::splaylimit,
   $autosign                        = $puppet::params::autosign,
+  $autosign_mode                   = $puppet::params::autosign_mode,
   $runinterval                     = $puppet::params::runinterval,
   $usecacheonfailure               = $puppet::params::usecacheonfailure,
   $runmode                         = $puppet::params::runmode,
@@ -800,7 +802,8 @@ class puppet (
   validate_string($service_name)
 
   if ! is_bool($autosign) {
-    validate_string($autosign)
+    validate_absolute_path($autosign)
+    validate_string($autosign_mode)
   }
 
   validate_array($listen_to)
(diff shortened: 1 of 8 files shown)


spec/classes/puppet_config_spec.rb
```diff
@@ -130,23 +130,23 @@
...
-      end
+      # describe 'when autosign_location => /somedir/custom_autosign, autosign_mode => 664' do
+      #   let :pre_condition do
+      #     "class { '::puppet': autosign_location => '/somedir/custom_autosign', autosign_mode => '664', }"
+      #   end
+      #
+      #   it 'should contain puppet.conf [main] with autosign = /somedir/custom_autosign {mode = 664}' do
+      #     verify_concat_fragment_contents(catalogue, 'puppet.conf+10-main', [
+      #       '[main]',
+      #       '    autosign = /somedir/custom_autosign {mode = 664}',
```
mmoll: I don't know why this breakes

mmoll: was actually a whitespace problem

templates/server/puppet.conf.erb
```diff
@@ -2,7 +2,12 @@
 ## Module:           '<%= scope.to_hash['module_name'] %>'
 
 [master]
-    autosign       = <%= scope.lookupvar('::puppet::autosign') %>
+<% autosign = scope.lookupvar('::puppet::autosign') -%>
+<% if !!autosign == autosign -%>
+  autosign         = <%= autosign %>
```
domcleal: Nitpick, indentation should be four spaces

domcleal:
Looks fine, could you also update the CHANGELOG please?  Do you have a corresponding branch for foreman_proxy?

mmoll:
updated and opened theforeman/puppet-foreman_proxy#230

manifests/init.pp
```diff
@@ -90,13 +90,14 @@
...
-#                                     is set to a script, make sure that script
-#                                     considers the content of autosign.conf, as it
-#                                     might break Foreman functionality if it
-#                                     doesn't. If this is set to a boolean, it can
-#                                     be cast as a string or a boolean.
+# $autosign::                         If set to a boolean, autosign is enabled or disabled
+#                                     for all incoming requests. Otherwise this has to be
+#                                     set to the full file path of an autosign.conf file or
+#                                     an autosign script. If this is set to a script, make
+#                                     sure that script considers the content of autosign.conf
```
domcleal: If this is a script, should the file still be managed?  Now it's going to have a file { } resource, while before we only set autosign in puppet.conf.

mmoll: I would say yes. In that case the mode would need to be 755 or similar, but other than that is there any impediment?

domcleal: I don't think so, just a distinction perhaps to be wary of if we do manage the content of the file, as the parameter's a bit overloaded.

mmoll: As the content is not managed for now (will be part of #370), this should be OK?

domcleal: Yes, I think so - but we might have to re-examine the difference in #370.

spec/classes/puppet_config_spec.rb
```diff
@@ -130,20 +130,20 @@
...
           verify_concat_fragment_contents(catalogue, 'puppet.conf+10-main', [
             '[main]',
-            '    autosign       = true',
+            '    autosign = true',
           ])
         end
       end
 
-      describe 'when autosign => $confdir/custom_autosign {mode = 664}' do
+      describe 'when autosign => /somedir/custom_autosign, autosign_mode => 664' do
```
domcleal: Could you add a test for the presence of the autosign file resource?

mmoll: added to/puppet_server_config_spec.rb - a better test will be probably possible after #375, as there will be an explicit paramter also to the server class.

domcleal:
:+1: when tests pass.

domcleal:
Merged, thanks @mmoll!

PR #382, opened 2016-04-01
add passenger_min_instances and passenger_pre_start parameters

timogoebel:
This PR adds the parameters server_passenger_min_instances and server_passenger_pre_start to tune passenger a bit.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -420,6 +420,16 @@
 #                                     to lower this. Defaults to 12.
 #                                     type:integer
 #
+# $server_passenger_min_instances::   The PassengerMinInstances parameter. Sets the
+#                                     minimum number of application processes to run.
+#                                     Defaults to the number of processors on your
+#                                     system.
+#                                     type:integer
+#
+# $server_passenger_pre_start::       Pre-start the first passenger worker instance
+#                                     process during httpd start.
+#                                     type:boolean
+#
 # $server_config_version::            How to determine the configuration version. When
 #                                     using git_repo, by default a git describe
 #                                     approach will be installed.
@@ -662,6 +672,8 @@ class puppet (
   $server_puppetserver_dir         = $puppet::params::server_puppetserver_dir,
   $server_service_fallback         = $puppet::params::server_service_fallback,
   $server_passenger_max_pool       = $puppet::params::server_passenger_max_pool,
+  $server_passenger_min_instances  = $puppet::params::server_passenger_min_instances,
+  $server_passenger_pre_start      = $puppet::params::server_passenger_pre_start,
   $server_httpd_service            = $puppet::params::server_httpd_service,
   $server_external_nodes           = $puppet::params::server_external_nodes,
   $server_template                 = $puppet::params::server_template,
@@ -748,6 +760,8 @@ class puppet (
   validate_bool($server_puppetdb_swf)
   validate_bool($server_default_manifest)
   validate_bool($server_ssl_dir_manage)
+  validate_bool($server_passenger_pre_start)
+  validate_integer($server_passenger_min_instances)
 
   validate_hash($additional_settings)
   validate_hash($agent_additional_settings)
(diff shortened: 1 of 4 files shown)


mmoll:
please see https://github.com/theforeman/puppet-puppet/pull/333#issuecomment-187738163

domcleal:
@mmoll these settings are contained within the vhost, they're not server-wide.  I think it's the same as the ones in puppet-foreman.

timogoebel:
@mmoll : Yes, I agree with not exposing "::apache::mod::passenger". We need to remove "passenger_max_pool" from puppet::server::passenger as it has no effect right now.
As @domcleal stated, my settings are vhost specific settings. There is no way I know of to overwrite them so I disagree with you closing this PR.

mmoll:
oh, ok, I was just scanning for "passenger_*"... per chost it doesn't clash :bow:

timogoebel:
@mmoll : No worries, thanks for reopening this. What do you think of the code?

manifests/init.pp
```diff
@@ -420,6 +420,16 @@
 #                                     to lower this. Defaults to 12.
 #                                     type:integer
 #
+# $server_passenger_min_instances::   The PassengerMinInstances parameter. Sets the
+#                                     minimum number of application processes to run.
+#                                     Defaults to the number if processors on your
```
mmoll: **of** processors

manifests/server/passenger.pp
```diff
@@ -42,11 +44,21 @@
...
     'passenger_enabled' => 'On',
   }
-  
+
   $directories = [
     $directory,
   ]
 
+  $http_pre_start = $passenger_pre_start ? {
+    true  => "http://${::fqdn}:${http_port}",
```
mmoll: I guess you tested this and it's not needed, I only did read the passenger docs on that topic, but is there a / needed at the end?

mmoll:
one nitpick, one small question, :+1: otherwise

manifests/server/passenger.pp
timogoebel: Without the / should be finde, see https://github.com/theforeman/puppet-foreman/blob/76c7f4bb52383458ba6abc49a02c049e067986f3/manifests/config/passenger.pp#L160

timogoebel:
@mmoll: Thanks for taking a look. I fixed the nitpick and answered your question inline.

spec/classes/puppet_server_passenger_spec.rb
```diff
@@ -66,6 +71,30 @@
...
+        end
+
+        it 'should include the puppet http vhost' do
+          should contain_apache__vhost('puppet-http').with({
+            :passenger_min_instances => 10,
+            :passenger_pre_start     => 'http://puppet.example.com:8139',
+          })
+        end
+      end
+
```
mmoll: superfluous newline here

spec/classes/puppet_server_passenger_spec.rb
```diff
@@ -6,10 +6,15 @@
     next if exclude_test_os() and exclude_test_os.include?(os)
     next if os_facts[:osfamily] == 'windows'
     context "on #{os}" do
+      let(:pre_condition) do
+        'include ::puppet'
+      end
```
mmoll: what's happening without this?

timogoebel: init.pp and params.pp are not evaluated, therefore e.g. passenger_pre_start is null and the tenary operator that sets http_pre_start is not evaluated correctly (because it only accepts true and false).

timogoebel:
@mmoll : What's the status of this?

spec/classes/puppet_server_passenger_spec.rb
mmoll: The parameters of such a sub-class needs to get set in each test then, to have that isolated.

domcleal: Yep, just edit the `default_params` let block below and add :passenger_pre_start, :port and :http_port.  This saves us from loading the entire module in, which should make a big difference to the speed of the tests.

mmoll:
One test somehow was looping at travis, but the corresponding other Ruby 1.8.7 test was green.
Merged, Danke @timogoebel!

PR #384, opened 2016-04-01
Add Archlinux params block for OS defaults

GregSutcliffe:
Mostly the same as the AIO block. I'm unsure on $sharedir and $server_lenses_dir as I only run Archlinux clients, but these are my best guesses. 

EDIT: forgot tests. looking at that now

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,10 +8,10 @@ env:
   - PUPPET_VERSION=4.0 ONLY_OS=centos-7-x86_64,ubuntu-16-x86_64,ubuntu-16.04-x86_64
   - PUPPET_VERSION=3.5 ONLY_OS=centos-7-x86_64,ubuntu-16-x86_64,ubuntu-16.04-x86_64 FUTURE_PARSER=yes TRUSTED_NODE_DATA=yes
   - PUPPET_VERSION=3.5 ONLY_OS=centos-7-x86_64,ubuntu-16-x86_64,ubuntu-16.04-x86_64
-  # Test the rest of the supported platforms
+  # Test the rest of the supported platforms (Arch only has 4.0+ available in it's repos)
   - PUPPET_VERSION=4.0 EXCLUDE_OS=centos-7-x86_64,ubuntu-16-x86_64,ubuntu-16.04-x86_64,scientific-6-x86_64
-  - PUPPET_VERSION=3.5 EXCLUDE_OS=centos-7-x86_64,ubuntu-16-x86_64,ubuntu-16.04-x86_64,scientific-6-x86_64 FUTURE_PARSER=yes TRUSTED_NODE_DATA=yes
-  - PUPPET_VERSION=3.5 EXCLUDE_OS=centos-7-x86_64,ubuntu-16-x86_64,ubuntu-16.04-x86_64,scientific-6-x86_64
+  - PUPPET_VERSION=3.5 EXCLUDE_OS=centos-7-x86_64,ubuntu-16-x86_64,ubuntu-16.04-x86_64,scientific-6-x86_64,archlinux-3-x86_64 FUTURE_PARSER=yes TRUSTED_NODE_DATA=yes
+  - PUPPET_VERSION=3.5 EXCLUDE_OS=centos-7-x86_64,ubuntu-16-x86_64,ubuntu-16.04-x86_64,scientific-6-x86_64,archlinux-3-x86_64
 matrix:
   fast_finish: true
   include:
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -89,6 +89,22 @@ class puppet::params {
       $server_jruby_gem_home      = undef
     }
 
+    'Archlinux' : {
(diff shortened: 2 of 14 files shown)


mmoll:
Do these settings work for 3.x and 4.x agents? Or is only 4.x available in the Arch package repos nowadays?

GregSutcliffe:
@mmoll only 4.x is available in Arch

manifests/params.pp
```diff
@@ -88,6 +88,18 @@
       $server_lenses_dir = '/usr/local/share/augeas/lenses'
     }
 
+    /^Archlinux$/ : {
```
domcleal: `'Archlinux'`, no regexp required

manifests/params.pp
```diff
@@ -88,6 +88,18 @@
...
+        $dir               = '/etc/puppetlabs/puppet'
+        $codedir           = '/etc/puppetlabs/code'
+        $logdir            = '/var/log/puppetlabs/puppet'
+        $rundir            = '/var/run/puppetlabs'
+        $ssldir            = '/etc/puppetlabs/puppet/ssl'
+        $vardir            = '/opt/puppetlabs/puppet/cache'
+        $sharedir          = '/opt/puppetlabs/puppet'
+        $bindir            = '/usr/bin'
+        $server_lenses_dir = '/usr/share/augeas/lenses'
```
domcleal: This parameter's no longer needed.

GregSutcliffe: Oh good, I did wonder what it was for.

GregSutcliffe:
@domcleal thanks, updated. I do need to return to the tests when I have a moment though, last I looked it was causing an issue...

mmoll:
@GregSutcliffe what did you push there? ;)

GregSutcliffe:
@mmoll I think I rebased off the last tag rather than master :)

.travis.yml
```diff
@@ -8,10 +8,10 @@ env:
   - PUPPET_VERSION=4.0 ONLY_OS=centos-7-x86_64,ubuntu-16-x86_64,ubuntu-16.04-x86_64
   - PUPPET_VERSION=3.5 ONLY_OS=centos-7-x86_64,ubuntu-16-x86_64,ubuntu-16.04-x86_64 FUTURE_PARSER=yes TRUSTED_NODE_DATA=yes
   - PUPPET_VERSION=3.5 ONLY_OS=centos-7-x86_64,ubuntu-16-x86_64,ubuntu-16.04-x86_64
-  # Test the rest of the supported platforms
+  # Test the rest of the supported platforms (Arch only has 4.0+ available in it's repos)
```
domcleal: This file is controlled through modulesync, it must be added to foreman-installer-modulesync.

mmoll: I told Greg to already include it here (although I meant it to be squashed into the main commit) to get green tests, we are aware of it. :)

GregSutcliffe: Squashed, and PR opened - https://github.com/theforeman/foreman-installer-modulesync/pull/43

spec/classes/puppet_agent_config_spec.rb
```diff
@@ -36,25 +36,27 @@
         end
       end
 
-      context 'with runmode => cron' do
-        let :pre_condition do
-          'class { "::puppet": runmode => "cron" }'
-        end
+      unless facts[:osfamily] == 'Archlinux'
+        context 'with runmode => cron' do
```
domcleal: add `unless: facts[:osfamily] == 'Archlinux'` here instead of wrapping the whole block

manifests/params.pp
```diff
@@ -333,6 +349,10 @@
       $agent_restart_command = undef
       $unavailable_runmodes = ['cron', 'systemd.timer']
     }
+    'Archlinux': {
+      $agent_restart_command = "/usr/bin/systemctl reload-or-restart ${service_name}"
+      $unavailable_runmodes = ['cron']
```
domcleal: Is this a deliberate choice? Why's it disabled?

GregSutcliffe: There is no cron available in Arch in a clean install - https://wiki.archlinux.org/index.php/Cron#Installation

As such, even an absent resource will fail because there's no provider to execute it.

domcleal: Ok, that's alright then - the user can always override `unavailable_runmodes` if necessary.

metadata.json
```diff
@@ -97,6 +97,9 @@
       ]
     },
     {
+      "operatingsystem": "Archlinux"
```
ekohl: If no version is named, will rspec actually run the tests for Arch?

GregSutcliffe: It seems to, yes. I ran a full spec (no ONLY_OS exported) before starting on the tests and got a bunch of failures.

GregSutcliffe: @ekohl https://travis-ci.org/theforeman/puppet-puppet/jobs/171371748#L283

mmoll:
merged, thanks @GregSutcliffe!

PR #389, opened 2016-04-13
remove useless parameter server_passenger_max_pool

timogoebel:
@mmoll: Just some cleanup :-)

```passenger_max_pool``` is a global passenger setting which we do not want to support. The parameter is not passed through to puppetlabs-apache module and therefore has no effect.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 * New or changed parameters:
     * Add server_ip for configuring the listen IP (puppetserver only)
     * Add passenger_min_instances and passenger_pre_start for passenger tuning
+    * Remove passenger_max_pool which had no effect
 * Other features:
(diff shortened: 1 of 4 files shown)


ekohl:
It should be possible to pass this to a vhost, though puppetlabs-apache doesn't support it. Since I introduced this feature and broke it when refactoring to puppetlabs-apache, I'm fine with removing it.

timogoebel:
@ekohl : According to https://www.phusionpassenger.com/library/config/apache/reference/#passengermaxpoolsize it's only valid in "server config" context.

domcleal:
:+1: then.

ekohl:
@timogoebel then maybe it used to work before because it was placed outside of the vhost. :+1: on green tests.

mmoll:
merged, danke @timogoebel!

Issue #391, opened 2016-04-17
Release module against new extlib


jyaworski:
puppet-extlib has a new forge release with ip_to_cron imported. Please release a new version of this module.

PR #392, opened 2016-04-19
delete ip_to_cron and pull it in via extlib

mmoll:
closes GH-391


mmoll:
iirc @domcleal already tested, that it's still working with both modules bringing this file, so we don't need an immediate release of the module. Otherwise I'm of course also fine with having a 4.3.2.

domcleal:
Thanks @mmoll!

PR #393, opened 2016-04-20
Respect server_certname in puppetserver configuration

roidelapluie:


ekohl:
We have this assumption in a few modules I think :(

roidelapluie:
This changes works great here.

ekohl:
I think puppet-foreman and puppet-foreman_proxy have the same assumption for their certificates. Have you checked those too?

spec/classes/puppet_server_puppetserver_spec.rb
```diff
@@ -191,6 +191,24 @@
...
+          default_params.merge({
+            :server_puppetserver_dir => '/etc/custom/puppetserver',
+          })
+        end
+
+        let :pre_condition do
+          "class {'puppet': server_certname => 'puppetserver43.example.com', server_implementation => 'puppetserver', server_ssl_dir => '/etc/custom/puppet/ssl'}"
+        end
+
+        it 'should put the correct ip address in webserver.conf' do
```
domcleal: This description doesn't appear to be correct.

spec/classes/puppet_server_puppetserver_spec.rb
```diff
@@ -191,6 +191,24 @@
           should contain_file('/etc/custom/puppetserver/conf.d/webserver.conf').with_content(/ssl-host\s+=\s127\.0\.0\.1/)
         end
       end
+
+      describe 'with server_certname parameter given to the puppet class' do
+        let(:params) do
```
domcleal: Is params used if pre_condition's given?

domcleal:
> I think puppet-foreman and puppet-foreman_proxy have the same assumption for their certificates. Have you checked those too?

I think they assume $::fqdn, which suggests this can be changed to $::clientcert - _if_ the case sensitivity is correct, I've not verified yet.  If it's lower case for an upper case FQDN, then we should change them.

I think the point of this change though is to permit server_certname to change without also needing to override these file parameters, which isn't a parameter in other modules.

roidelapluie:
@domcleal my point is to use a different cert for the puppet server and the agent that runs on that node. In my setup the certname of the agent is FQDN and the certname on the server is its IP address.

spec/classes/puppet_server_puppetserver_spec.rb
roidelapluie: apparently yes

roidelapluie:
Upstream foreman bug #6352 looks related

roidelapluie:
@domcleal ran the following test in a EL7 box:
```
[root@CAPS ~]# facter fqdn
CAPS.EXAMPLE.COM
[root@CAPS ~]# puppet config print|grep certname
certname = caps.example.com
[root@CAPS ~]# 
```

roidelapluie:
certname is always lowercase

https://github.com/puppetlabs/puppet/blob/9b80121f10ca66939ee9acd3f469c6c615607da6/lib/puppet/defaults.rb#L636

domcleal:
:+1: subject to tests, looks good and makes sense.

ekohl:
:+1:

domcleal:
Thanks @roidelapluie!

PR #394, opened 2016-04-22
Ensure `$puppet::server_ssl_dir` is created before managing `private_…

tdevelioglu:
This should prevent the following error when changing the location of server_ssl_dir:
```
Error: Cannot create /var/lib/puppet/ssl/private_keys; parent directory /var/lib/puppet/ssl does not exist
Error: /Stage[main]/Puppet::Server::Config/File[/var/lib/puppet/ssl/private_keys]/ensure: change from absent to directory failed: Cannot create /var/lib/puppet/ssl/private_keys; parent directory /var/lib/puppet/ssl does not exist
```


domcleal:
:+1:

mmoll:
merged, thanks @tdevelioglu!

PR #402, opened 2016-05-24
fixes #15132 - support services.d/ configuration in Puppet Server 2.5

domcleal:
PS 2.5 moves CA configuration from bootstrap.cfg to a services.d/ca.cfg
file, separating user config (CA) from app config (e.g. versioned code
service). This is now configured instead of bootstrap.cfg when the
server_puppetserver_version parameter is set to 2.5.

A special version value of 2.4.99 (now the default) configures both the
2.4 and 2.5 files simultaneously, creating both bootstrap.cfg and
services.d/ca.cfg so the module works across both versions, albeit with
logged warnings about duplicate CA definitions. This ensures the module
will still work on the release of 2.5, after which time the parameter
default can be changed to 2.5.0 to remove the duplicate configuration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,9 +14,10 @@ The Puppet master is configured under Apache and Passenger by default, unless
 to switch to the JVM-based Puppet Server.
 
 When using Puppet Server 2 (version 2.0 was the first version to support Puppet 4),
-the module supports and assumes you will be installing the latest version (currently 2.3.1).
-If you know you'll be installing an earlier version, you will need to override
-`server_puppetserver_version`.
+the module supports and assumes you will be installing the latest version.
+If you know you'll be installing an earlier or specific version, you will
+need to override `server_puppetserver_version`. More information in the Puppet
+Server section below.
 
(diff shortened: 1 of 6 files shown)


alexjfisher:
LGTM

Tested on EL7 with server 2.3.2 and puppet 4.4.

brandonweeks:
Working perfectly with puppetserver 2.4 as well. 👍

mmoll:
Does not work with PS 2.4.0 on Debiab/jessie when setting the version to 2.4.99, I'm unsure what's best to deal with the situation, but managing the dir with the same owner/group/permission as coming with the package should be OKish? `Exception in thread "main" java.lang.IllegalArgumentException: Specified bootstrap config file does not exist: '/opt/puppetlabs/server/apps/puppetserver/config/services.d/'`

domcleal:
@mmoll thanks for catching that, I was able to reproduce it. I may have had some left over directories on my test installation.

I've updated the PR to create the services.d/ dir and its parent under /opt.

mmoll:
merged, thanks @domcleal!

PR #405, opened 2016-05-26
Replace UTF-8 apostrophe [skip ci]

domcleal:
The UTF-8 character has triggered a few interesting bugs, including
Kafo's #14473 and #14932, and also triggers PUP-5995 in earlier Puppet 4
releases which prevents the installer working with Puppet Strings as the
default values can't be determined.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -227,7 +227,7 @@
 # $remove_lock::                      Remove the agent lock when running.
 #                                     type:boolean
 #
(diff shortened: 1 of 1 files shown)


ekohl:
Cherry-picked as https://github.com/theforeman/puppet-puppet/commit/138fff6197cb9a93c900aa2d2b0787b10d2c3781, thanks!

PR #406, opened 2016-05-26
Permit access to v3 resource_type API on Puppet 4

domcleal:
When using Puppet 4 without Puppet Server (i.e. Rack configuration),
the resource_types API is still required by the smart proxy and is
configured in the regular auth.conf.


ekohl:
:+1:

mmoll:
merged, thanks @domcleal!

PR #408, opened 2016-06-09
Add support for the PassengerRuby setting.

tdevelioglu:
This allows setting the PassengerRuby setting in the vhost configuration for puppet's rack application.


ekohl:
:+1: I'm slightly surprised we didn't already have such a flag, but I must be confusing it with puppet-foreman.

brandonweeks:
LGTM.

domcleal:
Merged, thanks @tdevelioglu!

PR #411, opened 2016-06-15
Remove Puppet 3 endpoints when using Puppet 4

sathieu:
They are not working anyway, except with puppetserver (with legacy-routes-service).

Ref: https://docs.puppet.com/puppetserver/2.1/compatibility_with_puppet_agent.html#details-about-backwards-compatibility

... and add /puppet-ca/v1/certificate_revocation_list/ca endpoint.

### DIFF
--- a/spec/classes/puppet_config_spec.rb
+++ b/spec/classes/puppet_config_spec.rb
@@ -62,10 +62,11 @@ describe 'puppet::config' do
         end
 
         it 'should contain auth.conf' do
-          should contain_file("#{confdir}/auth.conf").with_content(%r{^path /certificate_revocation_list/ca\nmethod find$})
           if Puppet.version >= '4.0'
+            should_not contain_file("#{confdir}/auth.conf").with_content(%r{^path /certificate_revocation_list/ca\nmethod find$})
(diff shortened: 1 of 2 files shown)


mmoll:
@brandonweeks iirc you were using Puppet 3 agents against Puppetserver 2.x, would that break it?

domcleal:
> They are not working anyway, except with puppetserver (with legacy-routes-service).

If they work with Puppet Server, wouldn't we want to keep them?

sathieu:
`` If they work with Puppet Server, wouldn't we want to keep them?
They'll still work (not tested) as, according to the doc, the URLs are rewritten with the legacy-routes-service.

It seems to be here: https://github.com/puppetlabs/puppetserver/blob/9ea00dba3b6221ed09fad004315943e3874986b8/src/clj/puppetlabs/services/legacy_routes/legacy_routes_core.clj#L129

mmoll:
@sathieu culd you test this? Also, tests fail.

sathieu:
I have tested and updated rspec.

mmoll:
Still working fine here with a Puppet 3.8 agent against a 2.4 puppetserver.
Merged, merci @sathieu!

PR #412, opened 2016-06-15
Don't deploy site.pp on Puppet >= 4.0

sathieu:
even if puppet#15106 was fixed in Puppet 3.3

### DIFF
--- a/spec/classes/puppet_server_config_spec.rb
+++ b/spec/classes/puppet_server_config_spec.rb
@@ -137,12 +137,6 @@ describe 'puppet::server::config' do
             :mode => '0755',
           })
 
-          should contain_file("#{codedir}/manifests/site.pp").with({
-            :ensure  => 'file',
-            :replace => false,
-            :content => "# site.pp must exist (puppet #15106, foreman #1708)\n",
-          })
-
           should contain_puppet__server__env('development')
           should contain_puppet__server__env('production')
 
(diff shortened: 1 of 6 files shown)


mmoll:
If it's only for that puppet bug, I'd say, we can raise the minimum version in metadata and just remove it, but I'd like to wait for other people's opinions.

domcleal:
I'd also be OK with raising the minimum to 3.3.0 and removing it.  Otherwise this PR's also fine.

ekohl:
I was also thinking about increasing the minimum version. I don't think there's a reason to stay on puppet < 3.3.

mmoll:
@sathieu so, please remove it and bump the minimum Puppet version :)

sathieu:
I've done as @mmoll said, and I've added a second commit dropping $server_manifest_path.

spec/classes/puppet_server_config_spec.rb
```diff
@@ -136,12 +136,7 @@
             :group => nil,
             :mode => '0755',
           })
-
-          should contain_file("#{codedir}/manifests/site.pp").with({
-            :ensure  => 'file',
-            :replace => false,
-            :content => "# site.pp must exist (puppet #15106, foreman #1708)\n",
-          })
+	  end
```
mmoll: :end: :no_good: ;)

domcleal:
:+1: tested on 3.8 and it seems to work fine.

mmoll:
merged, merci @sathieu!

ekohl:
This should be a major version bump, right?

mmoll:
for the removed parameter, true.

PR #414, opened 2016-06-20
support Debian non-AIO Puppet 4.x packages

mmoll:
@sathieu please test
4.5.2-1 packages are in the new queue for unstable at the moment, there were no changes that would affect this PR compared to 4.5.0-4.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -53,12 +53,16 @@ class puppet::params {
   $classfile           = '$statedir/classes.txt'
   $syslogfacility      = undef
   $environment         = $::environment
+
   if versioncmp($::puppetversion, '4.0') < 0 {
-    $aio_package = false
+    $aio_package      = false
+    $deb_naio_package = false
   } elsif $::osfamily == 'Windows' or $::rubysitedir =~ /\/opt\/puppetlabs\/puppet/ {
-    $aio_package = true
+    $aio_package      = true
+    $deb_naio_package = false
   } else {
-    $aio_package = false
+    $aio_package      = false
+    $deb_naio_package = ($::osfamily == 'Debian')
   }
 
   case $::osfamily {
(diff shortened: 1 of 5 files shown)


metadata.json
```diff
@@ -63,7 +63,7 @@
     {
       "operatingsystem": "Fedora",
       "operatingsystemrelease": [
-        "21"
+        "24"
```
ekohl: This change looks unrelated, maybe this needs to be merged to master already?

manifests/params.pp
```diff
@@ -100,7 +103,10 @@
         $bindir            = '/opt/puppetlabs/bin'
       } else {
         $dir               = '/etc/puppet'
-        $codedir           = '/etc/puppet'
+        $codedir           =  $deb_naio_package ? {
+                                true    => '/etc/puppet/code',
+                                default => '/etc/puppet',
```
domcleal: Prefer "false" over "default", then if $deb_naio_package isn't set correctly then it should raise a compilation error.

manifests/params.pp
```diff
@@ -258,7 +264,10 @@
   if $aio_package {
     $client_package = ['puppet-agent']
   } elsif ($::osfamily == 'Debian') {
-    $client_package = ['puppet-common','puppet']
+    $client_package = $deb_naio_package ? {
+      true    => ['puppet-agent','puppet'],
```
domcleal: inconsistent spacing around `,` compared to the line below

metadata.json
mmoll: done.

manifests/params.pp
mmoll: done

manifests/params.pp
mmoll: done

mmoll:
anybody? :)

domcleal:
Works well, merged.  Thanks @mmoll.

Issue #413, opened 2016-06-18
Duplicate resource declaration if user drops their own autosign script/file


nick-george:
I have been deploying my own autosign.sh file via puppet, however, with version 5.0.0 of this module, that is no longer possible because of the code at manifests/server/config.pp:115. 

`Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Duplicate declaration: File[/etc/puppet/autosign.sh] is already declared`

An easy fix is to allow the user to supply the content of the autosign.sh, or don't try to set the mode of the file within this module.

mmoll:
This is fallout of #381 and also touches #370. Would you be willing to flesh out out a pull request for this (just being able to provide the content/source based on some parameter is set to undef or not, or similar. #370 itself wouldn't be needed)?

PR #416, opened 2016-06-27
Allowing the user to supply autosign content template.

nick-george:
Here is a patch that allows the module user to supply the content of the autosign file. It fixes issue GH-413. I attempted to run all the tests against the patch but there must be something wrong with my ruby setup, as "rake spec" spewed tons (thousands) of error messages.  

Cheers, 
Nick


mmoll:
Would it be possible to rewrite this PR so that #362 and #370 can get closed by it? The idea there is that one does provide an array as class parameter, which is getting expanded to file lines. When I saw your original issue, I thought you had the content provided by something like `source => `puppet:///...`, which is probably better when something like a shell script is used.

Regarding the tests, try it like this:
```
bundle install --path vendor/gems
bundle exec rake spec
```

mmoll:
@nick-george ping?

nick-george:
It looks like #427 will allow me to provide a script in array form which isn't ideal, but I can deal with. Shall I cancel this pull request?

mmoll:
@nick-george #427 was merged just now... it doesn't provide a script, but if you could rebase and rework your PR to allow contend as a string or a path to a script as string(?), we'd be happy to review that. As mentioned earlier in #427 I *think* to have an idea of your usecase, but am not sure.

nick-george:
Hi @mmoll, being able to just provide a path to my script (that I provide elsewhere in puppet) would suit me. I'll work on that.

james-powis:
Any hope this is getting any love? the \#Managed By Puppet line prevents passing any script into the autosign array...

nick-george:
Hi @mmoll,
Apologies for the long delay on this. What do you think of changing the following code:
```
if $::puppet::server_ca and ! is_bool($puppet::autosign) {  
  file { $puppet::autosign:
    ensure => file,
    owner  => $puppet::user,
    group  => $puppet::group,
    mode   => $puppet::autosign_mode,
  }
  if !empty($puppet::autosign_entries) {  
    File[$puppet::autosign] {
      content => template('puppet/server/autosign.conf.erb'),
    }
  }
}
```
to be
```
if $::puppet::server_ca and ! is_bool($puppet::autosign) and !empty($puppet::autosign_entries) {
  file { $puppet::autosign:
    ensure => file,
    owner  => $puppet::user,
    group  => $puppet::group,
    mode   => $puppet::autosign_mode,
    content => template('puppet/server/autosign.conf.erb'),
  }
} 
```
This would allow people who want to provide the autosign entries in an array to do that, and those who want to control the content of the file outside this module also can. I don't quick understand the motivation behind setting the File resource, but not providing its content (if the autosign array is empty)? This appears to block anyone from writing a custom autosign script.  

Regards, 
Nick

mmoll:
@nick-george then the ability to provide entires as array in the module will be lost, also see #370 for a lengthy comment about when and how managing this file. If it's only about the "managed by puppet" header, I would be completely fine with removing it.

nick-george:
Hi @mmoll, 

Please forgive me if I'm misreading the code, but doesn't my suggestion still allow the file to be filled in using the array of strings? Please see the last line of the block (content =>) and the updated conditional. 

As far as I'm aware, the effect of my change is that it will now only define the **file resource** in puppet if the array of strings isn't empty. This way, people who want to define this file outside of the module can - just by having an empty autosign_entries. 

Note this suggested change does not affect how the autosign entry appears in the puppet.conf file, or the overloading of the variable (both boolean and string). 

Cheers, 
Nick

mmoll:
@nick-george The problem I referred to in #370 would be there again: We need the default case to only manage the file but not the content for foreman-proxy to be able to do its work correctly.

I think extending the line `if !empty($puppet::autosign_entries) {` to have some more possibilities to control the content in the non-default case is the crucial step to solve this. If you want to cut out the whole file resource in this module (with another variable or similar) that's of course also one way to do it, but I guess it's better to go the way of just controlling the content.

nick-george:
Hi @mmoll, 

Thanks for clearing that up, I now understand why the proxy need to be able to control the file outside of puppet. I have added some additional changes that I believe resolve the functionality. I was going to collapse all my changes into a single commit, and submit that, but it looks like they have all now been added to this PR. Would you like me to close this one and create a new PR with a single commit in it? (or is there a way to modify this PR so that it only has a single commit). I'm not the best when it comes to rebase. 

In the meantime, you can view the proposed changes in this PR. 

Cheers, 
Nick

ekohl:
> Would you like me to close this one and create a new PR with a single commit in it? (or is there a way to modify this PR so that it only has a single commit). I'm not the best when it comes to rebase.

We can squash it all when merging so don't worry about that.

manifests/server/config.pp
```diff
@@ -155,11 +155,25 @@
 
   # autosign file
   if $::puppet::server_ca and ! is_bool($puppet::autosign) {
-    file { $puppet::autosign:
-      ensure => file,
-      owner  => $puppet::user,
-      group  => $puppet::group,
-      mode   => $puppet::autosign_mode,
+    if $::puppet::autosign_content and !empty($puppet::autosign_entries){
```
ekohl: I think the following should work (but untested). The benefit is that you have one place to define the autosign file resource.

```puppet
if $::puppet::server::autosign_content {
  if !empty($::puppet::server::autosign_entries) {
    fail('Cannot set both autosign_content and autosign_entries')
  }
  $autosign_content = $::puppet::server::autosign_content
} elsif !empty($::puppet::server::autosign_entries) {
  $autosign_content = template('puppet/server/autosign.conf.erb')
} else {
  $autosign_content = undef
}

file { $::puppet::server::autosign:
  ensure  => file,
  owner   => $::puppet::server::user,
  group   => $::puppet::server::group,
  mode    => $::puppet::server::autosign_mode,
  content => $autosign_content,
}
```

manifests/init.pp
```diff
@@ -105,6 +105,14 @@
 # $autosign_mode::                          mode of the autosign file/script
 #                                           type:Pattern[/^[0-9]{4}$/]
 #
+#
+# $autosign_content::                       If set, write the autosign file content
```
nick-george: Sure, does it matter that this argument can be overloaded with a string or file() or template()?

nick-george: As you can see in the latest commit, in server.pp I also added 'type:string'. I hope this is OK.

manifests/server.pp
```diff
@@ -18,6 +18,14 @@
 #
 # $autosign_mode::             mode of the autosign file/script
 #
+# $autosign_content::          If set, write the autosign file content
+#                              using the value of this parameter. 
+#                              Cannot be used at the same time as autosign_entries
+#                              For example, could be a string, or
+#                              file('another_module/autosign.sh') or
+#                              template('another_module/autosign.sh.erb')
+#                              type:string
```
mmoll: capital S please

manifests/init.pp
```diff
@@ -105,6 +105,15 @@
 # $autosign_mode::                          mode of the autosign file/script
 #                                           type:Pattern[/^[0-9]{4}$/]
 #
+#
```
mmoll: superfluous line

manifests/server.pp
mmoll: hm, we don't have new style typehints at that subclass anyway, so leave it as is, sorry

manifests/server/config.pp
```diff
@@ -154,17 +154,23 @@
...
-      group  => $puppet::group,
-      mode   => $puppet::autosign_mode,
-    }
-    if !empty($puppet::autosign_entries) {
-      File[$puppet::autosign] {
-        content => template('puppet/server/autosign.conf.erb'),
+  if $::puppet::server_ca and ! is_bool($puppet::server::autosign){
+    if $::puppet::server::autosign_content {
+      if !empty($::puppet::server::autosign_entries) {
+        fail('Cannot set both autosign_content and autosign_entries')
```
mmoll: could you add a test for this? grep for `it { should raise_error` in the existing tests.

mmoll:
I think we're coming to grips with that topic ❤️

manifests/server/config.pp
ekohl: (done)

nick-george:
At this point, should I assume that one of your project leads will take care of merging this PR? 

Cheers, 
Nick

mmoll:
@nick-george yes, once I find the time...

mmoll:
merged, thanks a lot @nick-george!

Issue #396, opened 2016-04-25
server_ca_proxy doesn't work with http vhost.


falencastro:
Hi,

Currently redirection from the http vhost to a https ca, through server_ca_proxy, doesn't work, as SSLProxyEngine On, is not being set.

Workaround is setting ssl_proxyengine on manifests/server/passenger.pp

```puppet
apache::vhost { 'puppet-http':

      docroot         => "${app_root}/public/",
      directories     => $directories_http,
      port            => $http_port,
      ssl_proxyengine => $ssl_proxyengine, # <------
      custom_fragment => join([
          $custom_fragment ? {
            undef   => '',
            default => $custom_fragment
          },
          'SetEnvIf X-Client-Verify "(.*)" SSL_CLIENT_VERIFY=$1',
          'SetEnvIf X-SSL-Client-DN "(.*)" SSL_CLIENT_S_DN=$1',
        ], "\n"),
      options         => ['None'],
      require         => Class['::puppet::server::rack'],
    }
```

mmoll:
Could you get a bit more into detail in what configurations this problem does occour and/or open a pull request with the changes for discussion?

falencastro:
Sure, a little more context: My F5 LoadBalancers are offloading SSL in front of my puppetmasters which are running on plain http on port 8139 (server_http => true). 

On the other hand my CA server is still running on just https/8140, so when I set server_ca_proxy to a https address, I end up with an internal server error, and apache complains on logs ([Wed May 11 19:23:48.251305 2016] [ssl:error] [pid 25742] [remote 172.xx.xxx.xx:8140] AH01961: SSL Proxy requested for default:80 but not enabled [Hint: SSLProxyEngine]).

So with the ssl_proxyengine => $ssl_proxyengine line also on the http vhost, apache is able to forward http->https properly.

I'll see about the pull request.

PR #417, opened 2016-06-27
Support http to https proxying

falencastro:
Puppet masters can't forward requests to a https puppet CA server
using 'server_ca_proxy', when they're running on plain http, HTTP
500 is presented. Fixes [GH-396](https://github.com/theforeman/puppet-puppet/issues/396).

* Enables SSLProxyEngine directive on puppet-http vhost

### DIFF
--- a/manifests/server/passenger.pp
+++ b/manifests/server/passenger.pp
@@ -133,6 +133,7 @@ class puppet::server::passenger (
       docroot                 => "${app_root}/public/",
       directories             => $directories_http,
       port                    => $http_port,
+      ssl_proxyengine         => $ssl_proxyengine,
       custom_fragment         => join([
           $custom_fragment ? {
             undef   => '',


mmoll:
Looks OK to me, could you extend the tests in spec/classes/puppet_server_passenger_spec.rb to also check the http vhost?

mmoll:
@falencastro ping?

falencastro:
Hi @mmoll, I'm not very familiar with spec tests, but I think both https and http vhost tests are being done already. https://github.com/theforeman/puppet-puppet/blob/master/spec/classes/puppet_server_passenger_spec.rb#L92-L97

ekohl:
@falencastro that block is what we mean. If you could check there if ssl_proxyengine is passed correctly then we know the code works and keeps working. Another place where it's probably good to check is in https://github.com/theforeman/puppet-puppet/blob/e53f29ccb0d0db89b83d31ee5109a94142f49262/spec/classes/puppet_server_passenger_spec.rb#L44-L57

PR #418, opened 2016-06-29
puppetserver AIO uses a different vardir

mmckinst:
puppetserver runs as the puppet user, the puppet agent normally runs as
the root user. When the two try to share the same vardir,
puppetserver (running as the puppet user) can't always read/write/create
directories if the puppet agent creates them as the root user. The
problem is described more thoroughly in
https://tickets.puppetlabs.com/browse/SERVER-357 .

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -528,7 +528,6 @@
 #                                     type:boolean
 #
 # $server_puppetserver_vardir::       The path of the puppetserver var dir
-#                                     Defaults to '/opt/puppetlabs/server/data/puppetserver'
 #                                     type:string
 #
 # $server_puppetserver_dir::          The path of the puppetserver config dir
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -527,6 +527,10 @@
 #                                     disable in case CA is delegated to a separate instance
 #                                     type:boolean
 #
+# $server_puppetserver_vardir::       The path of the puppetserver var dir
+#                                     Defaults to '/opt/puppetlabs/server/data/puppetserver'
+#                                     type:string
+#
 # $server_puppetserver_dir::          The path of the puppetserver config dir
 #                                     type:string
 #
@@ -679,6 +683,7 @@ class puppet (
   $server_implementation           = $puppet::params::server_implementation,
   $server_passenger                = $puppet::params::server_passenger,
   $server_puppetserver_dir         = $puppet::params::server_puppetserver_dir,
+  $server_puppetserver_vardir      = $puppet::params::server_puppetserver_vardir,
   $server_puppetserver_version     = $puppet::params::server_puppetserver_version,
   $server_service_fallback         = $puppet::params::server_service_fallback,
   $server_passenger_min_instances  = $puppet::params::server_passenger_min_instances,
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -23,15 +23,17 @@ class puppet::params {
   $show_diff           = false
   $module_repository   = undef
   if versioncmp($::puppetversion, '4.0') < 0 {
-    $configtimeout           = 120
-    $server_puppetserver_dir = '/etc/puppetserver'
-    $server_ruby_load_paths  = []
-    $server_jruby_gem_home   = '/var/lib/puppet/jruby-gems'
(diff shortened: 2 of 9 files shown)


manifests/params.pp
```diff
@@ -23,15 +23,17 @@
   $show_diff           = false
   $module_repository   = undef
   if versioncmp($::puppetversion, '4.0') < 0 {
-    $configtimeout           = 120
-    $server_puppetserver_dir = '/etc/puppetserver'
-    $server_ruby_load_paths  = []
-    $server_jruby_gem_home   = '/var/lib/puppet/jruby-gems'
+    $configtimeout              = 120
+    $server_puppetserver_dir    = '/etc/puppetserver'
+    $server_puppetserver_vardir = undef
```
ekohl: I think we also support puppetserver 1 which is puppet 3. Won't this break and should it be ```$vardir```?

manifests/init.pp
```diff
@@ -527,6 +527,10 @@
 #                                     disable in case CA is delegated to a separate instance
 #                                     type:boolean
 #
+# $server_puppetserver_vardir::       The path of the puppetserver var dir
+#                                     Defaults to '/opt/puppetlabs/server/data/puppetserver'
```
ekohl: Given we have multiple defaults depending on the version, I think we should leave out this line

mmckinst:
@ekohl I think 5519203 should work in terms of puppetserver 1.

domcleal:
It looks good generally, but it doesn't work with Foreman's ENC as the facts YAML files now move over to the master vardir, while foreman::puppetmaster is configured to use the agent vardir.

I think updating foreman::puppetmaster's puppet_home parameter in puppet::server::config to point to the master dir should fix it.

spec/classes/puppet_server_config_spec.rb
```diff
@@ -15,29 +15,31 @@
...
+        vardir              = '/opt/puppetlabs/puppet/cache'
+        puppetserver_vardir = '/opt/puppetlabs/server/data/puppetserver'
+        ssldir              = '/etc/puppetlabs/puppet/ssl'
+        sharedir            = '/opt/puppetlabs/puppet'
+        etcdir              = '/etc/puppetlabs/puppet'
+        puppetcacmd         = '/opt/puppetlabs/bin/puppet cert'
+        additional_facts    = {:rubysitedir => '/opt/puppetlabs/puppet/lib/ruby/site_ruby/2.1.0'}
       end
 
       if os_facts[:osfamily] == 'FreeBSD'
```
domcleal: I think you need to set `puppetserver_vardir` here too, as there are some test failures:

<pre>
  1) puppet::server::config on freebsd-10-amd64 with no custom parameters should set up the ENC

     Failure/Error: })

       expected that the catalogue would contain Class[foreman::puppetmaster] with puppet_home set to "/var/lib/puppet" but it is set to "/var/puppet"
</pre>

manifests/server/config.pp
```diff
@@ -192,7 +192,7 @@
     class {'::foreman::puppetmaster':
       foreman_url    => $::puppet::server::foreman_url,
       receive_facts  => $::puppet::server::server_facts,
-      puppet_home    => $puppet::vardir,
+      puppet_home    => $::puppet::server::server_puppetserver_vardir,
```
ekohl: Shouldn't this be ```$::puppet::server::puppetserver_vardir```?

mmoll:
I opended mmckinst/puppet-puppet#1 with fixes to the tests.

mmckinst:
@mmoll thanks so much for taking care of that!

manifests/server/puppetserver.pp
```diff
@@ -25,6 +25,9 @@
 # * `server_puppetserver_dir`
 # Puppetserver config directory
 #
+# * `server_puppetserver_vardir`
+# Puppetserver config directory
```
mmoll: that comment needs updating

mmoll:
One small nitpick, :+1: otherwise.
Tested and found working.

mmoll:
:+1:

ekohl:
Thanks! I took the liberty of squash all commits, with @mmoll's permission.

PR #419, opened 2016-06-30
Add archlinux distro support

simonsd:
Allowing aio_package to be overridden from the main ::puppet class seems to be a bit of a pickle, so just a quick fix for Archlinux.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -206,17 +206,10 @@ class puppet::params {
   $server_http_allow          = []
 
   # use puppetserver (JVM) or puppet master (Ruby)?
-  case $::osfamily {
-    'Archlinux' : {
-      $server_implementation = 'puppetserver'
-    }
-    default : {
-      if $aio_package {
-        $server_implementation = 'puppetserver'
-      } else {
-        $server_implementation = 'master'
-      }
-    }
+  if $aio_package or ($::osfamily == 'Archlinux') {
+    $server_implementation = 'puppetserver'
+  } else {
+    $server_implementation = 'master'
   }
 
   # Need a new master template for the server?
@@ -347,17 +340,9 @@ class puppet::params {
   $lower_fqdn              = downcase($::fqdn)
   $server_foreman          = true
   $server_facts            = true
-  case $::osfamily {
-    'Archlinux' : {
-      $server_puppet_basedir = '/usr/lib/ruby/vendor_ruby/2.3.0/puppet'
-    }
-    default : {
-      if $aio_package {
-        $server_puppet_basedir = '/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet'
-      } else {
-        $server_puppet_basedir = undef
-      }
-    }
+  $server_puppet_basedir   = $aio_package ? {
+    true  => '/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet',
+    false => undef,
   }
   $server_foreman_url      = "https://${lower_fqdn}"
   $server_foreman_ssl_ca   = undef
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -57,7 +57,7 @@ class puppet::params {
   if versioncmp($::puppetversion, '4.0') < 0 {
     $aio_package      = false
     $deb_naio_package = false
-  } elsif $::osfamily == 'Windows' or $::rubysitedir =~ /\/opt\/puppetlabs\/puppet/ {
+  } elsif $::osfamily in ['Archlinux','Windows'] or $::rubysitedir =~ /\/opt\/puppetlabs\/puppet/ {
     $aio_package      = true
     $deb_naio_package = false
   } else {
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -96,7 +96,7 @@ class puppet::params {
       $dir               = '/etc/puppetlabs/puppet'
       $codedir           = '/etc/puppetlabs/code'
       $logdir            = '/var/log/puppetlabs/puppet'
-      $rundir            = '/var/run/puppetlabs'
+      $rundir            = '/run/puppetlabs'
       $ssldir            = '/etc/puppetlabs/puppet/ssl'
       $vardir            = '/opt/puppetlabs/puppet/cache'
       $sharedir          = '/opt/puppetlabs/puppet'
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -57,7 +57,7 @@ class puppet::params {
   if versioncmp($::puppetversion, '4.0') < 0 {
     $aio_package      = false
     $deb_naio_package = false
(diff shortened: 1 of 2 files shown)


ekohl:
@GregSutcliffe mind having a look?

GregSutcliffe:
The OS detection will work, sure, but it will cause larger problems I think. We use the $aio_package flag to set a lot of paths in the "default" section from line 92, but I think those paths will be wrong - eg I have no "/opt/puppetlabs/bin" directory on my system, the binary is /usr/bin/puppet as you would expect.

simonsd:
Good point, didn't see that one.
Actually that's the only path I didn't check.
Will run over all the vars affected by aio_package and see what other issues arise.

GregSutcliffe:
@simonsd I made a start on it here (https://github.com/theforeman/puppet-puppet/pull/384) a while back but got blocked by the tests and didn't finish it - might be of use to you.

simonsd:
Adjusted all necessary parameters manually now instead of relying on the aio_package variable.
Tests seem to work locally, let's see what travis makes of it.

mmoll:
Sorry for hijacking this, but #414 is touching a similar (but inverted) topic, could that get in first and if that's acceptable a variable like `archlinux_naio_package` be introduced and used as determinator?

AFAICT, the Arch packages do use the system Ruby and libraries, so I wouldn't use the term "AIO" here?

GregSutcliffe:
I'd be +1 for standardising our approach to Puppet3-non-AIO packages, for sure.

simonsd:
@mmoll Indeed the term "AIO" is a bit misplaced, apologies. I have renamed the pull request to something (a bit) more appropriate.

As for #414, are there multiple packages available for debian?
Otherwise I don't see the value in the debian_naio_package variable over another osfamily block in params.pp.

ekohl:
Yes, on Debian Testing there's a puppet 4 package which is non-AIO while Debian stable has puppet 3. There's also the puppetlabs repositories which have puppet 4 AIO packages.

mmoll:
@simonsd I'm unsure if I'm getting the question right, but anyway: There are Puppet 4.x AIO packages for Debian/Ubuntu from PuppetLabs or native Puppet 4 Debian packages from debian.org. For the later case there are no Puppet Server packages, but for 4.x using the Rack master under Passenger still works (which is why AIO -> Puppet Server, non-AIO -> Puppet master was implemented for Debian).

What I don't like about the current approach in this PR is the repeated use of something like
```
case $::osfamily {
  'Archlinux' : {}
  default:{}
}
```
I would need to have a closer look onto the situation, maybe re-using `server_implementation` would already improve the situation.

Also, I'd suggest to split this PR into one for Puppet Agent support (where also all the testing with the help of metadata.json/facterdb would need to happen) and then another one for Puppet Server support on Arch.

mmoll:
For the record: I don't advise the re-usage of `$debian_naio_package` for anything other than Debian.

mmoll:
#414 went in

manifests/params.pp
```diff
@@ -88,6 +88,17 @@
       $root_group        = undef
     }
 
+    'Archlinux' : {
+      $dir               = '/etc/puppetlabs/puppet'
+      $codedir           = '/etc/puppetlabs/code'
+      $logdir            = '/var/log/puppetlabs/puppet'
+      $rundir            = '/var/run/puppetlabs'
```
SimonPe: @simonsd can't we just set this to /run/puppetlabs? (/var/run should be a symlink to /run, but i'd rather see the real path here than trough a symlink.)

simonsd: Excellent point, updated the path

manifests/params.pp
```diff
@@ -315,9 +334,17 @@
   $lower_fqdn              = downcase($::fqdn)
   $server_foreman          = true
   $server_facts            = true
-  $server_puppet_basedir   = $aio_package ? {
-    true  => '/opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet',
-    false => undef,
+  case $::osfamily {
+    'Archlinux' : {
+      $server_puppet_basedir = '/usr/lib/ruby/vendor_ruby/2.3.0/puppet'
```
mmoll: if there's no other possible place for this, `$::rubyversion` should be used here

mmoll:
Please mark ArchLinux as supported in metadata.json, so it gets potentially tested, I think facts are there in facterdb...

simonsd:
Adjusted issue title again for clarity, rebased branch on top of current master and cleaned up the code a bit, added archlinux as supported distro in metadata.json as requested.
I think this should be less invasive now, while still retaining all functionality.

If it's ok with all involved parties I'd like to squash the commits into one simpler commit, as I feel it's become a bit of a mess with the back and forth.

metadata.json
```diff
@@ -107,6 +107,9 @@
         "2012",
         "2012 R2"
       ]
+    },
+    {
+      "operatingsystem": "Archlinux",
```
mmoll: the comma on the line end has to go

mmoll:
also: `Evaluation Error: Error while evaluating a Function Call, foo: This module does not support osfamily Archlinux at /home/mmoll/mystuff/gitrepos/theforeman/puppet/puppet-puppet/spec/fixtures/modules/foreman/manifests/params.pp:214:7 on node xxx`

please run the tests locally, if possible before submitting changes, a run in travis is taking very long, especially for puppet-puppet, blocking all our travis slots.

mmoll:
Question as a side-note: was this code actually tested on a real ArchLinux? If yes, also the master/server part?

simonsd:
@mmoll I will try to have a look when I find some time.
Yes this code has been tested on archlinux, though not the server part.

mmoll:
@simonsd perfectly fine then. I saw two non-master/server tests failing, which need to get fixed, otherwise you should just disable the tests for master/server like we do for Windows.

manifests/params.pp
```diff
@@ -195,9 +206,10 @@
   $server_http_allow          = []
 
   # use puppetserver (JVM) or puppet master (Ruby)?
-  $server_implementation = $aio_package ? {
-    true    => 'puppetserver',
-    default => 'master',
+  if $aio_package or ($::osfamily == 'Archlinux') {
+    $server_implementation = 'puppetserver'
```
mmoll: I only agent/client is support for now, I guess this switch can get ommited for now or am I missing something?

simonsd: it could be, yes, though I would prefer it stay in as the default way for running a puppetmaster on archlinux is currently puppetserver

mmoll: how does this setting influence the agent configuration, when only agent configuration is supported?

mmoll: What I'm saying is, that this PR should bring everything needed (not less or more) to configure all the agent related stuff. Master/server support will require in a seperate PR.

simonsd: excellent point, will fix the remaining issues, run tests and squash all commits for clarity.

mmoll:
what's the status here?

mmoll:
closing in favour of #384

PR #420, opened 2016-06-30
Fix Kafo data types on manage_packages/client_package

domcleal:
manage_packages can be either a bool or string data type.


ekohl:
:+1:

mmoll:
merged, thanks @domcleal!

PR #421, opened 2016-07-07
Fix indentation and whitespace in puppet.conf

alexjfisher:
`basemodulepath` setting was indented two extra spaces more than it
should have been.


ekohl:
:+1: on green tests

domcleal:
Thanks @alexjfisher!

PR #422, opened 2016-08-01
Introduce parameter 'server_envs_target' that, when set, turns puppet's …

pkranenburg:
… environments directory into a symbolic link. Useful for referring to remote storage holding the puppet code.

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -1,10 +1,10 @@
 {
-  "name": "desiderius-puppet",
+  "name": "theforeman-puppet",
   "version": "6.0.0",
   "author": "theforeman",
-  "summary": "foreman-puppet clone: Puppet agent and server configuration",
+  "summary": "Puppet agent and server configuration",
   "license": "GPL-3.0+",
-  "source": "git://github.com/pkranenburg/puppet-puppet",
+  "source": "git://github.com/theforeman/puppet-puppet",
   "project_page": "https://github.com/theforeman/puppet-puppet",
   "issues_url": "https://github.com/theforeman/puppet-puppet/issues",
   "description": "Module for installing the Puppet agent and Puppet server",
--- a/metadata.json
+++ b/metadata.json
@@ -1,10 +1,10 @@
 {
-  "name": "theforeman-puppet",
+  "name": "desiderius-puppet",
   "version": "6.0.0",
   "author": "theforeman",
-  "summary": "Puppet agent and server configuration",
+  "summary": "foreman-puppet clone: Puppet agent and server configuration",
   "license": "GPL-3.0+",
-  "source": "git://github.com/theforeman/puppet-puppet",
+  "source": "git://github.com/pkranenburg/puppet-puppet",
   "project_page": "https://github.com/theforeman/puppet-puppet",
   "issues_url": "https://github.com/theforeman/puppet-puppet/issues",
   "description": "Module for installing the Puppet agent and Puppet server",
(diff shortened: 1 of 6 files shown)


ekohl:
Wouldn't it be better to just point the config setting to the right location rather than a symlink?

metadata.json
```diff
@@ -1,10 +1,10 @@
 {
-  "name": "theforeman-puppet",
+  "name": "desiderius-puppet",
```
mmoll: wat? :)

pkranenburg: Sorry,  commit should not be paert of pull-up request; reverted.

--pk
________________________________

Subject: Re: [theforeman/puppet-puppet] Introduce parameter 'server_envs_target' that, when set, turns puppet's … (#422)


In metadata.json<https://github.com/theforeman/puppet-puppet/pull/422#discussion_r73171915>:

> @@ -1,10 +1,10 @@
>  {
> -  "name": "theforeman-puppet",
> +  "name": "desiderius-puppet",


wat? :)

pkranenburg:
If you just point $server_envs_dir there, puppet will try to manage it as a file resource, trying to set ownership and group, etc. This is not always desirable and won't even work if the location is actually on a remote (possibly read-only) filesystem. Such a setup is very useful in a multi-puppetmaster configuration and the local puppet master configuration is much better served by a symbolic link pointing at the remote location.

Cheers, --Paul Kranenburg
________________________________

ekohl:
That makes sense. The alternative is creating a manage_envs_dir parameter so either way we end up with a new parameter. Then I think having something that does manage it is nicer. @theforeman/installer any opinions?

ekohl:
Would you mind adding a testcase where you pass a target?

pkranenburg:
Not sure where to start, but I've made a first attempt in the existing 'puppet_server_env_spec.rb'...

--pk
________________________________
From: Ewoud Kohl van Wijngaarden [notifications@github.com]


Would you mind adding a testcase where you pass a target?

—

mmoll:
the following diff should fix the tests (and also test the default case):
```diff
diff --git a/spec/defines/puppet_server_env_spec.rb b/spec/defines/puppet_server_env_spec.rb
index 6827502..10e3ce1 100644
--- a/spec/defines/puppet_server_env_spec.rb
+++ b/spec/defines/puppet_server_env_spec.rb
@@ -57,6 +57,13 @@ describe 'puppet::server::env' do
           end
 
           it 'should only deploy directories' do
+            should contain_file("#{codedir}/environments").with({
+              :ensure => 'directory',
+              :owner => 'puppet',
+              :group => nil,
+              :mode => '0755',
+            })
+
             should contain_file("#{codedir}/environments/foo").with({
               :ensure => 'directory',
               :owner => 'puppet',
@@ -183,12 +190,13 @@ describe 'puppet::server::env' do
 
         context 'with directory environments link' do
           let :pre_condition do
-            "class {'puppet': server => true, server_envs_target => /foo}"
+            "class {'puppet': server => true, server_envs_target => '/foo'}"
           end
 
           it 'should produce a symbolic link "environments" in codedir' do
-            should be_symlink("#{codedir}/environments").
-              be_linked_to('/foo').
+            should contain_file("#{codedir}/environments").
+              with_target('/foo').
+              with_ensure('link').
               with({}) # So we can use a trailing dot on each with_content line
           end
         end
```

pkranenburg:
Thanks.  It has been pushed.
________________________________

mmoll:
I'm :ok: with this :point_up:

domcleal:
Merged, thanks @pkranenburg!

PR #425, opened 2016-08-24
Fix missing default parameters for strict variable tests

domcleal:


ekohl:
It appears the FreeBSD and Windows tests still fail

domcleal:
Thanks, I think I've fixed that now by defining server_puppetserver_* related parameter defaults for FreeBSD and Windows - set to `undef` where n/a.

ekohl:
:+1: on green tests

mmoll:
merged, thanks @domcleal!

PR #427, opened 2016-08-26
reimplentation of #416 also addressing #413, #370 #362. 

attachmentgenie:
This patch implements a parameter and template to be able to set autosign entries from cod.

also addressing #413, #370 #362.

### DIFF
--- a/templates/server/autosign.conf.erb
+++ b/templates/server/autosign.conf.erb
@@ -1,0 +1,4 @@
+# Managed by Puppet
+<% scope.lookupvar("puppet::server::autosign_entries").each do |entry| -%>
+<%= entry %>
+<% end -%>
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -96,6 +96,11 @@
 #                                     sure that script considers the content of autosign.conf
 #                                     as otherwise Foreman functionality might be broken.
 #
+# $autosign_entries::                 A list of certnames or domain name globs
+#                                     whose certificate requests will automatically be signed.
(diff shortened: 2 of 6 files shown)


mmoll:
This does not implement rolling out a script, which (I think) is the background of #416, but is looking good for the #362 usecase, so I'd like to get this in.

As mentioned in #370, just rolling out that file unconditionally is not what we want as default, as the smart-proxy will normally add its entries to that file and if these get overwritten by puppet we do have a serious problem. I think the contents of that file should *only* be manged, if the array is not an empty one (or something similar).

Also, please add tests.

attachmentgenie:
fair enough comments. this does not address the intention of #416 only your comments to parameterise the content of the file.

I have now added some code to only use the template when autosign_entries is not empty and added some tests to validate this behavior.

mmoll:
:+1:, IMHO this closes #362 and #370

templates/server/autosign.conf.erb
```diff
@@ -0,0 +1,4 @@
+# Managed by Puppet
+<% scope.lookupvar("puppet::server::autosign_entries").each do |entry| -%>
+<%= entry %>
+<% end -%>
```
ekohl: nit: we have line endings in our files

ekohl:
Other than the small nit :+1:

attachmentgenie:
go on then, third time lucky.

ekohl:
:+1:

mmoll:
bedankt @attachmentgenie!

PR #432, opened 2016-09-09
add puppetserver 2.6.x support

mmoll:
See https://docs.puppet.com/puppetserver/2.6/release_notes.html#new-feature-jvm-metrics-endpoint-statusv1services


domcleal:
:+1: tested with 2.6.0 and works fine, thanks.

domcleal:
Merged, thanks @mmoll.

PR #434, opened 2016-09-23
Permit access to environment_classes Puppet Server API

domcleal:
To be used by the smart proxy in #15095.

More info: https://docs.puppet.com/puppetserver/latest/puppet-api/v3/environment_classes.html


mmoll:
merged, thanks @domcleal!

PR #435, opened 2016-09-28
only set puppetmaster var if fact is available

bastelfreak:
we can use defined() or getar() here to determine the presence of the
fact. defined() requires puppet3.6, metadata.json currently mentions at
least 3.3. getvar() requires stdlib 4.4, metadata.json currently
mentions at least 4.2. I implemented defined() because this is already
used two lines higher.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -168,19 +168,21 @@ class puppet::params {
   $auth_allowed = ['$1']
 
   # Will this host be a puppet agent ?
-  $agent                     = true
-  $remove_lock               = true
-  $client_certname           = $::clientcert
+  $agent                      = true
+  $remove_lock                = true
+  $client_certname            = $::clientcert
 
   # Custom puppetmaster
-  if defined('$trusted') and $::trusted['authenticated'] == 'local' {
-    $puppetmaster            = undef
+  # needed due to a PUP-4072
+  # more information in https://github.com/theforeman/puppet-foreman/commit/5fe3239da0c6fbac76172f61042a69ab3a7eb4e6
+  if versioncmp($::puppetversion, '3.7.5') < 0 or defined('$::puppetmaster') {
+    $puppetmaster             = $::puppetmaster
   } else {
-    $puppetmaster            = $::puppetmaster
+    $puppetmaster             = undef
   }
 
   # Hashes containing additional settings
-  $additional_settings   =      {}
+  $additional_settings        = {}
   $agent_additional_settings  = {}
   $server_additional_settings = {}
 
--- a/metadata.json
+++ b/metadata.json
@@ -35,7 +35,7 @@
   "requirements": [
     {
       "name": "puppet",
-      "version_requirement": ">= 3.3.0 < 5.0.0"
+      "version_requirement": ">= 3.6.0 < 5.0.0"
     }
   ],
   "operatingsystem_support": [


manifests/params.pp
```diff
@@ -176,7 +176,11 @@
   if defined('$trusted') and $::trusted['authenticated'] == 'local' {
     $puppetmaster            = undef
   } else {
-    $puppetmaster            = $::puppetmaster
+    if defined('$::puppetmaster') {
```
domcleal: As defined() requires Puppet 3.6 to return true, this will need to be mandatory or the code will stop working on older versions.

Change metadata.json to require 3.6 and remove the obsolete trusted check above, else you'd have to introduce version checks.

bastelfreak:
and updated. Please review again.

domcleal:
The check isn't working on Puppet 3.6/7, I think it's the same issue as fixed in https://github.com/theforeman/puppet-foreman/commit/5fe3239da0c6fbac76172f61042a69ab3a7eb4e6.

I'd suggest a similar check, which would use $::puppetmaster without checking on 3.6 - 3.7.4, but then on newer versions it uses defined() to check for its presence, which allows strict vars to be enabled on 3.7.5 or higher.

Admittedly with a version check the metadata.json change is obsolete, but I don't mind it being increased to 3.6.

bastelfreak:
I updated the code to check the puppet version, As an alternative, we could just drop everything older than 3.8?

manifests/params.pp
```diff
@@ -168,19 +168,21 @@
...
+  $agent                      = true
+  $remove_lock                = true
+  $client_certname            = $::clientcert
 
   # Custom puppetmaster
-  if defined('$trusted') and $::trusted['authenticated'] == 'local' {
-    $puppetmaster            = undef
+  # needed due to a PUP-4072
+  # more information in https://github.com/theforeman/puppet-foreman/commit/5fe3239da0c6fbac76172f61042a69ab3a7eb4e6
+  if if versioncmp($::puppetversion, '3.7.5') < 0 or defined('$::puppetmaster') {
```
domcleal: Two `if`s

domcleal:
Thanks, apart from the one syntax error it's fine by me. I wouldn't mind dropping < 3.8 too much either in master, it would only affect EPEL7 and Jessie (stable, with backports is OK) versions of Puppet from our supported OSes.

manifests/params.pp
```diff
@@ -168,19 +168,21 @@
...
+  $agent                      = true
+  $remove_lock                = true
+  $client_certname            = $::clientcert
 
   # Custom puppetmaster
-  if defined('$trusted') and $::trusted['authenticated'] == 'local' {
-    $puppetmaster            = undef
+  # needed due to a PUP-4072
+  # more information in https://github.com/theforeman/puppet-foreman/commit/5fe3239da0c6fbac76172f61042a69ab3a7eb4e6
+  if versioncmp($::puppetversion, '3.7.5') < 0 or defined('$::puppetmaster') {
```
ekohl: Doesn't this change the behavior? IIRC this was needed in case of masterless setups but I don't have such a setup.

ekohl:
FYI I'm fine with changing the minimum puppet version to 3.6.

manifests/params.pp
domcleal: It was needed for masterless setups with strict variables enabled, but this should have it work in all strict variable setups (from 3.7.5 onwards). I think we preferred this over making it 3.6-only at the time in #237.

In non-strict either masterless or master setups then it should take the `else` branch and leave the parameter set to undef.

ekohl: Sounds like it makes sense.

mmoll:
merged, danke @bastelfreak!

PR #436, opened 2016-09-30
Adding max-requests-per-instance property to puppetserver.conf.

kilianw:
I am looking to add the max-requests-per-instance to puppetserver.conf as described: https://docs.puppet.com/puppetserver/latest/config_file_puppetserver.html

Including the change won't affect any installations unless the property is set explicitly.


domcleal:
> The configuration option is supported since at least version 2.2 which i believe is the oldest we support.

I'm not sure if we support 1.x or not, we used to and haven't explicitly dropped support (though I'd be a little surprised if it's still working). That said, it'd documented in 1.1 so no need for a version check: https://docs.puppet.com/puppetserver/1.1/configuration.html#puppetserverconf

manifests/server/puppetserver.pp
```diff
@@ -63,6 +66,7 @@
   $server_ruby_load_paths      = $::puppet::server::ruby_load_paths,
   $server_cipher_suites        = $::puppet::server::cipher_suites,
   $server_max_active_instances = $::puppet::server::max_active_instances,
+  $server_max_requests_per_instance = $::puppet::server::max_active_instances,
```
domcleal: If I could be a pain and ask you to realign the variables both here and the other files, it'd help when somebody edits it for something else later and feels the need to realign them!

kilianw: No problem at all.

manifests/server/puppetserver.pp
```diff
@@ -52,28 +55,29 @@
...
+  $jvm_extra_args                   = $::puppet::server::jvm_extra_args,
+  $jvm_min_heap_size                = $::puppet::server::jvm_min_heap_size,
+  $jvm_max_heap_size                = $::puppet::server::jvm_max_heap_size,
+  $server_puppetserver_dir          = $::puppet::server::puppetserver_dir,
+  $server_puppetserver_vardir       = $::puppet::server::puppetserver_vardir,
+  $server_jruby_gem_home            = $::puppet::server::jruby_gem_home,
+  $server_ruby_load_paths           = $::puppet::server::ruby_load_paths,
+  $server_cipher_suites             = $::puppet::server::cipher_suites,
+  $server_max_active_instances      = $::puppet::server::max_active_instances,
+  $server_max_requests_per_instance = $::puppet::server::max_active_instances,
```
mmoll: the right side of the `=` was not adjusted after copying

mmoll:
an added test to catch errors would be nice

spec/classes/puppet_server_puppetserver_spec.rb
```diff
@@ -42,6 +42,7 @@
                                           'TLS_RSA_WITH_AES_128_CBC_SHA256',
                                           'TLS_RSA_WITH_AES_128_CBC_SHA', ],
         :server_max_active_instances => 2,
+        :server_max_requests_per_instance => 10000,
```
mmoll: that's not a "default_param" :)

spec/classes/puppet_server_puppetserver_spec.rb
```diff
@@ -100,6 +101,10 @@
             content = catalogue.resource('file', '/etc/custom/puppetserver/conf.d/puppetserver.conf').send(:parameters)[:content]
             expect(content).to include(%Q[    master-var-dir: /opt/puppetlabs/server/data/puppetserver\n])
           end
+          it 'should have max-requests-per-instance: /opt/puppetlabs/server/data/puppetserver' do
```
mmoll: I know it's a bit overengineered, but please add a seperate test for this with a non-zero value and check for the zero here.

mmoll:
two comments reagding the test, :+1: otherwise

mmoll:
Test failures are unrelated, caused by the latest change in puppetlabs-stdlib.

ekohl:
Thanks!

PR #438, opened 2016-10-02
External SSL Termination with Apache for Puppetserver

superfantasticawesome:
This functionality exists for puppet master configurations in the puppet-puppet module but is not available for puppetserver. Puppet Labs' documentation on configuring compile masters with external SSL termination is here: https://docs.puppet.com/puppetserver/latest/external_ssl_termination.html

This code is tested and working on OracleLinux 6.8. 
I'd appreciate any feedback/suggestions on making this more Foreman-friendly.

Sample code to configure a compile master:

```
  class { '::puppet':
    agent                       => true,
    server                      => true,
    server_ca                   => false,
    ca_server                   => "<fqdn of master of masters>",
    ca_port                     => '8140',
    server_implementation       => 'puppetserver',
    server_foreman              => false,
    server_http                 => true,
    server_http_port            => '8139',
  }
```

### DIFF
--- a/manifests/server/reverseproxy.pp
+++ b/manifests/server/reverseproxy.pp
@@ -52,14 +52,14 @@ class puppet::server::reverseproxy (
     ],
     proxy_pass_match     => [
       {
-        'path'           => '^/.*/certificate.*/',
-        'url'            => "https://${ca_server}:${ca_port}",
-        'reverse_urls'   => "https://${ca_server}:${ca_port}",
+         'path'          => '^/.*/certificate.*/',
+          'url'          => "https://${ca_server}:${ca_port}",
+          'reverse_urls' => "https://${ca_server}:${ca_port}",
       },
       {
-        'path'           => '/',
-        'url'            => "http://localhost:${http_port}",
-        'reverse_urls'   => "http://localhost:${http_port}",
+          'path'         => '/',
+          'url'          => "http://localhost:${http_port}",
+          'reverse_urls' => "http://localhost:${http_port}",
       },
     ],
     require              => File["$confdir/public"],
--- a/manifests/server/reverseproxy.pp
+++ b/manifests/server/reverseproxy.pp
@@ -15,8 +15,7 @@ class puppet::server::reverseproxy (
   $http_port             = $::puppet::server::http_port,
   $confdir               = $::puppet::server::dir,
   $ca_port               = $::puppet::ca_port,
-  $ca_server             = $::puppet::ca_server,
-  $vardir                = $::puppet::vardir
+  $ca_server             = $::puppet::ca_server
 ) {
   include ::apache
 
--- a/manifests/server/reverseproxy.pp
+++ b/manifests/server/reverseproxy.pp
@@ -20,7 +20,7 @@ class puppet::server::reverseproxy (
 ) {
   include ::apache
   
-  file { "$confdir/public":
+  file { "${confdir}/public":
     ensure => directory,
   }
  
@@ -30,7 +30,7 @@ class puppet::server::reverseproxy (
     servername           => "$fqdn",
     vhost_name           => '*',
     priority             => false,
-    docroot              => "$confdir/public",
+    docroot              => "${confdir}/public",
     port                 => $port,
     ssl                  => true,
     ssl_cert             => $ssl_cert,
@@ -63,6 +63,6 @@ class puppet::server::reverseproxy (
         'reverse_urls'   => "http://localhost:${http_port}",
       },
     ],
-    require              => File["$confdir/public"],
+    require              => File["${confdir}/public"],
   }
 }
--- a/manifests/server/reverseproxy.pp
+++ b/manifests/server/reverseproxy.pp
@@ -1,0 +1,67 @@
+# == Class: puppet::server::reverseproxy
+#
+# Set up apache to proxy certificate requests to the 
+# designated ca server.
+#
+class puppet::server::reverseproxy (
+  $port                    = $::puppet::server::port,
+  $ssl_ca_cert             = $::puppet::server::ssl_ca_cert,
+  $ssl_ca_crl              = $::puppet::server::ssl_ca_crl,
+  $ssl_cert                = $::puppet::server::ssl_cert,
+  $ssl_cert_key            = $::puppet::server::ssl_cert_key,
+  $ssl_chain               = $::puppet::server::ssl_chain,
+  $ssl_dir                 = $::puppet::server::ssl_dir,
+  $http                    = $::puppet::server::http,
+  $http_port               = $::puppet::server::http_port,
+  $confdir                 = $::puppet::server::dir,
+  $ca_port                 = $::puppet::ca_port,
+  $ca_server               = $::puppet::ca_server,
+) {
+  include ::apache
+
+  file { "$confdir/public":
+    ensure => directory,
+  }
+  
+  ::apache::listen {'8140':}
+  
+  ::apache::vhost { 'puppetserver-reverse-proxy':
+    servername             => "$fqdn",
+    vhost_name             => '*',
+    priority               => false,
+    docroot                => "$confdir/public",
+    port                   => $port,
+    ssl                    => true,
+    ssl_cert               => $ssl_cert,
+    ssl_key                => $ssl_cert_key,
+    ssl_ca                 => $ssl_ca_cert,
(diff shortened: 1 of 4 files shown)


mmoll:
The puppetserver configuration is about where I was when trying to implement that. However,  normally it would be wanted to have the master serve requests by HTTP with *external* SSL termination, which (AFAICT from the Puppet docs) is leaving the puppetserver port open for all connections, which was why I didn't look into that further, but still seems to be the current state of affairs.

Having the SSL termination part also implemented here is IMHO not what's needed for most people wanting to use HTTP or do I miss a specific usecase here?

superfantasticawesome:
@mmoll The backend service listens on 127.0.0.1 and all connections to the backend service are CA-verified TLS.

mmoll:
@superfantasticawesome what if I want to offload the whole SSL dance onto my fast, shiny external loadbalander?

superfantasticawesome:
@mmoll PS: I've never used Travis before, can you help me get a build to passing state?

mmoll:
@superfantasticawesome What's failing at the moment are only the lint checks. However, you can (and should, because the real spec tests are taking hours on the limited travis quota we have) replicate everything on your machine. Assuming you do have a working Ruby and bundler, it's just (in the puppet-puppet dir):
```
bundle install -path vendor/gems
bundle exec rake
```
To configure specific Puppet versions or limit the test run to specific OSes, see .travis.yml for the values you can set regarding PUPPET_VERSION and ONLY_OS.

superfantasticawesome:
@mmoll The use case for offloading to an external LB is still possible. The "reverseproxy" subclass is only included if:
http = true
ca_server != ''
ca_port != ''

If you were just to set http = true (and not set the ca parameters) then puppetserver would be configured to listen on the desired httpd_port. Additionally, as a consequence of setting "http = true", the /conf.d/auth.conf file would have the allow-header-cert-info directive set to true... which is required for external SSL termination and puppetserver.

mmoll:
@superfantasticawesome Point taken on that, if that get in, it needs to get documented in the "Advanced Scenarios" section of the REAFDME.

However, if I get that correctly, the http server of puppetserver is always listening on 127.0.0.1, so it's never exposed as non-https wo the outside world. How would an external SSL LB connect to the puppetserver backend?

Also, the non-puppetserver implementation can have http and https configured simultaneously, so that else clause in the template is not retaining the current behaviour. As said, I was reluctant back then to configure it that way but as it seems puppetserver has no way to implement rules like we do in Apache for the `server_http_allow` parameter so it has to be open that way...

mmoll:
b6e4249 was what I did back then

superfantasticawesome:
Yes, that was my concern when I decided to set the listen address to 127.0.0.1... THe only workaroundsfor this that I can think of is to either set up an http apache proxy in front with allowed hosts directives configured... OR simply modify the rules in conf.d/auth.conf using his module: https://github.com/puppetlabs/puppetlabs-puppet_authorization/blob

superfantasticawesome:
Gotcha. I will set that.

superfantasticawesome:
So the more I think about it, the more appealing it is to leverage auth.conf rules for this... it could read directly from the "server_http_allow" array and configure the appropriate rules in auth.conf.

Thoughts?

mmoll:
Personally (without having taken a deeper look) I wouldn't be against using a new module, but that would necessitate to migrate all the contents and logic we do have in `templates/server/puppetserver/conf.d/auth.conf.erb` before taking a stab on the http topic.

superfantasticawesome:
I'm happy to do that bit of work; e.g., migrate the existing logic employed in auth.conf to a module-based solution. Personally, I feel that's a better approach to managing auth.conf anyway, as it fits more neatly into a n overall configuration management paradigm... that is, if you allow for changes to one area of configuration for a given technology, you should also allow it universally to the same technology; in this case, Puppet's authorization rules. Also, by providing a capability for atomic changes, there's more control for auditing, etc.

I'll get started on it.

superfantasticawesome:
@mmoll Bit of a thorny rose that auth.conf template with all the disparate versions of puppetserver. Are you planning continued support for older versions of puppetserver? I realize that's a moving target but at some point it will be more trouble than it's worth. I initially solved the issue during my recent development efforts by version locking puppetserver at 2.5.0, but what's the gameplan for this type of support in the future?

mmoll:
@superfantasticawesome  We'd like to keep puppetserver compatibility from 2.2 onwards, also as we're probably having to deal with such changes per version in the future anyway

superfantasticawesome:
I've got something in the works for authorization. I'll submit a new PR for
it shortly. If it gets merged I can use that as a basis for introducing a
proper CA proxy for catalog masters with the options we've discussed here.

On Oct 3, 2016 4:26 PM, "Michael Moll" <notifications@github.com> wrote:

@superfantasticawesome <https://github.com/superfantasticawesome> We'd like
to keep puppetserver compatibility from 2.2 onwards, also as we're probably
having to deal with such changes per version in the future anyway

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/theforeman/puppet-puppet/pull/438#issuecomment-251216647>,
or mute the thread

mmoll:
As this PR already has 17 commits and after the latest merge to our current master also merge conflicts, I'd really ask you to switch away from the GitHub web editor and also to use a topic branch in your fork, please see the docs on help.github.com for further information.

Furthermore, at least all the spec tests regarding auth.conf that exist now would need to be backed with new tests. Version dependen changes are also missing now, AFAICT?

superfantasticawesome:
Gotcha

superfantasticawesome:
@mmoll Is there someplace where I can make a donation to offset the travis quota I chewed through?

mmoll:
@superfantasticawesome we don't have a paid travis account, so everything is fine ;)
When the lint checks are failing, it's only a matter of about 10 minutes, because spec tests are not going to get executed in such a case.

manifests/server/reverseproxy.pp
```diff
@@ -0,0 +1,67 @@
...
+    ssl_honorcipherorder => 'on',
+    ssl_verify_client    => 'optional',
+    ssl_options          => '+StdEnvVars +ExportCertData',
+    ssl_verify_depth     => '1',
+    ssl_proxyengine      => true,
+    proxy_preserve_host  => true,
+    request_headers      => [
+      'set X-SSL-Subject %{SSL_CLIENT_S_DN}e',
+      'set X-Client-DN %{SSL_CLIENT_S_DN}e',
+      'set X-Client-Verify %{SSL_CLIENT_VERIFY}e',
```
mclueppers: X-Client-Cert is required in Puppetserver 2.x to populate Puppet's TrustedInformation. If that header is missing following WARN can be seen in puppetserver.log

```
2016-10-12 10:19:01,348 INFO  [qtp396374476-70] [puppetserver] Puppet TrustedInformation expected a certificate, but none was given.
```

superfantasticawesome: Thanks for pointing this out. What's the preferred method for decoding this cert info in Apache?

mclueppers: You can have a look at [external SSL termination](https://docs.puppet.com/puppetserver/2.6/external_ssl_termination.html#x-client-cert). According to the info there and the linked [issue](https://tickets.puppetlabs.com/browse/SERVER-217) you need to add

```
<If "-n '%{SSL_CLIENT_CERT}'">
  RequestHeader unset X-Client-Cert
  RequestHeader set X-Client-Cert %{SSL_CLIENT_CERT}s
  RequestHeader edit* X-Client-Cert "BEGIN CERTIFICATE" "-begin-"
  RequestHeader edit* X-Client-Cert "END CERTIFICATE" "-end-"
  RequestHeader edit* X-Client-Cert " " "%0A"
  RequestHeader edit* X-Client-Cert "-end-" "END%20CERTIFICATE"
  RequestHeader edit* X-Client-Cert "-begin-" "BEGIN%20CERTIFICATE"
</If>
```

mmoll:
I'm unsure how to proceed here. For terminating with an **external** loadbalancer, maybe the HTTP mode of puppetserver is already enough...

mmoll:
Closing due to inactivity and merge conflicts.

This would need to get split up in smaller parts anyway, e.g. the usage of puppetlabs-puppet_authorization which would be ok as the current master is Puppet 4.x only now.

Personally I'd still leave Apache out of the game.

Issue #440, opened 2016-10-11
$server_facts is a reserved variable when trusted_server_facts is true


earsdown:
As per [the doc](https://docs.puppet.com/puppet/latest/reference/lang_facts_and_builtin_vars.html#serverfacts-variable) `$server_facts` is a reserved variable if trusted_server_facts is enabled on puppet-servers:

```
Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Resource Statement, Evaluation Error: Error while evaluating a '=' expression, Attempt to assign to a reserved variable name: 'server_facts' at /etc/puppetlabs/code/environments/production/modules/puppet/manifests/params.pp:343:28 on node puppet03.example.org
Warning: Not using cache on failed catalog
Error: Could not retrieve catalog; skipping run
```

Would there be a massive impact to the installer and/or other modules if the variable name was changed to `$server_foreman_facts`?

PR #441, opened 2016-10-11
Deprecating and replacing server_facts parameter

earsdown:
The $server_facts variable name is reserved by puppet-servers when trusted_server_facts is enabled.
Using this variable in a manifest causes compilation to fail with "Attempt to assign to a reserved variable name: server_facts"
This commit deprecates this variable and introduces the boolean $server_foreman_facts as its replacement.
**Partially** fixes GH issue #440

### DIFF
--- a/manifests/server.pp
+++ b/manifests/server.pp
@@ -479,7 +479,6 @@ class puppet::server(
   validate_bool($passenger)
   validate_bool($git_repo)
   validate_bool($service_fallback)
-  validate_bool($server_facts)
   validate_bool($server_foreman_facts)
   validate_bool($strict_variables)
   validate_bool($foreman)
--- a/manifests/server.pp
+++ b/manifests/server.pp
@@ -217,10 +217,6 @@
 #                              approach will be installed.
 #                              type:string
 #
-# $server_facts::              Should foreman receive facts from puppet
-#                              DEPRECATION WARNING: Use server_foreman_facts.
-#                              type:boolean
-#
 # $server_foreman_facts::      Should foreman receive facts from puppet
 #                              type:boolean
 #
@@ -451,7 +447,6 @@ class puppet::server(
   $foreman_ssl_ca            = $::puppet::server_foreman_ssl_ca,
   $foreman_ssl_cert          = $::puppet::server_foreman_ssl_cert,
   $foreman_ssl_key           = $::puppet::server_foreman_ssl_key,
-  $server_facts              = undef,
   $server_foreman_facts      = $::puppet::server_foreman_facts,
   $puppet_basedir            = $::puppet::server_puppet_basedir,
   $puppetdb_host             = $::puppet::server_puppetdb_host,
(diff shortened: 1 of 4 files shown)


earsdown:
Btw, this PR still won't allow us to enable `trusted_server_facts`, but it's a step in the right direction?

manifests/server.pp
```diff
@@ -465,12 +470,16 @@
   $use_legacy_auth_conf      = $::puppet::server_use_legacy_auth_conf,
 ) {
 
+  if $server_facts != $server_foreman_facts {
```
domcleal: This will match when `server_foreman_facts` is changed, it'll force people to set the old parameter too, in order to prevent the warning. Perhaps set the old parameter default to `undef` and prefer it if set, or even remove it entirely (we can do a major release next, perhaps)?

earsdown: Yeah, it would be way better to just remove it entirely. I've done this already in a [separate branch](https://github.com/earsdown/puppet-puppet/tree/remove-server_facts) and am working on tests. Can we skip deprecation and just remove it? Or would you prefer I set it to undef and warn only if set?

earsdown:
How can I write a test that demonstrates the compilation failure when assigning a value to $server_facts when `trusted_server_facts` is enabled? I tried something like this but it runs for every scenario and doesn't fail when it's supposed to:

```ruby
describe 'with trusted_server_facts enabled' do
  context 'on Puppet 4.1.0+', :if => (Puppet.version >= '4.1.0') do
    trusted_server_facts = 'true'
    let :pre_condition do
      "class {'puppet': server => true, server_additional_settings => {trusted_server_facts => true} }"
    end
    it 'should compile' do
      should compile.with_all_deps
    end
  end
end
```

Or is such a test only possible in a real environment?

manifests/server.pp
mmoll: I'm in for removing it completely and bumping the major version.

domcleal:
You would have to call `Puppet[:trusted_server_facts] = true` in a before block, and reset it after I think to test that. Or if you remove it entirely per the inline comment, then a test probably isn't necesary.

mmoll:
@earsdown what's the status here?

earsdown:
@mmoll I'm still waiting on a decision on whether to remove the parameter completely or deprecate it. If we're okay with removing it, check out the "remove-server_facts" branch in my fork. The requested changes have been addressed.

mmoll:
@earsdown It's OK to remove the parameter, please push to the branch this PR originates from and we'll review again. The major version of the master branch has already been bumped, so now is a perfect time to get more API incompatible changes in. :boom:

mmoll:
what would be left to solve #440?

domcleal:
It looks like this PR contains the same as the remove-server_facts branch, and the diff now shows it entirely renames the parameter, so that's fine. I'll just squash it.

domcleal:
Merged, thanks @earsdown.

PR #443, opened 2016-10-17
Remove pre-3.4 support for exec without umask

domcleal:


mmoll:
merged, thanks @domcleal!

PR #444, opened 2016-10-17
fixes #15550 - start Puppet agent after server is running

domcleal:
Ensures no race condition will occur between the generate CA step of the
server configuration and the agent starting up, generating its own
private key.

---

The agent started by `puppet::agent::service` generates its private key _after_ forking (if entropy is low, it might even take a while), and so this might otherwise happen at the same time as `puppet::server::config` calling `puppet cert generate`. If the agent is slower to write to disk, its key overwrites the one used by the generate command to create the new certificate.

### DIFF
--- a/spec/classes/puppet_init_spec.rb
+++ b/spec/classes/puppet_init_spec.rb
@@ -52,7 +52,10 @@ describe 'puppet' do
       end
 
       describe 'with no custom parameters' do
+        it { is_expected.to compile.with_all_deps unless os_facts[:osfamily] == 'windows' }
+        it { should contain_class('puppet::agent') }
         it { should contain_class('puppet::config') }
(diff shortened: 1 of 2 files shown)


spec/classes/puppet_init_spec.rb
```diff
@@ -52,7 +52,10 @@
       end
 
       describe 'with no custom parameters' do
+        it { is_expected.to compile.with_all_deps }
```
mmoll: this fails with the windows OSes in the spec environment

domcleal: I think when running in a Unix environment, the path's being tested with Unix file path expectations too, because that's how Puppet's being run. I've excluded the test from Windows for now.

Issue #445, opened 2016-10-17
'certificate_status' and 'certificate_statuses' need full path in auth.conf


sedunne:
Currently it looks like requests to these two endpoints will fail, even in the request is coming from a whitelisted host.  Adding the full api path seems to allow these to function as intended.

Will have a PR submitted shortly.

PR #446, opened 2016-10-17
add full api path to certificate_status(es) in auth.conf

sedunne:
Fixes GH-445


domcleal:
Merged, thanks @sedunne.

PR #447, opened 2016-10-21
Unpin voxpupuli/extlib

domcleal:


ekohl:
I assume this was pinned because of Ruby 1.8.7

domcleal:
> I assume this was pinned because of Ruby 1.8.7

That's right, 1.0.0 dropped (broke) support, but otherwise it contains few changes at the moment.

ekohl:
:+1: on green tests

mmoll:
merged, thanks @domcleal!

PR #448, opened 2016-10-23
Clean up specs

ekohl:


spec/classes/puppet_config_spec.rb
```diff
@@ -1,18 +1,8 @@
...
 describe 'puppet::config' do
-  on_supported_os.each do |os, os_facts|
-    next if only_test_os() and not only_test_os.include?(os)
-    next if exclude_test_os() and exclude_test_os.include?(os)
+  on_os_under_test.each do |os, facts|
     context "on #{os}" do
-      let (:default_facts) do
-        os_facts.merge({
-          :concat_basedir => '/foo/bar',
-          :domain         => 'example.org',
```
domcleal: I think this might be required to make the tests pass, it looks like srv_domain is set in puppet.conf from $::domain.

ekohl: It's odd, the domain fact doesn't appear to be available at all while you'd expect it to be. I'll look into this.

mmoll:
@ekohl what's the status here? :)

ekohl:
Updated

domcleal:
Thanks @ekohl!

PR #450, opened 2016-11-10
add puppetserver 2.7.x support

mmoll:


mmoll:
@ekohl It does indeed copy the default and the difference in 2.7 is that one can turn it off now. On the other hand, we would be free to change it, I guess... and I would also vote for disabling it.

mmoll:
@domcleal any opinion on disabling the update check per default?

domcleal:
I think it's important for the module to reflect the defaults as closely as possible, so I think it needs to remain true. (Even if I dislike it.)

ekohl:
Looks like we're all in agreement then: we dislike the default but it is the default and the module should have no bias.

domcleal:
Thanks @mmoll.

PR #451, opened 2016-11-10
add puppetserver 2.7.x support

mmoll:
cherry-picked for 6.0-stable

### DIFF
--- a/README.md
+++ b/README.md
@@ -176,7 +176,8 @@ version available.
 
 Currently supported values and configuration behaviours are:
 
-* `2.6.x` (default) - configures the /status API endpoint in `web-routes.conf`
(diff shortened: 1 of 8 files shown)


domcleal:
I don't think this is strictly needed (2.6 mode works with 2.7), but it's OK if you'd like to.

mmoll:
tbh, this time I didn't check if something breaks with the old config... if it's not needed I'd rather close this.

domcleal:
It seems OK to me, the service runs and I can fetch and apply a catalog from it. I'm unsure about the gem_home change, but I can't see any impact from having the 2.6.0 configuration - it may define a default internally if it's even needed.

mmoll:
allright, closing :)

Issue #439, opened 2016-10-04
Atomic changes with (puppetserver) auth.conf using puppetlabs-puppet_authorization


superfantasticawesome:
Currently, the content of auth.conf is supplied via template. This prevents atomic changes to authorization rules, thus prohibiting customizations to server roles.

By implementing [this module](https://github.com/puppetlabs/puppetlabs-puppet_authorization), authorization rules could be added in a more granular fashion, thus giving users access to all authorization rules for more customized deployments of puppetserver.

See this [sample code](https://github.com/superfantasticawesome/puppet-puppet/blob/authorization/manifests/server/authorization.pp) for reference.

superfantasticawesome:
PS: The sample code as is does not expose any parameters, nor has a model been developed to allow user customizations the authorization rules. The sample code merely rewrites the existing erb template in code that the puppet_authorization module could parse and apply as a catalog.

mmoll:
As said in #438, pulling in a new module is OK, but the existing auth.conf rules and version dependent changes need to get transplanted and backed by spec tests.

superfantasticawesome:
@mmoll The puppet_authorization module includes its own battery of spec tests; however, the only tests I can see that are relevant to authorization in the current spec tests are that auth.conf exists and contains version specific lines. Truth be told, I couldn't write a spec test with a gun to my head... but I have added the existing rules and version-specific configurations in this [forked repo](https://github.com/superfantasticawesome/puppet-puppet/blob/authorization/manifests/server/authorization.pp).

Regarding any additional required tests, I would most definitely need some help writing them... but currently the code works as is... and delivers an auth.conf file that filters just as the existing one.

What I'm noodling over now is the use case mentioned in #438 where a loadbalancer is terminating SSL... and determining which resources will need array variables for the allow parameter. I was looking at the legacy auth.conf for examples and was thinking that the "auth_allowed" array could be used for this purpose. Thanks.

ekohl:
I agree that using that module is a good idea because the template shouldn't be this complex. What you can also do is have multiple presets of auth settings. So maybe ```auth/base```, ```auth/standalone``` and ```auth/loadbalanced```. All depending on how big the various if blocks become.

superfantasticawesome:
[...]

PR #452, opened 2016-11-14
allow the user to add custom auth rules in auth.conf without managing the entire template

cristifalcas:
File "${puppet_dir}/auth.conf" was modified to concat resource.
Added a new define puppet::auth.
Update tests.

Usage: 

		  puppet::auth { 'allow all machines to upload their own facts':
		    path   => '/facts/',
		    method => 'save, find, search',
		    auth   => 'yes',
		    allow  => '*',
		  }

### DIFF
--- a/manifests/auth.pp
+++ b/manifests/auth.pp
@@ -1,0 +1,59 @@
+# == Define: puppet::auth
+#
+# === Parameters
+#
+# [*name*]
+#   String::            Name of the auth directive
+#
+# [*path*]
+#   String::            Which URLs the ACL applies to. Required. Must be the first directive in the ACL.
+#                       Default: the value of $name
+#
+# [*environment*]
+#   String::            A valid environment or a comma-separated list of environments. Optional;
+#                       defaults to all environments if omitted.
+#
+# [*method*]
+#   String::            Which HTTP methods the ACL applies to.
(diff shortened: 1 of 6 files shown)


domcleal:
This seems reasonable on the surface, but I don't think it's going to do anything in a Puppet Server configuration as the HOCON auth config is used by default (in this module).

I think the level of abstraction needs to be higher to cover both the old auth.conf and the HOCON version (i.e. #439).

cristifalcas:
But it doesn't change the current status, no?

We are not using puppet 4 yet, so I don't know how to manage the HOCON config.

mmoll:
We do try to manage both configs from the same "source data" in this module, if possible, so this does make a difference right now.

cristifalcas:
It will be acceptable to do the same thing (with concat) for HOCON config?

cristifalcas:
It will be acceptable to do the same thing (with concat) for HOCON config?

mmoll:
@cristifalcas as long as it works correctly I'd say yes, as that would hopefully allow us to refactor further later to have easier possibilities to construct/add auth rules for both formats in order to support HA configurations better.

mmoll:
closing due to inactivity, this module is puppet 4.x only in current master now

PR #456, opened 2016-11-16
fixes #17062 - move master rack restart to puppet::server::service

domcleal:
Includes minor changes to containment of server configuration classes
to ensure notifies of changes from puppet::server::passenger and ::rack
all reach puppet::server::service.

### DIFF
--- a/manifests/server/install.pp
+++ b/manifests/server/install.pp
@@ -31,6 +31,17 @@ class puppet::server::install {
     }
   }
 
+  # Prevent the master service running and preventing Apache from binding to the port
+  if $::puppet::server::passenger and $::osfamily == 'Debian' {
+    file { '/etc/default/puppetmaster':
+      content => "START=no\n",
+    }
+
+    if $::puppet::manage_packages == true or $::puppet::manage_packages == 'server' {
+      File['/etc/default/puppetmaster'] -> Package[$server_package]
+    }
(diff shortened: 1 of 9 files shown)


manifests/server/install.pp
```diff
@@ -31,6 +31,17 @@
     }
   }
 
+  # Prevent the master service running and preventing Apache from binding to the port
+  if $::puppet::server::passenger and $::osfamily == 'Debian' {
+    file { '/etc/default/puppetmaster':
```
ekohl: It feels like this should be in ```config.pp```?

domcleal: Yes, I'd agree except that I think it has to run before the package installation, or it needs reimplementing into debconf (if possible, I've not checked).

It used to be in puppet::config::passenger but had a `before => Class['puppet::server::install']` relationship, which caused a loop when I contained the puppet::config::passenger class. I had to move it here to better match the relationship it had.

ekohl: I had a feeling it was something like that.

domcleal: Sorry, I just noticed that I specified the relationship back to front, which may have confused matters. It now specifies that the config is _before_ the package.

mmoll:
merged, thanks @domcleal!

PR #462, opened 2016-12-03
desupport FreeBSD 9.x

mmoll:
It will be EOLed at 2016-12-31 and there are no version based differences in this module, so this will reduce the test run time a little bit.


domcleal:
Merged, thanks @mmoll.

Issue #465, opened 2016-12-07
Non existent facts / top scope vars on Puppet 4.x


james-powis:
2016-12-07 13:22:12,464 WARN  [qtp1376229403-70] [puppetserver] Puppet Unknown variable: '::domain'. at /etc/puppetlabs/code/environments/production/modules/puppet/manifests/params.pp:37:26
2016-12-07 13:22:12,471 WARN  [qtp1376229403-70] [puppetserver] Puppet Unknown variable: '::puppetmaster'. at /etc/puppetlabs/code/environments/production/modules/puppet/manifests/params.pp:179:32

The domain and puppetmaster facts are non existent, where are these supposed to be sourced from?

ekohl:
Are you running masterless by any chance? I also noticed the domain one wasn't present in some rspec-puppet-facts version so there may have been some chance in facter as well. Could you show the versions of puppet, facter along with your OS? I'm also interested in output of ```hostname``` and ```hostname -f``` to see if they include a domain.

james-powis:
```logs
[root@88662-docker01 ~]# puppet --version
4.5.2
[root@88662-docker01 ~]# facter --version
3.2.0 (commit 148948d795fbdcf6bcf51ed36a73f496775835dc)
[root@88662-docker01 ~]# facter domain

[root@88662-docker01 ~]# hostname
88662-docker01
[root@88662-docker01 ~]# hostname -f
88662-docker01
[root@88662-docker01 ~]# 
```

and to describe why we don't have the domain set is largely because of multi tenancy and hostname conventions of near folklore times... however with strict variables coming I think the appropriate path is check for key existence in $facts and evaluate if undef / null or something like pick($::domain, undef)

james-powis:
and sorry no this is Masterfull Runs

ekohl:
That confirms my suspicions. Would you be able to come up with a PR for this?

james-powis:
Sure no problem

alexjfisher:
@james-powis Rather annoyingly, the `$facts` hash isn't available by default on puppet 3.

Instead, you could use the `defined` function, but that doesn't work properly on puppet < 3.7.5.
Or you can use `getvar('::some_fact')` but that's broken on recent puppet 4 unless stdlib >= 4.13.0

james-powis:
[...]

PR #467, opened 2016-12-11
addressing use cases where $::domain fact is null

james-powis:
Ref: #465 

Some times it is possible for the domain fact to be null
these changes address such an instance, sometimes...

On puppet version < 3.7.5 due to the absence of a defined function we
hope for the best and insist that strict variables is disabled.

Evaluating defined('$::domain') before assigning it to srv_domain

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -41,6 +41,9 @@ class puppet::config(
     puppet::config::main{'dns_alt_names': value => $dns_alt_names; }
   }
   if $use_srv_records {
+    unless $srv_domain {
+      fail('$::domain fact found to be undefined and $srv_domain is undefined')
+    }
     puppet::config::main{
       'use_srv_records': value => true;
       'srv_domain': value => $srv_domain;
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -34,7 +34,17 @@ class puppet::params {
   $postrun_command     = undef
   $dns_alt_names       = []
   $use_srv_records     = false
-  $srv_domain          = $::domain
+
+  # On versions < 3.7.5 the defined function was optional as such strict variables cannot be used
+  # In some edge cases 'domain' fact may be not defined
+  if versioncmp($::puppetversion, '3.7.5') < 0 {
+    $srv_domain = $::domain
+  } elsif defined('$::domain') {
+    $srv_domain = $::domain
+  } else {
+    $srv_domain = undef
+  }
+
   # lint:ignore:puppet_url_without_modules
   $pluginsource        = 'puppet:///plugins'
   $pluginfactsource    = 'puppet:///pluginfacts'
--- a/spec/classes/puppet_config_spec.rb
+++ b/spec/classes/puppet_config_spec.rb
@@ -138,16 +138,45 @@ describe 'puppet::config' do
         end
       end
 
-      describe "when use_srv_records => true" do
-        let :pre_condition do
-          "class { 'puppet': use_srv_records => true }"
+      describe "when use_srv_records => true, and domain fact" do
+        context 'is defined' do
+          let :pre_condition do
+            "class { 'puppet': use_srv_records => true }"
+          end
+
+          it 'should contain puppet.conf [main] with SRV settings' do
+            should contain_puppet__config__main("use_srv_records").with({'value' => "true"})
+            should contain_puppet__config__main("srv_domain").with({'value' => "example.org"})
+            should contain_puppet__config__main("pluginsource").with({'value' => "puppet:///plugins"})
+            should contain_puppet__config__main("pluginfactsource").with({'value' => "puppet:///pluginfacts"})
+          end
         end
 
-        it 'should contain puppet.conf [main] with SRV settings' do
-          should contain_puppet__config__main("use_srv_records").with({'value' => "true"})
-          should contain_puppet__config__main("srv_domain").with({'value' => "example.org"})
-          should contain_puppet__config__main("pluginsource").with({'value' => "puppet:///plugins"})
-          should contain_puppet__config__main("pluginfactsource").with({'value' => "puppet:///pluginfacts"})
+        context 'is unset' do
+          let(:facts) { facts.merge({domain: nil}) }
+          let :pre_condition do
+            'class { ::puppet:
+               use_srv_records => true
+            }'
+          end
+
+          it 'should fail with a helpful message' do
+            should raise_error(Puppet::Error, /\$::domain fact found to be undefined and \$srv_domain is undefined/)
+          end
+        end
+
+        context 'is overriden via param' do
+          let :pre_condition do
+            'class { "::puppet":
+               use_srv_records => true,
+               srv_domain      => "special_domain.com"
+            }'
+          end
+          
+          it 'should configure srv domain' do
+            should contain_puppet__config__main('use_srv_records').with_value(true)
+            should contain_puppet__config__main('srv_domain').with_value('special_domain.com')
+          end
         end
       end
 


spec/classes/puppet_config_spec.rb
```diff
@@ -138,16 +138,46 @@
...
+        context 'is overriden via param' do
+          let :pre_condition do
+            'class { "::puppet":
+               use_srv_records => true,
+               srv_domain      => "special_domain.com"
+            }'
+          end
+
+          it 'should configure srv domain' do
+            should compile.with_all_deps
```
domcleal: You may need to leave this out, since validation of the catalog fails when testing Windows (since it runs on Linux), causing test failures.

spec/classes/puppet_config_spec.rb
```diff
@@ -138,16 +138,46 @@
...
+        context 'is unset' do
+          let(:facts) { facts.merge({domain: nil}) }
+          let :pre_condition do
+            'class { ::puppet:
+               use_srv_records => true
+            }'
+          end
+
+          it 'should fail with a helpful message' do
+            should_not compile
```
domcleal: Check the message here by preferring `should raise_error(Puppet::Error, /.../)`.

manifests/config.pp
```diff
@@ -41,6 +41,9 @@
     puppet::config::main{'dns_alt_names': value => $dns_alt_names; }
   }
   if $use_srv_records {
+    if $srv_domain == undef {
```
domcleal: `unless $srv_domain`

manifests/params.pp
```diff
@@ -34,7 +34,24 @@
...
   $dns_alt_names       = []
   $use_srv_records     = false
-  $srv_domain          = $::domain
+
+  # On versions < 3.7.5 there was no defined function as such strict variables cannot be used
+  # On verisons < 4.0.0 and >= 3.5.0 the facts hash was optional disabled by default, on version < 3.5.0 the facts hash did not exist
+  # In some edge cases 'domain' fact may be not defined
+  if versioncmp($::puppetversion, '3.7.5') < 0 {
+    $srv_domain = $::domain
+  } elsif versioncmp($::puppetversion, '4.0.0') >= 0 {
```
domcleal: Is this block necessary? The `defined('$::domain')` test should work for 4 too.

manifests/params.pp
```diff
@@ -34,7 +34,24 @@
   $postrun_command     = undef
   $dns_alt_names       = []
   $use_srv_records     = false
-  $srv_domain          = $::domain
+
+  # On versions < 3.7.5 there was no defined function as such strict variables cannot be used
```
alexjfisher: There is a defined function, it's just broken for this use case.

manifests/params.pp
james-powis: The problem is it floods the puppetserver log with:
2016-12-07 13:22:12,464 WARN [qtp1376229403-70] [puppetserver] Puppet Unknown variable: '::domain'. at /etc/puppetlabs/code/environments/production/modules/puppet/manifests/params.pp:37:26

spec/classes/puppet_config_spec.rb
james-powis: Sweet tip, thanks!

spec/classes/puppet_config_spec.rb
james-powis: Removed the compile.with_all_deps

manifests/config.pp
james-powis: switched to unless

manifests/params.pp
james-powis: changed veribage

manifests/params.pp
domcleal: The `defined` function call in the conditional shouldn't log that warning, the variable isn't referenced there. Are you sure you're testing the `elsif defined('$::domain')` code on line 49?

The conditional should return false if the variable is undefined and never reference the non-existent variable. This should be identical on Puppet 3 and 4.

james-powis: Sorry I should have been a bit more clear, So the domain error provided above is identical to ::puppetmaster as set by theforeman (in my use case I use the module without foreman)

Here are the 2 entries from a single agent run:
```log
/etc/puppetlabs/code/environments/production/modules/puppet/manifests/params.pp:37:26
2016-12-07 13:22:12,471 WARN [qtp1376229403-70] [puppetserver] Puppet Unknown variable: '::puppetmaster'. at /etc/puppetlabs/code/environments/production/modules/puppet/manifests/params.pp:179:32
```

as you will see $::puppetmaster is wrapped in a defined function, unfortunately for $::puppetmaster I am unable to device a solution for that error, but this PR addresses the $::domain fact.

james-powis: strait from master branch: https://github.com/theforeman/puppet-puppet/blob/master/manifests/params.pp#L37

domcleal: Right, but this comment is about the PR here, not the code in master. This block for the 4.0.0 version test is probably unnecessary, the `elsif defined(..)` test should be sufficient.

james-powis: @domcleal the problem is that the $::domain is not even undef (it is non existent) when facter fails to evaluate. As such while yes defined helps insulate when strict vars is enabled allowing for a successful compile, when it does cause a flooding of the puppetserver logs regardless of strict var being enabled or disabled... 

On 4.x we can test for the existence of the "domain" key in the facts hash and eliminate the useless (and rather annoying) output to the logs. That is why we are testing for 4.x and not using defined on that fact.

james-powis: rather annoyingly pick also fails in this context...

domcleal: > As such while yes defined helps insulate when strict vars is enabled allowing for a successful compile, but because of the absence of the variable entirely it causes a flooding of the puppetserver logs regardless of strict var being enabled or disabled...

Which line causes this? The `defined` check itself? I couldn't reproduce this locally.

domcleal: <pre>
$ puppet --version
4.7.0
$ puppet apply -e "defined('$::domain')"
Notice: Compiled catalog for cobalt in environment production in 0.08 seconds
Notice: Applied catalog in 0.02 seconds
$ puppet apply --modulepath /tmp/mp -e 'pick($::domain, "2nd")'
Warning: Unknown variable: '::domain'.  at line 1:6
Notice: Compiled catalog for cobalt in environment production in 0.06 seconds
Notice: Applied catalog in 0.02 seconds
$ puppet apply --modulepath /tmp/mp -e '$::domain'
Warning: Unknown variable: '::domain'.  at line 1:1
Notice: Compiled catalog for cobalt in environment production in 0.03 seconds
Notice: Applied catalog in 0.03 seconds
</pre>

Only referencing the variable should cause a warning, `defined` should be fine (unless perhaps you're passing the variable into defined rather than the name).

james-powis: domcleal I think you are on to something here... one sec I will amend this PR

james-powis: FYI implemented defined

james-powis:
ready for review

manifests/params.pp
```diff
@@ -34,7 +34,17 @@
   $postrun_command     = undef
   $dns_alt_names       = []
   $use_srv_records     = false
-  $srv_domain          = $::domain
+
+  # On versions < 3.7.5 the defined function was optional as such strict variables cannot be used
+  # In some edge cases 'domain' fact may be not defined
+  if versioncmp($::puppetversion, '3.7.5') < 0 {
```
ekohl: Nit: you can combine this line with the ```elsif```

mmoll: I think it's OK because of the comment above

mmoll:
merged, thanks @james-powis!

PR #468, opened 2016-12-12
fixes #17634 - set vardir, rundir and logdir explicit in puppet.conf

schurzi:
set the vardir, logdir and rundir in puppet.conf, so values are equal to puppetserver/conf.d/puppetserver.conf
see: https://docs.puppet.com/puppetserver/2.7/puppet_conf_setting_diffs.html#overriding-puppet-settings-in-puppet-server

PS:
I have a really hard time testing this with foreman-installer. Help is much appreciated.

### DIFF
--- a/manifests/server/config.pp
+++ b/manifests/server/config.pp
@@ -6,6 +6,21 @@ class puppet::server::config inherits puppet::config {
 
   if $::puppet::server::implementation == 'puppetserver' {
     contain 'puppet::server::puppetserver' # lint:ignore:relative_classname_inclusion (PUP-1597)
+    unless empty($::puppet::server::puppetserver_vardir) {
+      puppet::config::master {
+        'vardir': value => $::puppet::server::puppetserver_vardir;
+      }
+    }
+    unless empty($::puppet::server::puppetserver_rundir) {
+      puppet::config::master {
+        'rundir': value => $::puppet::server::puppetserver_rundir;
+      }
+    }
+    unless empty($::puppet::server::puppetserver_logdir) {
+      puppet::config::master {
+        'logdir': value => $::puppet::server::puppetserver_logdir;
+      }
+    }
   }
 
   # Mirror the relationship, as defined() is parse-order dependent
@@ -69,6 +84,7 @@ class puppet::server::config inherits puppet::config {
     'parser':             value => $::puppet::server::parser;
     'strict_variables':   value => $::puppet::server::strict_variables;
   }
+
   if $::puppet::server::ssl_dir_manage {
     puppet::config::master {
       'ssldir':           value => $::puppet::server::ssl_dir;
--- a/spec/classes/puppet_server_config_spec.rb
+++ b/spec/classes/puppet_server_config_spec.rb
@@ -14,6 +14,8 @@ describe 'puppet::server::config' do
         rundir              = '/var/run/puppet'
         vardir              = '/var/lib/puppet'
         puppetserver_vardir = '/var/lib/puppet'
+        puppetserver_logdir = '/var/log/puppet'
+        puppetserver_rundir = '/var/run/puppet'
         ssldir              = '/var/lib/puppet/ssl'
         sharedir            = '/usr/share/puppet'
         etcdir              = '/etc/puppet'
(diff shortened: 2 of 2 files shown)


mmoll:
@schurzi could you describe the error you were seeing? Best would be an added test with these values set and compared...

mmoll:
I'm seeing these values in my puppet.conf under the [main] section

schurzi:
@mmoll the settings should be in the [master] section, I will recheck this.

The problem is, that starting with Puppet4, they separated the vardir of the agent and the puppet server. In the default Foreman install the vardir for the puppetserver (JVM) is set in the puppetserver.conf but not in the puppet.conf. we are using "puppet master --compile <node>" which throws errors, because it cannot find the cached facts.

I will look into creating tests for this.

Any advice on testing this with foreman-install?

mmoll:
@schurzi You'd need to regenerate the parser cache with Pupept 4, that's why I'm testing these things at the moment with a modules dir filled by librarian-puppet and the Puppetfile of the foreman-installer repo and a manifest I can tweak.

mmoll:
I know it's kind of an edge case, but e.g. on OSes with a non-AIO Ruby Puppet package, there can still be a rack/passenger setup and I guess in that case the old paths should be used, so wrapping these in something like `if $::puppet::server::implementation == 'puppetserver' ` feels advisable.

schurzi:
I just did a test, the aditional variables are in the correct section:
`[ WARN 2016-12-12T20:11:15 verbose]  /Stage[main]/Puppet::Config/Concat[/etc/puppetlabs/puppet/puppet.conf]/File[/etc/puppetlabs/puppet/puppet.conf]/content:
[ WARN 2016-12-12T20:11:15 verbose] --- /etc/puppetlabs/puppet/puppet.conf      2016-12-12 20:02:54.469677359 +0000
[ WARN 2016-12-12T20:11:15 verbose] +++ /tmp/puppet-file20161212-18075-1ol0yix  2016-12-12 20:11:14.950078064 +0000
[ WARN 2016-12-12T20:11:15 verbose] @@ -37,7 +37,10 @@
[ WARN 2016-12-12T20:11:15 verbose]      ca = true
[ WARN 2016-12-12T20:11:15 verbose]      certname = puppet.localdomain
[ WARN 2016-12-12T20:11:15 verbose]      external_nodes = /etc/puppetlabs/puppet/node.rb
[ WARN 2016-12-12T20:11:15 verbose] +    logdir =
[ WARN 2016-12-12T20:11:15 verbose]      node_terminus = exec
[ WARN 2016-12-12T20:11:15 verbose]      parser = current
[ WARN 2016-12-12T20:11:15 verbose] +    rundir =
[ WARN 2016-12-12T20:11:15 verbose]      ssldir = /etc/puppetlabs/puppet/ssl
[ WARN 2016-12-12T20:11:15 verbose] +    vardir = /var/lib/puppet
`

Plase don't mix this with the values in the [main] section, that were there before!

Regarding the `if`section. Yes, that's a good idea, since this is only relevant with puppetserver. I will also add checks that only set these lines, if there is a value. 

I will modify this. Tomorrow I will lok into tests.

mmoll:
@schurzi by "test" I meant a spec test.

mmoll:
looking good otherwise

schurzi:
yes, sorry for the misunderstanding. I also meant spec tests.
Should I also add spec tests for absence?

spec/classes/puppet_server_config_spec.rb
```diff
@@ -604,6 +610,21 @@
...
+      describe 'with server_implementation => "puppetserver"', :if => (Puppet.version >= '4.0.0') do
+        let :pre_condition do
+          "class {'puppet':
+            server => true,
+            server_implementation => 'puppetserver'
+          }"
+        end
+
+        it 'should configure puppet.conf' do
+          should contain_puppet__config__master("vardir").with({'value' => "#{puppetserver_vardir}"})
```
ekohl: I think you can write ```.with_value(puppetserver_vardir)```

manifests/server/config.pp
```diff
@@ -6,6 +6,21 @@
 
   if $::puppet::server::implementation == 'puppetserver' {
     contain 'puppet::server::puppetserver' # lint:ignore:relative_classname_inclusion (PUP-1597)
+    if !empty($::puppet::server::puppetserver_vardir) {
```
ekohl: I think you can use ```unless empty() {``` which is slightly more readable.

schurzi:
@ekohl thank you for your input! I'm just learning ruby and try to stick to what I see in the code I'm modifying. Your suggestions are welcome!

If everything is ok by your standards, I will squash the commits to keep the log clean.

mmoll:
merged, danke @schurzi!

PR #469, opened 2017-01-05
Acceptance tests with beaker-rspec

vinzent:
I've written simple tests with [beaker-rspec](https://github.com/puppetlabs/beaker-rspec) to play with the module. 

Is it of any interest for you?

Run tests one by one with:
 
 ```
PUPPET_INSTALL_TYPE=agent \
  BEAKER_set=centos-7-x64_docker \
  bundle exec rspec spec/acceptance/puppetserver_upgrade_2_6_0_to_2_7_2_spec.rb

# or the other test with EL6
PUPPET_INSTALL_TYPE=agent \
  BEAKER_set=centos-6-x64_docker \
  bundle exec rspec spec/acceptance/puppetserver_latest_spec.rb
 ```

Further  `BEAKER_*` env params:

* `BEAKER_destroy=no`: don't delete docker process after finishing the test. Can afterwards be inspected with `docker exec -i $ID /bin/bash`
* `BEAKER_provision=no`: don't re-initialize - works for me with vagrant+virtualbox but not with docker
* `BEAKER_debug=yes`: verbose output of whats going on

### DIFF
--- a/.sync.yml
+++ b/.sync.yml
@@ -2,3 +2,13 @@
 Rakefile:
   param_docs_pattern:
     - manifests/init.pp
+Gemfile:
+  extra:
+    - gem: 'beaker-rspec'
+      options:
+        groups:
+          - 'system_tests'
+    - gem: 'beaker-puppet_install_helper'
+      options:
+        groups:
+          - 'system_tests'
--- a/Gemfile
+++ b/Gemfile
@@ -27,5 +27,7 @@ gem 'json', '~> 1.0', {"platforms"=>["ruby_19"], "groups"=>["test"]}
 gem 'json_pure', '~> 1.0', {"platforms"=>["ruby_19"], "groups"=>["test"]}
 gem 'metadata-json-lint'
 gem 'kafo_module_lint'
+gem 'beaker-rspec', {"groups"=>["system_tests"]}
+gem 'beaker-puppet_install_helper', {"groups"=>["system_tests"]}
 
 # vim:ft=ruby
--- a/spec/acceptance/nodesets/centos-6-x64.yml
+++ b/spec/acceptance/nodesets/centos-6-x64.yml
@@ -1,0 +1,14 @@
+---
+HOSTS:
+  centos-6-x64:
+    roles:
+      - master
+    platform: centos-6-x86_64
+    box: centos/6
+    hypervisor: vagrant
+    vagrant_memsize: 3072
+CONFIG:
+  type: aio
+...
+# vim: syntax=yaml
+
--- a/spec/acceptance/nodesets/centos-6-x64_docker.yml
+++ b/spec/acceptance/nodesets/centos-6-x64_docker.yml
@@ -1,0 +1,11 @@
+HOSTS:
+  centos-6-x64-docker:
+    platform: centos-6-x86_64
+    hypervisor : docker
+    image: centos:6
+    # docker_preserve_image: true
+    docker_cmd: '["/sbin/init"]'
+    docker_image_commands:
+      - 'yum install -y crontabs tar wget'
+CONFIG:
+  type: aio
--- a/spec/acceptance/nodesets/centos-7-x64.yml
+++ b/spec/acceptance/nodesets/centos-7-x64.yml
@@ -1,0 +1,14 @@
+---
+HOSTS:
+  centos-7-x64:
+    roles:
+      - master
+    platform: centos-7-x86_64
+    box: centos/7
+    hypervisor: vagrant
+    vagrant_memsize: 3072
+CONFIG:
+  type: aio
+...
+# vim: syntax=yaml
+
--- a/spec/acceptance/nodesets/centos-7-x64_docker.yml
+++ b/spec/acceptance/nodesets/centos-7-x64_docker.yml
@@ -1,0 +1,15 @@
+HOSTS:
+  centos-7-x64-docker:
+    platform: centos-7-x86_64
+    hypervisor : docker
+    image: centos:7
+    # docker_preserve_image: true
+    docker_cmd: '["/usr/sbin/init"]'
+    docker_image_commands:
+      - 'yum install -y -q crontabs tar wget which iproute'
+      # crontab: required for  puppet cron resources
+      # iproute: required for beaker/serverspec port tests which require the 
+      #   ss utility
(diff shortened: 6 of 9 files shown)


domcleal:
I think this is very useful, thanks. The Travis CI build failures need to be fixed, perhaps by grouping the acceptance gems and excluding that in .travis.yml, or pin based on Ruby version. Note that the Gemfile is controlled via modulesync, it should be added also to .sync.yml or theforeman/foreman-installer-modulesync if it's generic.

Ideally the tests should be run on Travis CI too, but that can always be done later.

mmoll:
@vinzent plase change the Ruby version used for the Ruby 2.2 tests to 2.2.6, I'll open a corresponding change to our modulesync repo with an explanation.

ekohl:
I agree that it's useful, but IMHO it's only useful if it's really tested. To me that means that we shouldn't rely on a manual run so Travis is the most logical place.

vinzent:
@ekohl I don't think it's possible to run beaker+vagrant on travis. I think some thing with beaker+docker would work: https://blog.travis-ci.com/2015-08-19-using-docker-on-travis-ci/ , https://www.camptocamp.com/de/actualite/puppet-acceptance-tests-travis-ci-docker/

ekohl:
@vinzent that's a good point. Maybe we could borrow some things from voxpupuli.

vinzent:
puppet-puppetserver has docker nodesets: https://github.com/voxpupuli/puppet-puppetserver/blob/master/spec/acceptance/nodesets/centos-7-x86_64-docker.yml

tried to run this test which ran fine on centos-7 with vagrant/virtualbox. it fails:

```
BEAKER_debug=yes PUPPET_INSTALL_TYPE=agent BEAKER_set=centos-7-x64_docker BEAKER_destroy=no bundle exec rspec spec/acceptance/puppetserver_latest_spec.rb
...
  Notice: Compiled catalog for centos-7-x64.local in environment production in 1.27 seconds
  Info: Applying configuration version '1483740345'
...
  Info: Puppet::Config::Master[parser]: Scheduling refresh of Puppet::Config::Entry[master_parser]
  Info: Puppet::Config::Master[strict_variables]: Scheduling refresh of Puppet::Config::Entry[master_strict_variables]
  Info: Puppet::Config::Master[ssldir]: Scheduling refresh of Puppet::Config::Entry[master_ssldir]
  Notice: /Stage[main]/Puppet::Server::Config/Exec[puppet_server_config-create_ssl_dir]: Triggered 'refresh' from 1 events
  Error: Could not find user puppet
  Error: /Stage[main]/Puppet::Server:
  Error: Could not find user puppet
Error: /Stage[main]/Puppet::Server::Config/File[/etc/puppetlabs/puppet/ssl/private_keys]/owner: change from root to puppet failed: Could not find user puppet
  Notice: /Stage[main]/Puppet::Server::Config/File[/etc/puppetlabs/puppet/ssl/private_keys]/group: group changed 'root' to 'puppet'
  Notice: /Stage[main]/Puppet::Server::Config/File[/etc/puppetlabs/puppet/ssl/private_keys/centos-7-x64.local.pem]: Dependency File[/etc/puppetlabs/puppet/ssl/private_keys] has failures: true

...
```

vinzent:
found the problem. which is not available and the puppetserver rpm uses it in the pre/post scripts. opend a ticket https://tickets.puppetlabs.com/browse/CPR-400 .

now the test for centos-7 on docker ran successfully.

vinzent:
pushed centos docker nodesets. 

Ran a successfull test with:

```
BEAKER_debug=yes PUPPET_INSTALL_TYPE=agent \
  BEAKER_set=centos-6-x64_docker BEAKER_destroy=no \
  bundle exec rspec spec/acceptance/puppetserver_latest_spec.rb
```

I'm new with travis. don't know what best way is to workaround beaker requiring newish ruby versions. The tests that now run don't require beaker - it just fails to install the rubygem. If the tests shall run it looks like it only can on Ruby >= 2.2.5 with beaker 3.7.0

ekohl:
You can put the gems in a separate group and have it excluded in the default runs like the development group. That should at least get the normal tests to pass again.

Then for the case of running the beaker tests, I do wonder if it won't take up too much resources. I ran them locally and they're quite heavy. Maybe we should just get this merged and then look at running them scheduled. https://docs.travis-ci.com/user/cron-jobs/ can be used to run this weekly. No idea how well that would work in practice :)

vinzent:
@ekohl @domcleal @mmoll added some FIXUP commits with lower puppetserver memory footprint, Gemfile system_tests group and .traivs.yaml with excluded system_tests group.

only a bit related: why are you running spec tests against Puppet 3.5? And not 3.8? And 4.0+4.4 instead something like of 4.8? Is there a distro still shipping 3.5?

ekohl:
> only a bit related: why are you running spec tests against Puppet 3.5? And not 3.8? And 4.0+4.4 instead something like of 4.8? Is there a distro still shipping 3.5?

We're using ```~> VERSION``` so 3.5 will also match 3.8 as you can see on https://travis-ci.org/theforeman/puppet-puppet/jobs/190726536#L200. Same for 4.4 which on https://travis-ci.org/theforeman/puppet-puppet/jobs/190726554#L217 means 4.8.1.

ekohl:
We use [modulesync](https://github.com/voxpupuli/modulesync) to manage (among other files) ```Gemfile```. Local gems should also be added to ```.sync.yml```. Can you add the gems there as well?

Other than that I think it's ready to be merged.

vinzent:
@ekohl is there an example what I need to put in .sync.yml?

vinzent:
@ekohl  and what about the .travis.yml `--withtout system_tests` and .gitignore?

ekohl:
For ```.travis.yml``` and ```.gitignore``` I submitted https://github.com/theforeman/foreman-installer-modulesync/pull/49

I believe if you add the following to ```.sync.yml``` then the ```Gemfile``` changes to that are persisted as well.

```
Gemfile:
  extra:
    - gem: 'beaker-rspec'
      options:
        groups:
          - 'system_tests'
    - gem: 'beaker-puppet_install_helper'
      options:
        groups:
          - 'system_tests'
```

vinzent:
updated the .sync.yml and rebased to latest master.

domcleal:
Merged, thanks @vinzent!

PR #475, opened 2017-02-09
Add optional ssl_chain_filepath option for puppet::server

Ngibb:
Motivation for the change is to let the user specificity the value of `ssl-cert-chain` in the `webserver.conf` file for puppetserver.
Previously the module hard coded the value of `ssl-cert-chain` to be `${ssl_dir}/ca/ca_crt.pem`.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -591,6 +591,11 @@
 #                                           Defaults to [ 'TLSv1.2' ]
 #                                           type:Array[String]
 #
+# $server_ssl_chain_filepath::              Path to certificate chain for puppetserver
+#                                           Only used when $ca is true
+#                                           Defaults to "${ssl_dir}/ca/ca_crt.pem"
+#                                           type:Optional[Stdlib::Absolutepath]
+#
 # $server_cipher_suites::                   List of SSL ciphers to use in negotiation
 #                                           Defaults to [ 'TLS_RSA_WITH_AES_256_CBC_SHA256', 'TLS_RSA_WITH_AES_256_CBC_SHA',
 #                                           'TLS_RSA_WITH_AES_128_CBC_SHA256', 'TLS_RSA_WITH_AES_128_CBC_SHA', ]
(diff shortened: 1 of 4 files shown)


manifests/server.pp
```diff
@@ -569,7 +574,7 @@
   if $ca {
     $ssl_ca_cert = "${ssl_dir}/ca/ca_crt.pem"
     $ssl_ca_crl  = "${ssl_dir}/ca/ca_crl.pem"
-    $ssl_chain   = "${ssl_dir}/ca/ca_crt.pem"
+    $ssl_chain   = pick($ssl_chain_filepath, "${ssl_dir}/ca/ca_crt.pem")
```
domcleal: The default should be given in params.pp rather than using pick() here, it'll be easier for users and tools like kafo (for foreman-installer) to see the default value.

Ngibb:
Is there a resource to help me write a test?

manifests/init.pp
```diff
@@ -591,6 +591,10 @@
 #                                           Defaults to [ 'TLSv1.2' ]
 #                                           type:Array[String]
 #
+# $server_ssl_chain_filepath::              Path to certificate chain for puppetserver
```
ekohl: Maybe add a note that it's only used when ```$ca``` is true?

Ngibb:
I gave writing a test a shot. Let me know what you think.

manifests/init.pp
```diff
@@ -591,6 +591,11 @@
...
+#                                           Only used when $ca is true
+#                                           Defaults to "${ssl_dir}/ca/ca_crt.pem"
+#                                           type:Optional[String]
```
mmoll: Shouldn't this be `[Stdlib::Absolutepath]`?

manifests/server.pp
```diff
@@ -348,6 +348,10 @@
 #                              Defaults to [ 'TLSv1.2' ]
 #                              type:array
 #
+# $ssl_chain_filepath::        Path to certificate chain for puppetserver
+#                              Defaults to "${ssl_dir}/ca/ca_crt.pem"
+#                              type:string
```
mmoll: also here

mmoll:
one question, looking good otherwise

domcleal:
Merged, thanks @Ngibb.

PR #476, opened 2017-02-11
Debian 9 removed its native puppet-agent package

mmoll:
see Debian bug #827867 for details


domcleal:
Merged, thanks @mmoll.

PR #477, opened 2017-02-15
Allow disabling crl when server_ca => true

Ngibb:
This allows the CRL checking to be disabled when server_ca is enabled.
Useful for users who have an external CA configured.

### DIFF
--- a/manifests/server.pp
+++ b/manifests/server.pp
@@ -570,7 +570,7 @@ class puppet::server(
     $ssl_ca_cert = "${ssl_dir}/ca/ca_crt.pem"
     $ssl_ca_crl  = "${ssl_dir}/ca/ca_crl.pem"
     $ssl_chain   = "${ssl_dir}/ca/ca_crt.pem"
-    $_crl_enable = true
+    $_crl_enable = pick($crl_enable, true)
   } else {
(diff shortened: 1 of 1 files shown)


mmoll:
@james-powis coculd you have a look here?

james-powis:
I like this change, should not impact anything from a backwards compatibility perspective.

james-powis:
I am a bit confused though on how ca => true and yet the ca is external...

Ngibb:
Our puppetserver is an subordinate CA to our root trust. Puppetserver has a hard time validating CRL's outside of it's vanilla self-signed configuration. Which satisfies `ca => true` and our `root` CA being external.

mmoll:
The idea with a subordinate CA sounds like it solves some  problems in company setups with internal CAs, merged, thanks @Ngibb!

james-powis:
Is this something which should be documented? I know i personally hate finding nuggets of information in Merge requests and closed issues...

mmoll:
@james-powis Indeed, but I guess it would make sense to open up a new MD file for the "Advanced Scenarios" setion, where more complicated setups regarding clusting, load balancing and also using 3rd party CAs are described.

PR #478, opened 2017-02-27
Option disable ssl for reporting

Conzar:
Its unnecessary in some environments to utilise SSL certificates when sending reports from the puppet server to the foreman server (not on the same server).   In cases where there are multiple puppet masters with different CA's, its necessary to disable ssl when sending a report.  

This patch provides an additional option called server_foreman_enable_ssl which is set to true by default which has the same behaviour as before.  When set to false, the following parameters are set to empty strings in foreman.yaml in the puppet configuration directory.

```
:ssl_ca: ""
:ssl_cert: ""
:ssl_key: ""
```

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -483,6 +483,9 @@
 # $server_foreman_ssl_key::                 Key for authenticating against Foreman server
 #                                           type:Optional[Stdlib::Absolutepath]
 #
+# $server_foreman_enable_ssl::              Should puppet use foreman ssl to send reports
+#                                           type:Boolean
+#
 # $server_puppet_basedir::                  Where is the puppet code base located
 #                                           type:Optional[Stdlib::Absolutepath]
 #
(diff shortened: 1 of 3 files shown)


mmoll:
I think in that case it's rather common practise to disable the cert checking on the Foreman side, or am I missing something there?

Conzar:
How do you disable the cert checking on the foreman side?  Also, I still want https for the web UI.

mmoll:
Try adding the respective master in the "Trusted puppetmaster hosts" setting

Conzar:
I don't see that setting in the foreman puppet module.

mmoll:
In that case use the foreman::foreman_config_entry from theforeman-foreman on trusted_puppetmaster_hosts.

Conzar:
I just tried setting the trusted hosts and that seems to work.  So I will close this as its not necessary.  Thank you.

Conzar:
I was too quick to close this PR.  The following is what I am able to observe.
* Reports are able to be sent from puppet server to foreman
* pushing facts are NOT able to be sent from puppet server to foreman

So this is still a problem.

Conzar:
Closing again, as I think pushing facts might be another issue altogether.

PR #480, opened 2017-02-27
change private_keys from file declaration to ensure_resource()

### LINKED ISSUE #479: Unnecessary file declaration in "puppet-puppet/manifests/server/config.pp"
On line 142 of "puppet-puppet/manifests/server/config.pp" starts the declaration of the puppetserver cert private key. The file is defined without content, and I don't see the possibility to hand over content to the module. So as far as I see it, this only gives me a "Duplicate declaration" error if I try to fill the file outside of the module.
Sadly that's currently the showstopper for my role-out of the module, and I would be very grateful for any hint on how to make this part of the module work.

```
  file { "${::puppet::server::ssl_dir}/private_keys/${::puppet::server::certname}.pem":
    owner => $::puppet::server::user,
    group => $::puppet::server::group,
    mode  => '0640',
  }
```

preisbeck:
fixes #479
Prevent a "Double declaration" error when defining the content of private keys outside the puppet module.

### DIFF
--- a/manifests/server.pp
+++ b/manifests/server.pp
@@ -570,7 +570,7 @@ class puppet::server(
     $ssl_ca_cert = "${ssl_dir}/ca/ca_crt.pem"
     $ssl_ca_crl  = "${ssl_dir}/ca/ca_crl.pem"
     $ssl_chain   = "${ssl_dir}/ca/ca_crt.pem"
-    $_crl_enable = true
+    $_crl_enable = pick($crl_enable, true)
   } else {
     $ssl_ca_cert = "${ssl_dir}/certs/ca.pem"
     $ssl_ca_crl  = pick($ca_crl_filepath, "${ssl_dir}/crl.pem")


manifests/server/config.pp
```diff
@@ -139,11 +139,7 @@
     require => Exec['puppet_server_config-create_ssl_dir'],
   }
 
-  file { "${::puppet::server::ssl_dir}/private_keys/${::puppet::server::certname}.pem":
-    owner => $::puppet::server::user,
-    group => $::puppet::server::group,
-    mode  => '0640',
-  }
+  ensure_resource('file', "${::puppet::server::ssl_dir}/private_keys/${::puppet::server::certname}.pem", {ensure => file})
```
mmoll: The user/group/mode is important for a default Foreman setup for the foreman-proxy to be able to re-use the key for its TLS communication. From your issue I guess you are defining the content elsewhere?

preisbeck: Yes, we have a module that handles all our ssl keys and certificates. I will change the pull-request and add a switch to disable the certificate handling, so that the default behavior stays the same and we can toggle it off the let our module do the work.

PR #482, opened 2017-03-05
Lower JVM heap size when low memory is detected

ekohl:
This is untested, but should help especially when testing and you don't really need performance but it just needs to be up.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -411,15 +411,29 @@ class puppet::params {
   }
 
   $server_jvm_java_bin      = '/usr/bin/java'
-  $server_jvm_min_heap_size = '2G'
-  $server_jvm_max_heap_size = '2G'
   $server_jvm_extra_args    = '-XX:MaxPermSize=256m'
+
+  # This is some very trivial "tuning". See the puppet reference:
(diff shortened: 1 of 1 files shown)


manifests/params.pp
```diff
@@ -411,15 +411,24 @@
   }
 
   $server_jvm_java_bin      = '/usr/bin/java'
-  $server_jvm_min_heap_size = '2G'
-  $server_jvm_max_heap_size = '2G'
   $server_jvm_extra_args    = '-XX:MaxPermSize=256m'
 
+  # This is some very trivial "tuning". See the puppet reference:
+  # https://docs.puppet.com/puppetserver/latest/tuning_guide.html
+  if $::memorysize_mb >= 2048 {
```
domcleal: This is a string on pre-4.x versions, consider using `0 + $::memorysize_mb` to force the type change.

ekohl:
I also noticed that VMs with 1G RAM and crashkernel enabled end up with 992MB RAM so fail to pass with 1G. I'm adding another case for that as well.

manifests/params.pp
ekohl: Fixed

mmoll:
merged, bedankt @ekohl!

PR #483, opened 2017-03-08
add sles support

timogoebel:


manifests/params.pp
```diff
@@ -15,6 +13,17 @@
   $runinterval         = '1800'
   $runmode             = 'service'
 
+  case $::osfamily {
+    'Suse': {
+      $user            = 'root'
```
domcleal: If this is to affect only the ownership of the Puppet dir (e.g. `/etc/puppet`) then you should probably change `$dir_owner` circa line 188 instead.

Gemfile
```diff
@@ -29,5 +29,6 @@ gem 'metadata-json-lint'
 gem 'kafo_module_lint'
 gem 'beaker-rspec', {"groups"=>["system_tests"]}
 gem 'beaker-puppet_install_helper', {"groups"=>["system_tests"]}
+gem 'facterdb', :git => 'https://github.com/camptocamp/facterdb', :ref => 'master'
```
domcleal: This file is modulesynced, so ideally should be released first to avoid this modification.

manifests/params.pp
timogoebel: On SLES12 the user `puppet` does not exist at all.

domcleal: Yes, that's the same as the AIO packages then. I don't think $user has any other use, but the other AIO and Windows logic is on $dir_owner rather than $user. It should be kept in one place.

Gemfile
ekohl: Opened https://github.com/camptocamp/facterdb/issues/35

metadata.json
```diff
@@ -99,6 +99,12 @@
       "operatingsystem": "Archlinux"
     },
     {
+      "operatingsystem": "SLES",
+      "operatingsystemrelease": [
+        "12"
```
mmoll: also works on 11 for me :)

timogoebel:
Deactivated the puppetmaster tests for SLES with a tiny helper as they don't work and this PR should just add agent support. Changed the code to use `dir_*` parameters.

mmoll:
I'll probably work further on SLES support (maybe even OpenSUSE) later, but LGTM now.

mmoll:
merged, danke @timogoebel!

PR #484, opened 2017-03-09
use server_idle_timeout on puppetserver

disconn3ct:
Otherwise server_idle_timeout is ignored.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -482,6 +482,11 @@
 #
 # $server_allow_header_cert_info::          Enable client authentication over HTTP Headers
 #                                           Defaults to false, is also activated by the $server_http setting
+#
+# $web_idle_timeout::                       Time in ms that Jetty allows a socket to be idle, after processing has
+#                                           completed.
+#                                           Defaults to 30000, using the Jetty default of 30s
+#
 # === Usage:
 #
 # * Simple usage:
@@ -664,6 +669,7 @@ class puppet (
   Boolean $server_check_for_updates = $puppet::params::server_check_for_updates,
   Boolean $server_environment_class_cache_enabled = $puppet::params::server_environment_class_cache_enabled,
   Boolean $server_allow_header_cert_info = $puppet::params::server_allow_header_cert_info,
+  Integer[0] $web_idle_timeout = $puppet::params::web_idle_timeout,
 ) inherits puppet::params {
   include ::puppet::config
   Class['puppet::config'] -> Class['puppet']
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -438,6 +438,7 @@ class puppet::params {
   $server_default_manifest_content        = '' # lint:ignore:empty_string_assignment
   $server_max_requests_per_instance       = 0
   $server_idle_timeout                    = 1200000
+  $web_idle_timeout                       = 300000
   $server_connect_timeout                 = 120000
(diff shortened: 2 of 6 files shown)


ekohl:
It's used in templates/server/puppetserver/conf.d/puppetserver.conf.erb but that's for outbound connections. I'm unsure if we should overload the parameter for both settings. Personally I have little experience with puppetserver so I'd like someone with more knowledge to have a look.

mmoll:
I was also confused by this and the PL docs on this, but as these are two different things, there should also be an extra parameter for this. Also, this would need added tests.

mmoll:
@disconn3ct what's the status here?

disconn3ct:
Haven't had time to look at it. If anyone wants to pick it up great, otherwise I'm hoping sometime to do it. Doesn't look like a lot of work, just a matter of time to do it in.

disconn3ct:
@mmoll better?

mmoll:
@disconn3ct I pushed some minor changes, could you check if that's OK?

manifests/init.pp
```diff
@@ -482,6 +482,11 @@
 #
 # $server_allow_header_cert_info::          Enable client authentication over HTTP Headers
 #                                           Defaults to false, is also activated by the $server_http setting
+#
+# $server_web_idle_timeout::                Time in ms that Jetty allows a socket to be idle, after processing has
+#                                           completed.
+#                                           Defaults to 30000, using the Jetty default of 30s
```
domcleal: The default according to https://docs.puppet.com/puppetserver/latest/config_file_puppetserver.html is 20 minutes (1,200,000 milliseconds), should this be the same?

manifests/params.pp
```diff
@@ -438,6 +438,7 @@
   $server_default_manifest_content        = '' # lint:ignore:empty_string_assignment
   $server_max_requests_per_instance       = 0
   $server_idle_timeout                    = 1200000
+  $server_web_idle_timeout                = 300000
```
domcleal: The comment states 30,000, but this is 300,000 (5 mins, not 30 seconds).

manifests/init.pp
disconn3ct: wrong config file, this is in webserver.conf https://docs.puppet.com/puppetserver/2.7/config_file_webserver.html -> https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/blob/master/doc/jetty-config.md

manifests/params.pp
disconn3ct: updated

manifests/init.pp
domcleal: Ah thanks, so that's the _server_ idle time and not the HTTP client idle time.

domcleal:
Where did the "server_" prefix go? It shows up in my last comments as "outdated", but now appears to be missing. The prefix should be kept, especially if it's the server's socket rather than HTTP clients.

disconn3ct:
Temporarily closing to fix up mmoll's changes (I overwrote them, too early in the morning.)

disconn3ct:
@mmoll apologies for undoing your changes, please make sure I got them fixed (and I'll stop git'ing while tired..)

And it won't reopen because github (https://github.com/isaacs/github/issues/361) so I'll open a fresh one and reference this one. Sorry about that..

PR #485, opened 2017-03-10
replace validate_* calls with datatypes

bastelfreak:


manifests/server.pp
```diff
@@ -391,183 +391,123 @@
...
-
-  validate_re($implementation, '^(master|puppetserver)$')
-  validate_re($parser, '^(current|future)$')
-
   if $environment_timeout {
     validate_re($environment_timeout, '^(unlimited|0|[0-9]+[smh]{1})$')
   }
 
   if $implementation == 'puppetserver' {
     validate_re($jvm_min_heap_size, '^[0-9]+[kKmMgG]$')
```
ekohl: Given there are sane defaults, should this be enforced in the parameters as well?

bastelfreak: Should be possible. I need to take a look at the datatypes documentation. I'm not sure how I can implement this.

ekohl: https://docs.puppet.com/puppet/latest/lang_data_regexp.html#the-regexp-data-type tells me there should be a ```Regexp``` type for this.

bastelfreak: I think this is correct:
https://docs.puppet.com/puppet/4.6/lang_data_abstract.html#pattern

domcleal:
Replaced by #498, converting all parameters. Thanks @bastelfreak.

PR #486, opened 2017-03-10
Update acceptance tests

ekohl:


domcleal:
Can you resolve the merge conflict?

ekohl:
Updated. I left out Debian tests for now because they fail on the upgrade because those versions are not available to install.

mmoll:
versions of puppetserver?

ekohl:
> versions of puppetserver?

The upgrade integration test installs 2.6.0 and then upgrades to 2.7.2. Those versions weren't installable in the container I tried with. I haven't looked at the logs, but my suspicion is that those versions aren't in the Debian repo (anymore).

domcleal:
I think they are, but the full version suffix may be required (`2.7.0-1puppetlabs1`) for the package provider and apt to install them. At worst, the upgrade test could be skipped on Debian-based OSes but the regular installation could be enabled?

ekohl:
You are correct that it needs to be the full version. Updated the test.

mmoll:
merged, bedankt @ekohl!

PR #487, opened 2017-03-10
add support for Parallels PSBM

bastelfreak:
PSBM is a propritary CentOS 6 from Parallels. It reports osreleasemajor
as 2, not 6.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -357,14 +357,17 @@ class puppet::params {
       }
     }
     'Redhat' : {
+      # PSBM is a CentOS 6 based distribution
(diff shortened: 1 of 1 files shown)


ekohl:
I think that deserves at least some comment

bastelfreak:
what kind of comment do you mean? In the code?

ekohl:
Not sure tbh, but if I saw this without the context of PSBM I'd assume it was a mistake and submit a PR to remove it.

bastelfreak:
@ekohl I added some inline comments as well.

domcleal:
Merged, thanks @bastelfreak.

PR #488, opened 2017-03-14
Fix missing acceptance test module dependencies

domcleal:
May replace #486, now that master's been modulesynced.


mmoll:
merged, thanks @domcleal!

PR #490, opened 2017-03-24
Fixes #19019 - Generate Puppet cert with --allow-dns-alt-names

sean797:
With `--allow-dns-alt-names` Puppet will use the `dns_alt_names` config option in `puppet.conf` when generating the certificate. 

To test run the installer with `--puppet-dns-alt-names my-other-hostname.example.com,another-name.example.com`, you should then be able to configure an agent to use `my-other-hostname.example.com` & `another-name.example.com` as well as the servers hostname.


sean797:
Thanks @ekohl & @domcleal updated, and tests passed 👍

domcleal:
Merged, thanks @sean797.

PR #491, opened 2017-03-24
dedicated setting for allow-header-cert-info

lobeck:
add a dedicated setting to enable allow-header-cert-info without http being active

- i just want to enable allow-header-cert-info, but not the http server, so there should be a setting with only activates the header support.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -643,7 +643,7 @@
 #
 # $server_allow_header_cert_info::          Enable client authentication over HTTP Headers
 #                                           Defaults to false, and can also be activated with the $server_http setting
-#                                           type:boolean
+#                                           type:Boolean
 # === Usage:
 #
 # * Simple usage:
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -641,6 +641,9 @@
 #                                           Defaults to false
 #                                           type:Boolean
 #
+# $server_allow_header_cert_info::          Enable client authentication over HTTP Headers
+#                                           Defaults to false, and can also be activated with the $server_http setting
+#                                           type:boolean
 # === Usage:
 #
 # * Simple usage:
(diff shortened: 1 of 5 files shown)


manifests/init.pp
```diff
@@ -641,6 +641,9 @@
 #                                           Defaults to false
 #                                           type:Boolean
 #
+# $server_allow_header_cert_info::          Enable client authentication over HTTP Headers
+#                                           Defaults to false, and can also be activated with the $server_http setting
+#                                           type:boolean
```
ekohl: Please use ```Boolean``` so it's a Puppet 4 type.

lobeck:
As a clarification: We have a reverse proxy in front of puppetserver and don't want to allow http connections, as https works just fine. As there was no option to toggle allow-header-cert-info without enabling http too, we've added it.

manifests/init.pp
```diff
@@ -641,6 +641,9 @@
 #                                           Defaults to false
 #                                           type:Boolean
 #
+# $server_allow_header_cert_info::          Enable client authentication over HTTP Headers
+#                                           Defaults to false, and can also be activated with the $server_http setting
```
mmoll: the correct wording here is IMHO "..and is also activated by the...", but we can fix that on merge

manifests/server.pp
```diff
@@ -388,6 +388,9 @@
 #                              Defaults to false (the puppetserver will use its own conf.d/auth.conf)
 #                              type:boolean
 #
+# $allow_header_cert_info::    Allow client authentication over HTTP Headers
+#                              Defaults to false, and can also be activated with the $http setting
```
mmoll: here too

mmoll:
IMHO this is fine

mmoll:
merged, danke @lobeck!

PR #495, opened 2017-03-28
Change current version to 7.1.0

domcleal:
There no incompatible changes requiring a major bump.

[ci skip]


mmoll:
hrm... true.

ekohl:
Though soon I would like to go to 8.0.0 where we drop puppet 3, it would be good to release everything prior to that.

PR #497, opened 2017-03-31
Add Puppet 3 client compatibility under rack

sathieu:
Add a Rack middleware re-implementing the v1 and v2 API calls, as
per[1]. This should restore compatibility with Puppet 3 clients when
run under any rack-enabled server (including Passenger).

Closes: #832536

[1] https://github.com/puppetlabs/puppetserver/blob/puppet-server-2.6.0/src/clj/puppetlabs/services/legacy_routes/legacy_routes_core.clj

(Ported from https://anonscm.debian.org/cgit/pkg-puppet/puppet.git/commit/?id=1fb4e3e95b36b2c060320914452b48c29ffa4bde)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
     * Add support for Parallels PSBM
 * Other changes and fixes:
     * Lower JVM heap size when low memory is detected
+    * Add Puppet 3 client compatibility under rack
 
 ## 7.0.2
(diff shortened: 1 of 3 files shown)


mmoll:
@sathieu please add an entry in the changelog ("other changes and fixes"), thx

mmoll:
merged with minor rewording, merci @sathieu!

PR #498, opened 2017-04-06
Move existing Puppet 4 data types to class parameters

domcleal:
1. Moves existing Puppet 4 data types from the class/parameter docs into the class parameters themselves.
1. Converts some older Kafo-only data types into the Puppet 4 style and moves them into the class parameters.
1. Removes the validate_* family of functions, updates a few instances where they had better validation than the data types.

Replaces #485 by adding data types to every parameter.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -8,57 +8,42 @@
 #                                           install. The version should be the exact
 #                                           match for your distro.
 #                                           You can also use certain values like 'latest'.
-#                                           type:String
 #
 # $user::                                   Override the name of the puppet user.
-#                                           type:String
 #
 # $group::                                  Override the name of the puppet group.
-#                                           type:String
 #
 # $dir::                                    Override the puppet directory.
-#                                           type:Stdlib::Absolutepath
 #
 # $codedir::                                Override the puppet code directory.
-#                                           type:Stdlib::Absolutepath
 #
 # $vardir::                                 Override the puppet var directory.
-#                                           type:Stdlib::Absolutepath
 #
 # $logdir::                                 Override the log directory.
-#                                           type:Stdlib::Absolutepath
 #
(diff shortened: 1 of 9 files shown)


mmoll:
merged, thanks @domcleal!

manifests/init.pp
```diff
@@ -664,204 +502,169 @@
...
+  Hash[String, Data] $agent_additional_settings = $puppet::params::agent_additional_settings,
+  Optional[String] $agent_restart_command = $puppet::params::agent_restart_command,
+  String $classfile = $puppet::params::classfile,
+  String $hiera_config = $puppet::params::hiera_config,
+  String $main_template = $puppet::params::main_template,
+  String $agent_template = $puppet::params::agent_template,
+  String $auth_template = $puppet::params::auth_template,
+  Boolean $allow_any_crl_auth = $puppet::params::allow_any_crl_auth,
+  Array[String] $auth_allowed = $puppet::params::auth_allowed,
+  Array[String] $client_package = $puppet::params::client_package,
```
alexjfisher: This was a breaking change for me.  `client_package` could be a string *or* an array previously.

mmoll: could you open a PR for the CHANGELOG?

alexjfisher: I could, or a PR to change the type to
```
Variant[String, Array[String]]
```
??
It's passed to the title of a Package resource, so a plain string definitely works and is perhaps more what you might expect from a parameter called 'client_package' and not, for example, 'client_packages'? Either way, let me know and I'll create the PR.

mmoll: @ekohl any opinion? ☝️

ekohl: Given the variable name is singular I think the ```Variant``` is preferable.

PR #499, opened 2017-04-06
Add autosign_source, which may be used instead of autosign_content

downsj2:
autosign_content is nice, but very ugly if you only use hiera to pass parameters to classes.

This adds autosign_source, which can be used instead of autosign_content and does the obvious.

### DIFF
--- a/spec/classes/puppet_server_config_spec.rb
+++ b/spec/classes/puppet_server_config_spec.rb
@@ -266,11 +266,11 @@ describe 'puppet::server::config' do
       end
 
 
-      describe "when autosign_source => set to foo.bar and and autosign_entries set to ['foo.bar']=> true" do
+      describe "when autosign_source => set to puppet:///foo/bar and and autosign_entries set to ['foo.bar']=> true" do
         let :pre_condition do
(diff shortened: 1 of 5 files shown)


downsj2:
I don't believe it's possible to write an rspec test which tests against content specified via the "source" attribute of a file resource, only against contents specified via "content".

spec/classes/puppet_server_config_spec.rb
```diff
@@ -262,11 +262,24 @@
...
+              server           => true,
+              autosign_source  => 'puppet:///foo/bar',
+              autosign_entries => ['foo.bar'],
+           }"
+        end
+
+        it { should raise_error(Puppet::Error, /Cannot set both autosign_content\/autosign_source and autosign_entries/) }
+      end
+
+
```
mmoll: superfluous newline

mmoll:
@downsj2 please rebase onto current master and squash the commits togetehr into one. Also, this needs a datatype instead of the validation.

downsj2:
Replaced with PR #501, closing this one.

PR #501, opened 2017-04-10
Add autosign_source, which may be used instead of autosign_content

downsj2:
This replaces PR #499, it has been rebased to master and squashed.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -89,6 +89,9 @@
 #                                           file('another_module/autosign.sh') or
 #                                           template('another_module/autosign.sh.erb')
 #
+# $autosign_source::                        If set, use this as the source for the autosign file,
+#                                           instead of autosign_content.
+#
 # $usecacheonfailure::                      Switch to enable use of cached catalog on
 #                                           failure of run.
 #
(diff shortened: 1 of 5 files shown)


manifests/init.pp
```diff
@@ -89,6 +89,10 @@
 #                                           file('another_module/autosign.sh') or
 #                                           template('another_module/autosign.sh.erb')
 #
+# $autosign_source::                        If set, use this as the source for the autosign file,
+#                                           instead of autosign_content.
+#                                           type:Optional[String]
```
ekohl: Now that we use native Puppet 4 types, this line is no longer needed.

manifests/server.pp
```diff
@@ -24,6 +24,10 @@
 #                              file('another_module/autosign.sh') or
 #                              template('another_module/autosign.sh.erb')
 #
+# $autosign_source::           If set, use this as the source for the autosign file,
+#                              instead of autosign_content.
+#                              type:Optional[String]
```
ekohl: Same here

downsj2: Removed, branch updated.

manifests/init.pp
```diff
@@ -89,6 +89,9 @@
 #                                           file('another_module/autosign.sh') or
 #                                           template('another_module/autosign.sh.erb')
 #
+# $autosign_source::                        If set, use this as the source for the autosign file,
+#                                           instead of autosign_content.
```
downsj2: If you specify both content and source, puppet will just reject it.

For example, `puppet apply -e 'file { "/tmp/foo": content => "foo bar\n", source => "/tmp/motd" }'` produces:

`Error: Validation of File[/tmp/foo] failed: You cannot specify more than one of content, source, target at line 1`

While there is certainly value in having a slightly more specific error message, this is something meant for people running the module outside of the Foreman Installer.  I'm not sure it's worth this much hand-holding.  People running the module stand alone should understand basic puppet resources.

This suggestion just seems like over kill to me.

mmoll: > If you specify both content and source, puppet will just reject it.

👍 then

mmoll:
merged, thanks @downsj2!

PR #502, opened 2017-04-12
Add web_server_idle_timeout to puppetserver jetty configuration

disconn3ct:
This allows configuration of idle timeouts on the Jetty side. A reopening of #484

### DIFF
--- a/manifests/server.pp
+++ b/manifests/server.pp
@@ -265,6 +265,9 @@
 #
 # $connect_timeout::           How long the server will wait for a response to a connection attempt
 #
+# $web_idle_timeout::          Time in ms that Jetty allows a socket to be idle, after processing has completed.
+#                              Defaults to the Jetty default of 30s
+#
 # $ssl_protocols::             Array of SSL protocols to use.
 #                              Defaults to [ 'TLSv1.2' ]
 #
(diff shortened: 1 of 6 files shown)


manifests/server.pp
```diff
@@ -262,6 +262,9 @@
 #
 # $connect_timeout::           How long the server will wait for a response to a connection attempt
 #
+# $server_web_idle_timeout::   Time in ms that Jetty allows a socket to be idle, after processing has completed.
```
ekohl: This parameter isn't added to the class. Looks like this filename is not in the docs checks that would verify this.

disconn3ct: Added it in, good catch.

manifests/server.pp
```diff
@@ -334,6 +337,7 @@
   Array[String] $cipher_suites = $::puppet::server_cipher_suites,
   Optional[String] $config_version = $::puppet::server_config_version,
   Integer[0] $connect_timeout = $::puppet::server_connect_timeout,
+  Integer[0] $server_web_idle_timeout = $puppet::params::server_web_idle_timeout,
```
ekohl: In general we strip the ```server_``` prefix in this class. I think the same can be done here and just by named ```web_idle_timeout```. Note that in init.pp it should stay prefixed.

mmoll: in the template, it would also be just `@web_idle_timeout` then, as the var is locally available

disconn3ct:
@ekohl fixed I hope. Please take another look.

Thanks for everyone's patience!

mmoll:
I added fixes to the variable passing, so tests are green again.

mmoll:
merged, thanks @disconn3ct!

PR #503, opened 2017-04-12
Stop accepting Foreman Puppetmaster v1 APIs

ekohl:
theforeman-foreman dropped the v1 APIs in 2.2.0. We should not accept
these values anymore.


domcleal:
Merged, thanks @ekohl .

Issue #505, opened 2017-04-19
server config parameter should be moved to [main]


baurmatt:
If the server configuration parameters isn't set in [main], archive_file_server/ca_server/report_server will have strange defaults.

server in [agent]
```bash
root@mbaur-test:/etc/puppetlabs/puppet# cat puppet.conf
[main]
  environmentpath = /etc/puppetlabs/code/environments
[agent]
  environment = production
  server = puppet.example.com
root@mbaur-test:/etc/puppetlabs/puppet# puppet config print archive_file_server
puppet
root@mbaur-test:/etc/puppetlabs/puppet# puppet config print ca_server
puppet
root@mbaur-test:/etc/puppetlabs/puppet# puppet config print report_server
puppet
root@mbaur-test:/etc/puppetlabs/puppet#
```

server in [main]
```bash
root@mbaur-test:/etc/puppetlabs/puppet# cat puppet.conf
[main]
  environmentpath = /etc/puppetlabs/code/environments
  server = puppet.example.com
[agent]
  environment = production
root@mbaur-test:/etc/puppetlabs/puppet# puppet config print archive_file_server
puppet.example.com
root@mbaur-test:/etc/puppetlabs/puppet# puppet config print ca_server
puppet.example.com
root@mbaur-test:/etc/puppetlabs/puppet# puppet config print report_server
puppet.example.com
root@mbaur-test:/etc/puppetlabs/puppet#
```

I'm going to add a PR myself.

PR #506, opened 2017-04-19
Move server config parameter to [main]

baurmatt:
This is needed as some other config parameters (archive_file_server/ca_server/report_server) use the server config parameter as their default.

Fixes GH-505

### DIFF
--- a/manifests/agent/config.pp
+++ b/manifests/agent/config.pp
@@ -16,11 +16,6 @@ class puppet::agent::config inherits puppet::config {
     'noop':              value => $::puppet::agent_noop;
     'usecacheonfailure': value => $::puppet::usecacheonfailure;
   }
-  if !$::puppet::use_srv_records {
-    puppet::config::main {
-      'server':            value => pick($::puppet::config::puppetmaster, $::fqdn);
-    }
-  }
   if $::puppet::configtimeout != undef {
     puppet::config::agent {
       'configtimeout':     value => $::puppet::configtimeout;
--- a/manifests/agent/config.pp
+++ b/manifests/agent/config.pp
@@ -17,7 +17,7 @@ class puppet::agent::config inherits puppet::config {
     'usecacheonfailure': value => $::puppet::usecacheonfailure;
   }
   if !$::puppet::use_srv_records {
-    puppet::config::agent {
+    puppet::config::main {
       'server':            value => pick($::puppet::config::puppetmaster, $::fqdn);
     }
   }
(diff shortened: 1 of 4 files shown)


manifests/agent/config.pp
```diff
@@ -17,7 +17,7 @@
     'usecacheonfailure': value => $::puppet::usecacheonfailure;
   }
   if !$::puppet::use_srv_records {
-    puppet::config::agent {
+    puppet::config::main {
```
domcleal: Should this be moved into `puppet::config` now, if it's no longer an agent-only setting? Other common settings are there, e.g. SRV server settings. This would allow it to be set when $agent = false, but the server is enabled.

baurmatt: Sounds good, going to fix it.

baurmatt: Fixed with 26c61fdd4045905219bd396b56a099b9d1a082c0

spec/classes/puppet_agent_spec.rb
```diff
@@ -52,7 +52,7 @@
         end
 
         it do
-          should contain_puppet__config__agent('server').with_value('foo.example.com')
+          should contain_puppet__config__main('server').with_value('foo.example.com')
```
domcleal: These tests should be moved to puppet_config_spec.rb.

manifests/config.pp
```diff
@@ -48,6 +48,10 @@
       'use_srv_records': value => true;
       'srv_domain': value => $srv_domain;
     }
+  } else {
+    puppet::config::main {
+      'server':            value => pick($puppetmaster, $::fqdn);
```
domcleal: Nitpick, remove the excess whitespace between 'server' and 'value'.

spec/classes/puppet_agent_spec.rb
baurmatt: Sure? Shouldn't this be tested for the agent as well? Perhaps just a copy in puppet_config_spec.rb?

domcleal: Well, the agent config inherits the main config class so there shouldn't be any need to duplicate the test. If the code being tested is in the puppet::config class then the test should be in the corresponding spec file for puppet::config.

baurmatt: Alright. I just moved them to puppet_config_spec.rb. I also did a little bit of restructuring, hope thats ok.

manifests/config.pp
baurmatt: Fixed :)

mmoll:
merged, danke @baurmatt!

PR #507, opened 2017-04-27
Order stopping of standalone master service before starting Apache

domcleal:
When configuring a Passenger-based Puppet master on Debian, installation
of the master package will usually start the service and listen on the
master port. Starting Apache then fails as the port's in use.

This change orders the standalone master resource (with `ensure =>
stopped`) before the Apache service.


domcleal:
Test failures are due to https://tickets.puppetlabs.com/browse/MODULES-4790 in stdlib.

mmoll:
fixes -installer on Debian/stretch for me, thanks @domcleal!

PR #509, opened 2017-05-12
Correctly configure the metrics endpoint in web-routes.conf for Puppet server >= 5.0.0

oranenj:
I thought I'd test the Puppet 5 nightly and it turns out that this module breaks the configuration because there is a new endpoint.

This PR fixes the issue, at least enough to allow puppetserver to start correctly. It does require configuring the puppet::server_puppetserver_version parameter to be '5.0.0' explicitly, but that's probably acceptable at this point.

### DIFF
--- a/spec/classes/puppet_server_puppetserver_spec.rb
+++ b/spec/classes/puppet_server_puppetserver_spec.rb
@@ -487,6 +487,21 @@ describe 'puppet::server::puppetserver' do
         end
       end
 
+      describe 'metrics API endpoint' do
+        context 'when server_puppetserver_version >= 5.0.0' do
+          let(:params) do
+            default_params.merge({
+                                     :server_puppetserver_version => '5.0.0',
+                                     :server_puppetserver_dir => '/etc/custom/puppetserver',
+                                 })
+          end
+          it {
+            should contain_file('/etc/custom/puppetserver/conf.d/web-routes.conf').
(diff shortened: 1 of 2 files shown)


spec/classes/puppet_server_puppetserver_spec.rb
```diff
@@ -487,6 +487,21 @@
...
+        context 'when server_puppetserver_version >= 5.0.0' do
+          let(:params) do
+            default_params.merge({
+                                     :server_puppetserver_version => '5.0.0',
+                                     :server_puppetserver_dir => '/etc/custom/puppetserver',
+                                 })
+          end
+          it {
+            should contain_file('/etc/custom/puppetserver/conf.d/web-routes.conf').
+              with_content(/^\s+"puppetlabs.trapperkeeper.services.metrics.metrics-service\/metrics-webservice": "\/metrics"/)
```
ekohl: You can use ```with_content(%r{})``` and use slashes inside the ```{}```, it's easier to read and maintain.

oranenj: I copy-pasted from the other tests around this one. This module doesn't seem to have rubocop configured, so I didn't want to break the conventions of the surrounding code.

ekohl: We've never discussed rubocop and there are still a lot of style things we could improve on.

mmoll:
merged, kiitos @oranenj!
Maybe that's already enough to the the current systest with the nightly PL repos going again...

PR #510, opened 2017-05-15
Do not manage Puppet Server web-routes.conf

domcleal:
Maintaining the web routes config file has led to a number of bugs
between versions of Puppet Server and will in the future with the
changes between 2.7.x and 5.0.x.

As there are no user-configurable options in the file and the
documentation states that it should not be modified, it is safer to use
the file shipped with the server package. This will ensure the
configuration matches the codebase, preventing startup errors from
config mismatches.

Docs: https://docs.puppet.com/puppetserver/latest/config_file_web-routes.html


mmoll:
looks like a historical leftover from the original puppetserver support PR

ekohl:
I am wondering what will happen to users upgrading. If the file is changed then will rpm/dpkg update it or will things break for some users on yum/apt upgrades?

domcleal:
RPM upgrades will work if the file's unchanged (it's marked `%config(noreplace)`), and I think apt users would be prompted whether to replace the file or not. I think that may be why the upgrade via the module is failing in the Debian acceptance test, I need to take a closer look.

domcleal:
> and I think apt users would be prompted whether to replace the file or not

In my test this didn't happen, probably because the existing web-routes.conf template matches the packaged file exactly so there's no modification, so ought to be fine and matches the RPM behaviour.

> I think that may be why the upgrade via the module is failing in the Debian acceptance test, I need to take a closer look.

The acceptance test was removing the package and then deleting all of /etc/puppetlabs/puppetserver during cleanup between tests, then dpkg wasn't restoring the deleted config files. I changed the test to purge the package so the deleted configs would be created when the package was next installed. PR updated.

mmoll:
merged, thanks @domcleal!

PR #512, opened 2017-06-26
convert $::memorysize_mb when using facter 2.x

mmoll:
When using Puppet 4 with Facter 2 (mostly non-AIO built packages) the following Warning can be seen in the puppetserver log:
```
The string '994.95' was automatically coerced to the numerical value 994.95 at /etc/puppetlabs/code/environments/production/modules/puppet/manifests/params.pp:419:10
The string '994.95' was automatically coerced to the numerical value 994.95 at /etc/puppetlabs/code/environments/production/modules/puppet/manifests/params.pp:423:15
```

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -417,11 +417,16 @@ class puppet::params {
 
   # This is some very trivial "tuning". See the puppet reference:
   # https://docs.puppet.com/puppetserver/latest/tuning_guide.html
-  if 0 + $::memorysize_mb >= 2048 {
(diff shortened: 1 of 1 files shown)


ekohl:
Given we're on puppet 4, can we rely on structured facts? I believe those do have real data types.

mmoll:
ia_s is from stdlib and Facter 2 does not have structured facts :)

ekohl:
Thanks!

PR #514, opened 2017-06-29
fix git integration with theforeman-git 3.0.0

mmoll:
fails on Puppet 5 otherwise


ekohl:
Can we also increase the allowed version in ```metadata.json```?

mmoll:
@ekohl (i'm on mobile, so i can't check) i think it's not in metadata.json as it's an optional dependency.

ekohl:
You're right.

PR #515, opened 2017-06-30
Update Puppetserver configurations for 5.0.x

mmoll:
this is for discussion, no tests, yet.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -463,9 +463,6 @@
 #                                           can query the puppet-admin-api endpoint
 #                                           Defaults to [ '127.0.0.1', '::1', $::ipaddress ]
 #
-# $server_enable_ruby_profiler::            Should the puppetserver ruby profiler be enabled?
-#                                           Defaults to false
-#
 # $server_ca_auth_required::                Whether client certificates are needed to access the puppet-admin api
 #                                           Defaults to true
 #
@@ -485,6 +482,13 @@
 # $server_web_idle_timeout::                Time in ms that Jetty allows a socket to be idle, after processing has
 #                                           completed.
 #                                           Defaults to 30000, using the Jetty default of 30s
+#
+# $server_puppetserver_jruby9k::            For Puppetserver 5, use JRuby 9k? Defaults to false
+#
+# $server_puppetserver_metrics::            Enable metrics (Puppetserver 5.x only) and JRuby profiling?
+#                                           Defaults to true on Puppetserver 5.x and to false on Puppetserver 2.x
+#
+# $server_puppetserver_experimental::       For Puppetserver 5, enable the /puppet/experimental route? Defaults to true
 #
 # === Usage:
 #
@@ -611,7 +615,6 @@ class puppet (
   Boolean $server_default_manifest = $puppet::params::server_default_manifest,
   Stdlib::Absolutepath $server_default_manifest_path = $puppet::params::server_default_manifest_path,
   String $server_default_manifest_content = $puppet::params::server_default_manifest_content,
-  Boolean $server_enable_ruby_profiler = $puppet::params::server_enable_ruby_profiler,
   Array[String] $server_environments = $puppet::params::server_environments,
   String $server_environments_owner = $puppet::params::server_environments_owner,
   Optional[String] $server_environments_group = $puppet::params::server_environments_group,
@@ -670,6 +673,9 @@ class puppet (
   Boolean $server_environment_class_cache_enabled = $puppet::params::server_environment_class_cache_enabled,
   Boolean $server_allow_header_cert_info = $puppet::params::server_allow_header_cert_info,
   Integer[0] $server_web_idle_timeout = $puppet::params::server_web_idle_timeout,
+  Boolean $server_puppetserver_jruby9k = $puppet::params::server_puppetserver_jruby9k,
+  Boolean $server_puppetserver_metrics = $puppet::params::server_puppetserver_metrics,
+  Boolean $server_puppetserver_experimental = $puppet::params::server_puppetserver_experimental,
 ) inherits puppet::params {
   include ::puppet::config
   Class['puppet::config'] -> Class['puppet']
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -413,7 +413,12 @@ class puppet::params {
   }
 
   $server_jvm_java_bin      = '/usr/bin/java'
-  $server_jvm_extra_args    = '-XX:MaxPermSize=256m'
+
+  if versioncmp($::puppetversion, '5.0.0') < 0 {
+    $server_jvm_extra_args = '-XX:MaxPermSize=256m'
+  } else {
+    $server_jvm_extra_args = '-Djruby.logger.class=com.puppetlabs.jruby_utils.jruby.Slf4jLogger'
+  }
 
   # This is some very trivial "tuning". See the puppet reference:
   # https://docs.puppet.com/puppetserver/latest/tuning_guide.html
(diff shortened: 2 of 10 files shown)


manifests/params.pp
```diff
@@ -441,7 +441,7 @@
   $server_idle_timeout                    = 1200000
   $server_web_idle_timeout                = 30000
   $server_connect_timeout                 = 120000
-  $server_enable_ruby_profiler            = false
+  $server_enable_ruby_profiler            = true
```
mmoll: the default changed for the metrics (which are also on by default) - maybe that paramter should go and this get enabled and disabled with metrics?

GregSutcliffe: To clarify, you're suggesting to drop `$server_enable_ruby_profiler` and use `$server_puppetserver_metrics` for both?

templates/auth.conf.erb
```diff
@@ -84,11 +84,13 @@
 path /puppet/v3/environments
 method find
 allow *
+<% if @puppetversion.to_f < 5.0 -%>
```
GregSutcliffe: Does Puppet 5 enable the v3 resource API by default then?

manifests/params.pp
mmoll: yes, it's only an idea... otherwise the default value could also depend on the puppetversion...

templates/auth.conf.erb
mmoll: No, it's gone: https://docs.puppet.com/puppetserver/latest/release_notes.html#deprecation-removed-resourcetypes-api-endpoints-and-resourcetype-cli-face

However, this auth.conf file here is not used for puppetserver anyway now: https://docs.puppet.com/puppetserver/latest/release_notes.html#deprecation-legacy-authconf-settings-are-no-longer-enabled-by-default

manifests/params.pp
GregSutcliffe: Excuse the ignorance, it's been a while :P - is `$server_enable_ruby_profiler` used for anything else at this point? If it's just for debugging and such, then it's probably OK to drop it, I think.

templates/auth.conf.erb
GregSutcliffe: Hmm , does that need addressing in the proxy then? https://github.com/theforeman/smart-proxy/blob/develop/modules/puppet_proxy_puppet_api/v3_api_request.rb#L14

manifests/params.pp
mmoll: I can only cite the docs: `Puppet Server enables profiling for the Puppet Ruby code.` For < 5.0 this was disabled by default, but for 5.x it's enabled by default because the metrics that are coming newly with 5.x seem to depend on it, at least partly.

I can only guess that probably nobody really used this option. ;)

When further applying this logic: on a 2.7 server, the metrics switch would then only enable Ruby profiling, as there's no metrics service... via CHANGELOG entry and a migration in foreman-installer, people having the profiler enabled could get the metrics parameter enabled.

templates/auth.conf.erb
mmoll: That's the legacy API. On newer puppetservers (also 2.x), the proxy should use `/environment_classes` (just below in the file you pasted), but maybe @witlessbird can chime in here...

witlessbird: If you need to list classes, then yes, the new api is /environment_classes.

manifests/params.pp
GregSutcliffe: A migration and changelog sounds good. I agree this is likely an edge case though.

templates/auth.conf.erb
GregSutcliffe: Ah, already done, cool. Nothing to see here then, move along...

GregSutcliffe:
Looks ok, I'll try to find time to test it. Are the Travis failures significant?

manifests/params.pp
mmoll: pushed these changes

mmoll:
travis is running into the 50 minutes timeout again with one testset, I gues there's nothing we can really do, except reducing (or splitting) the test set

manifests/server/puppetserver.pp
```diff
@@ -232,4 +248,16 @@
     ensure  => $product_conf_ensure,
     content => template('puppet/server/puppetserver/conf.d/product.conf.erb'),
   }
+
+  # Do we want to leave this untouched for now?
+  # if versioncmp($server_puppetserver_version, '5.0') >= 0 {
```
mmoll: I guess this can be left out for now... otherwise we would need to provide a lot of configuration possibilities

templates/server/puppetserver/conf.d/auth.conf.erb
```diff
@@ -200,16 +242,29 @@ authorization: {
             sort-order: 500
             name: "puppetlabs static file content"
         },
+<%- if scope.function_versioncmp([@server_puppetserver_version, '5.0']) >= 0 -%>
+        {
+            # Allow all users access to the experimental endpoint
+            # which currently only provides a dashboard web ui.
```
mmoll: this needs maybe a parameter...

manifests/server/puppetserver.pp
GregSutcliffe: You mean possibilities for the `conf.d/metrics.conf` file? Or did I misunderstand?

templates/server/puppetserver/conf.d/auth.conf.erb
GregSutcliffe: For the experimental UI? Sure, or just leave it out until the endpoint matures...

manifests/server/puppetserver.pp
mmoll: exactly, parameters and templating for `conf.d/metrics.conf`

templates/server/puppetserver/conf.d/auth.conf.erb
mmoll: as it's enabled by default OOTB, added a parameter that's also true by default

manifests/init.pp
```diff
@@ -463,9 +463,6 @@
 #                                           can query the puppet-admin-api endpoint
 #                                           Defaults to [ '127.0.0.1', '::1', $::ipaddress ]
 #
-# $server_enable_ruby_profiler::            Should the puppetserver ruby profiler be enabled?
```
ekohl: Can you list this in the changelog under breaking changes or do you want to write those when we release?

manifests/params.pp
```diff
@@ -453,8 +456,26 @@
...
+  if versioncmp($::puppetversion, '5.0.0') < 0 {
+    $server_puppetserver_version    = '2.7.0'
+  } else {
+    $server_puppetserver_version    = '5.0.0'
+  }
+
+  # For Puppetserver 5, use JRuby 9k?
+  $server_puppetserver_jruby9k      = false
+
+  if versioncmp($::puppetversion, '5.0.0') < 0 {
```
ekohl: Maybe just write this as ```$server_puppetserver_metrics = versioncmp($::puppetversion, '5.0.0') >= 0```?

manifests/server/puppetserver.pp
```diff
@@ -128,6 +130,21 @@
       context => "/files${config}",
       changes => "set BOOTSTRAP_CONFIG '\"${bootstrap_paths}\"'",
     }
+
+    if versioncmp($server_puppetserver_version, '5.0') >= 0 {
+      if $server_jruby9k == true {
```
ekohl: The ```== true``` can be dropped.

manifests/init.pp
mmoll: when we release, there will be more than just one thing...

manifests/params.pp
mmoll: I did write this (and some other occurrences) as ternaries now

manifests/server/puppetserver.pp
mmoll: done

manifests/params.pp
mmoll: ok, ternaries do only work with variables, bit further expressions...

mmoll:
This should be good to go now. We really should refactor the puppetserver tests, but I'd like to get this in now to be on track for the 8.0.0 release to be included in Foreman 1.16. 🏎

GregSutcliffe:
Looks good to me, ty @mmoll!

PR #517, opened 2017-07-04
desupport Ubuntu/precise

mmoll:
Ubuntu/precise is EOL since 1st May 2017. This reduces the travis runtime hopefully enough to get under 50 minutes again.


mmoll:
this still times out... could it be that a new version of e.g. rspec-puppet is slower now?

PR #520, opened 2017-07-21
Fix "No supported formats are acceptable (Accept: s, pson)" while bucketing from Puppet Agent v3

sathieu:
See https://bugs.debian.org/869183

### DIFF
--- a/templates/server/config.ru.erb
+++ b/templates/server/config.ru.erb
@@ -79,9 +79,8 @@ class Puppet3Compat
       env['QUERY_STRING'] = Rack::Utils.build_query(@req.params)
     end
 
-    if @api =~ /^file_(content|bucket_file)/ && @req.get?
-      env["HTTP_ACCEPT"] = "binary"
-    elsif @api == "file_bucket_file" && (@req.post? || @req.put?)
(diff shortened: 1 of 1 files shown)


sathieu:
The behavior now matches https://github.com/puppetlabs/puppetserver/blob/ff0e5e84860ea9b035839e77e9b27d4c3a923e36/src/clj/puppetlabs/services/legacy_routes/legacy_routes_core.clj#L27-L42

templates/server/config.ru.erb
```diff
@@ -79,9 +79,8 @@ class Puppet3Compat
       env['QUERY_STRING'] = Rack::Utils.build_query(@req.params)
     end
 
-    if @api =~ /^file_(content|bucket_file)/ && @req.get?
-      env["HTTP_ACCEPT"] = "binary"
-    elsif @api == "file_bucket_file" && (@req.post? || @req.put?)
+    env["HTTP_ACCEPT"] = env["HTTP_ACCEPT"].split(/\s*,\s*/).map { |a| a.sub(/^raw|s$/, 'binary') }.join(', ')
```
ekohl: In the code you linked there's also a distinct which I think should be here as well before joining again.

sathieu: Good catch. Fixed.

mmoll:
I'll watch that Debian bug and make sure this PR will merged for the next patch release to be included for 1.16RC1.

mmoll:
merged, merci @sathieu!

PR #521, opened 2017-07-25
Handle $::processorcount as string

sathieu:


sathieu:
Error is:
```
Class[Puppet]: parameter 'server_passenger_min_instances' expects an Integer value, got String
```

mmoll:
similar case as #512

GregSutcliffe:
ty @sathieu!

PR #523, opened 2017-07-27
Add server_jvm_cli_args parameter

alexjfisher:
Sets the JAVA_ARGS_CLI variable in /etc/sysconfig/puppetserver
This environment variable gets used when `puppetserver gem` is run.

You might need to do this if, for example, your /tmp is mounted with
noexec and you're using the puppetserver_gem package provider...


mmoll:
merged, thanks @alexjfisher!

PR #524, opened 2017-07-27
Manage the puppet user

ekohl:
The use case is that puppet-certs deploys certificates with the owner puppet and now contains a require => Class['puppet::server::install'] but if the User resource is managed then puppet can autorequire it and there is less coupling between modules.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -326,6 +326,8 @@
 #
 # === Advanced server parameters:
 #
+# $server_manage_user::                     Whether to manage the server user resource
+#
 # $server_httpd_service::                   Apache/httpd service name to notify
 #                                           on configuration changes. Defaults
 #                                           to 'httpd' based on the default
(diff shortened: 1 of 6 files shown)


ekohl:
I was a bit hesitant with this initially but I do think this could be a good thing. I've avoided managing properties so packages can set those. Would it be a good idea to manage the group too or should this be sufficient?

GregSutcliffe:
less-coupling++ - I think this should be fine though, the group is less relevant (and I think the puppet server will fix those perms anyway)

alexjfisher:
Yes, RedHat puppetserver RPM seems to fixup group and homedir.

It might be worth watching https://tickets.puppetlabs.com/browse/SERVER-1381 in case the uid ever goes back to being 52.

ekohl:
Rebased and fixed the duplicate resource problem.

mmoll:
merged, bedankt @ekohl!

PR #525, opened 2017-07-28
Fix `client_package` puppet type

alexjfisher:
This parameter gets passed to a Package resource.  It should accept a
string or an array of strings.

See https://github.com/theforeman/puppet-puppet/pull/498#discussion_r129916260


mmoll:
merged, thanks @alexjfisher!

PR #526, opened 2017-07-29
update common_modules_path to work in the server::config subclass

alexcreek:
This allows defining `puppet::server::common_modules_path` to work the same as  `puppet::server_common_modules_path`  in the same fashion as the rest of the class parameters for `puppet::server`.

This fix addresses r10k purging $envpath of non-branch directories then puppet creating $envpath/common on each run triggering a refresh of the puppetserver service.


GregSutcliffe:
Ty @alexcreek!

PR #527, opened 2017-07-31
use match expressions instead of is_*()

mmoll:


GregSutcliffe:
Untested but looks sane to me

ekohl:
Thanks!

PR #528, opened 2017-08-02
Add `puppetserver_trusted_agents` parameter

alexjfisher:
I'm using octocatalog-diff to fetch catalogs from a puppetserver.
By default, agents can only fetch their own catalogs.  I need to
reconfigure the puppetserver so that octocatalog-diff can fetch
*any* catalog. This commit adds a new
`server_puppetserver_trusted_agents` parameter that makes this possible.

See.
https://github.com/github/octocatalog-diff/blob/master/doc/advanced-puppet-master.md#certificate-authorization

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -493,6 +493,9 @@
 #
 # $server_puppetserver_experimental::       For Puppetserver 5, enable the /puppet/experimental route? Defaults to true
 #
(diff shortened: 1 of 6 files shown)


mmoll:
@alexjfisher please rebase. looking good otherwise.

alexjfisher:
@mmoll Rebased

ekohl:
Thanks!

PR #529, opened 2017-08-05
use puppetlabs-hocon for product.conf

mmoll:


mmoll:
This is the first small step. Once this is in, other HOCON files can get converted - except maybe auth.conf, because there's puppetlabs-puppet_authorization.

manifests/server/puppetserver.pp
```diff
@@ -238,14 +238,23 @@
...
+
   if versioncmp($server_puppetserver_version, '2.7') >= 0 {
     $product_conf_ensure = file
+
+    hocon_setting { 'server_check_for_updates':
+      ensure  => present,
+      path    => $product_conf,
+      setting => 'product.check-for-updates',
+      value   => $server_check_for_updates,
+      require => File[$product_conf],
```
ekohl: I'm surprised this isn't autorequired.

mmoll: tbh, I didn't try it without.

ekohl: I checked the code and it's not there so I'm pretty sure it doesn't :)

PR #531, opened 2017-08-09
use puppetlabs-hocon for webserver.conf

mmoll:


manifests/server/puppetserver.pp
```diff
@@ -252,22 +252,112 @@
...
-    content => template('puppet/server/puppetserver/conf.d/auth.conf.erb'),
+  }
+
+  $default_webserver_setting = {
+    ensure  => present,
+    path    => $webserver_conf,
+    require => File[$webserver_conf],
+  }
+
+  $webserver_general_settings = {
```
ekohl: Alternative that I think is a bit less verbose:

```puppet
$webserver_general_settings = {
  'webserver.access-log-config' => "${server_puppetserver_dir}/request-logging.xml",
  # The rest
}

$webserver_general_settings.each |$setting, $value| {
  hocon_setting { $setting:
    ensure  => present,
    path    => $webserver_conf,
    setting => $setting,
    value   => $value,
    require => File[$webserver_conf],
  }
}
```

mmoll: thanks, that's nicer, changed.

manifests/server/puppetserver.pp
```diff
@@ -252,22 +252,90 @@
     content => template('puppet/server/puppetserver/conf.d/puppetserver.conf.erb'),
   }
 
-  file { "${server_puppetserver_dir}/conf.d/webserver.conf":
+  file { "${server_puppetserver_dir}/conf.d/auth.conf":
```
ekohl: Semi-related: would it be better to also manage user/group/permissions on these files?

mmoll: It worked until now the way it was, because the files came with the puppetserver package, but we can add that (later ;)).

ekohl: Later is fine, hence the Semi-related :)

PR #532, opened 2017-08-10
Fixes #20546 - Make passenger SSL settings parameters

ekohl:
This allows overriding through hiera. I chose not to expose them on the top level since the puppetserver uses different words for the same thing and this setup will eventually go away.


mmoll:
merged, bedankt @ekohl!

PR #536, opened 2017-08-14
[WIP] Always manage the ssl_ca_crl if a host is a CA

ekohl:
This fixes an issue where puppetserver is installed through puppet without the puppet user being available. It then falls back to the owner root. This causes puppetserver to refuse to restart.

### DIFF
--- a/manifests/server/config.pp
+++ b/manifests/server/config.pp
@@ -165,6 +165,13 @@ class puppet::server::config inherits puppet::config {
                   Exec['puppet_server_config-create_ssl_dir'],
                   ],
     }
+
+    file { $::puppet::server::ssl_ca_crl:
+      ensure => file,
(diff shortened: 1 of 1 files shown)


GregSutcliffe:
LGTM

ekohl:
Sadly this didn't fix the problem we had so I need to dive in deeper.

mmoll:
@ekohl what's the current status here?

ekohl:
I couldn't quite reproduce it so marking it WIP

mmoll:
what's the status here?

ekohl:
Thanks for reminding. This can be closed since puppetserver 2.8.0 deals with this. It still had something odd, but at least it doesn't fail to start anymore. It was most likely because the package was updated between the report and my reproduction attempts that I could no longer.

PR #537, opened 2017-08-14
use puppetlabs-hocon for puppetserver.conf

mmoll:


manifests/server/puppetserver.pp
```diff
@@ -245,9 +245,88 @@
...
+
+  hocon_setting { 'jruby-puppet.environment-class-cache-enabled':
+    ensure  => $envionment_class_cache_ensure,
+    path    => $puppetserver_conf,
+    setting => 'jruby-puppet.environment-class-cache-enabled',
+    value   => $server_environment_class_cache_enabled,
+    require => File[$puppetserver_conf],
+  }
+
+  # TODO: See https://tickets.puppetlabs.com/browse/HC-100
```
mmoll: this is ugly, but I don't see that value change in reality, anyway

manifests/server/puppetserver.pp
```diff
@@ -245,9 +245,88 @@
...
+      path    => $puppetserver_conf,
+      setting => $setting,
+      value   => $value,
+      type    => $type,
+      require => File[$puppetserver_conf],
+    }
+  }
+
+  if versioncmp($server_puppetserver_version, '2.3') >= 0 {
+    $envionment_class_cache_ensure = present
```
ekohl: ```s/envionment/environment/``` (and below as well)

ekohl:
What should we do with this? The issue was accepted by PL but ofc that's no ETA on a fix ...

mmoll:
unsure, but I think there's no need to rush this in...

ekohl:
In that case let's wait a bit.

PR #540, opened 2017-08-29
Allow setting foreman authentication parameters

LDaneliukas:
When using an external foreman instance via `foreman_url` it may require authentication. This allows setting the user and password via `foreman::puppetmaster`.


ekohl:
First of all the code in this patch looks good but I recently discovered that ENC scripts ([external-node_v2.rb](https://github.com/theforeman/puppet-foreman/blob/902cfa0481e45ba8e3ba8d665eab21490cf3d001/files/external_node_v2.rb) and [foreman-report_v2.rb](https://github.com/theforeman/puppet-foreman/blob/902cfa0481e45ba8e3ba8d665eab21490cf3d001/files/foreman-report_v2.rb)) actually use the username/password parameters. That makes me wonder if this actually solves an issue for you.

Another question: in our usual deployment this is handled through SSL certificate authentication. Did you explicitly configure it to require authentication and is this desired?

Lynce1:
@ekohl, yes, we are using an external Foreman machine and it requires user/password authentication in order to be accessed. When these values are not specified in `foreman.yaml`, the following error can be seen in `puppetserver.log`:

`Puppet Report processor failed: Could not send report to Foreman at https://ommited.foreman.url: Connection refused - Connection refused`

Everything works fine, once the user and password are added.

ekohl:
That is odd, connection refused is typically not a username/password issue but a networking issue. I wonder what happens when you specify an invalid username/password because I can't really see how they are used from the settings.

ekohl:
Closing for inactivity. Please reopen if it's still an issue.

PR #541, opened 2017-09-01
Remove Puppet < 4.5 support

ekohl:
We no longer test with Puppet < 4.0 so we can remove that conditional.  It also does some other style improvements.

By setting the ipaddress to a static value we get consistent output from ip_to_cron and don't rely on it to remain unchanged in facterdb.

### DIFF
--- a/.sync.yml
+++ b/.sync.yml
@@ -9,6 +9,14 @@ Rakefile:
     - manifests/init.pp
 spec/spec_helper.rb:
   extra_code: |
+    if Puppet.version >= '4.0'
+      aio = on_os_under_test.reject do |os, facts|
+        ['FreeBSD', 'DragonFly', 'Windows'].include?(facts[:operatingsystem])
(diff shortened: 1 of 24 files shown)


mmoll:
💔

ekohl:
It looks like there are still conditionals for puppet 3. Should I rip them out?

mmoll:
yes please 👍

ekohl:
So this PR has become a bit larger than anticipated but it was needed.

ekohl:
💚

mmoll:
Could you squash 140d5c5 into 3f8f7cc? Looking good otherwise, tested with native OS Debian packages, resulting in a working passenger puppetmaster setup. 👍

ekohl:
You're right, that's a nicer cleanup which also makes the patch after it easier.

mmoll:
merged, bedankt @ekohl!

PR #542, opened 2017-09-01
use local variables for HOCON auth.conf template

mmoll:


manifests/server/puppetserver.pp
```diff
@@ -104,6 +104,7 @@
   $server_metrics                         = $::puppet::server::puppetserver_metrics,
   $server_experimental                    = $::puppet::server::puppetserver_experimental,
   $server_trusted_agents                  = $::puppet::server::puppetserver_trusted_agents,
+  $allow_header_cert_info                 = $::puppet::server::allow_header_cert_info
```
ekohl: trailing comma

mmoll: fixed.

PR #543, opened 2017-09-02
Set a fixed ipaddress fact in ip_to_cron tests

ekohl:
We don't want to rely on facterdb for this value because they don't guarantee it won't change.

Split from https://github.com/theforeman/puppet-puppet/pull/541 since that'll take a bit longer to properly fix.


mmoll:
merged, bedankt @ekohl!

PR #546, opened 2017-09-05
remove EOL OSes, add newer ones

mmoll:


mmoll:
waiting for a gem relase with https://github.com/camptocamp/facterdb/pull/52 included

mmoll:
BTW, what should we do with Fedora?

ekohl:
> BTW, what should we do with Fedora?

Good question. Since we don't really test it I have no idea if it actually works but that's also true for some of the other versions. Since I only see F24 and F25 RPMs on yum.puppetlabs.com I'd just list those.

mmoll:
https://fedoramagazine.org/fedora-24-eol/

ekohl:
Or just 25 then :)

mmoll:
this should be ready now 💚

ekohl:
Thanks!

PR #552, opened 2017-09-18
Restore the share directory

ares:
since the very first commit with [puppet 4 support](https://github.com/theforeman/puppet-puppet/commit/168ba94ae885acf8b5827a64e2c9609001156c72#diff-2f383a197b1a428561b364755701d60eR84) the shared directory does not seem to be compatible with our [puppet-foreman_scap_client](https://github.com/theforeman/foreman-packaging/blob/rpm/develop/puppet-foreman_scap_client/puppet-foreman_scap_client.spec#L4). We rely on this directory to be included in every environment. I wonder if there's any reason not to keep it to `/usr/share/puppet/modules`? I'd like to avoid changing the packaging definition since the module itself does not really care what puppet version it's being used with.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -280,7 +280,7 @@ class puppet::params {
   $server_envs_dir             = "${codedir}/environments"
   $server_envs_target          = undef
   # Modules in this directory would be shared across all environments
-  $server_common_modules_path  = ["${server_envs_dir}/common", "${codedir}/modules", "${sharedir}/modules"]
+  $server_common_modules_path  = unique(["${server_envs_dir}/common", "${codedir}/modules", "${sharedir}/modules", '/usr/share/puppet/modules'])
 
   # Dynamic environments config, ignore if the git_repo is 'false'
   # Path to the repository
--- a/manifests/server/config.pp
+++ b/manifests/server/config.pp
@@ -129,6 +129,12 @@ class puppet::server::config inherits puppet::config {
     mode   => '0750',
   }
 
+  # Create Foreman share dir which does not depend on Puppet version
+  exec { 'mkdir -p /usr/share/puppet/modules':
+    creates => '/usr/share/puppet/modules',
+    path    => ['/usr/bin', '/bin'],
+  }
+
   ## SSL and CA configuration
   # Open read permissions to private keys to puppet group for foreman, proxy etc.
   file { "${::puppet::server::ssl_dir}/private_keys":
--- a/spec/classes/puppet_server_config_spec.rb
+++ b/spec/classes/puppet_server_config_spec.rb
@@ -158,7 +158,7 @@ describe 'puppet::server::config' do
           if Puppet.version >= '3.6'
             should contain_puppet__config__main("environmentpath").with({'value' => "#{codedir}/environments"})
             should contain_puppet__config__main("basemodulepath").with({
-              'value'  => ["#{codedir}/environments/common","#{codedir}/modules","#{sharedir}/modules"],
+              'value'  => ["#{codedir}/environments/common","#{codedir}/modules","#{sharedir}/modules","/usr/share/puppet/modules"],
               'joiner' => ':'})
           end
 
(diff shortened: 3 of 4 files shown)


ekohl:
We already divert from the default ```/etc/puppetlabs/code/environments/production/modules:/etc/puppetlabs/code/modules:/opt/puppetlabs/puppet/modules``` by adding the ```common``` directory but I'd like to at least include the upstream default here.

Upstream it might also be better to change the ```puppet-foreman_scap_client``` packages to install to a new location and drop puppet 3 support but I realize that downstream it might not be an option.

ares:
do I understand correctly that you'd prefer me keeping sharedir as is and rather changing https://github.com/ares/puppet-puppet/blob/39be646c3b2bc7fd0043810d8f805129c140e04e/manifests/params.pp#L283 
to include 
* /etc/puppetlabs/code/environments/production/modules
* /etc/puppetlabs/code/modules
* and the desired /usr/share/puppet

I don't like the idea of changing puppet-foreman_scap_client install path too much. The module is generally usable with puppet 2.7+. If there are users using it with puppet 3, it would force them to create puppet 4 config directory structure next to it. I think having one shared directory not specific to puppet version for our modules is fine.

ekohl:
@ares yes, I want to at least keep the current list because that's the puppet default (+1 directory) but I have no objection to adding another.

ares:
@ekohl I think it should be now what was asked, code dir was there already, the "/etc/puppetlabs/code/environments/production/modules" does not sound right to me, wouldn't it add all production modules everywhere?

ekohl:
```puppet::server::env``` is a define to manually create an env so that should be correct. Most users now will just be using directory environments which is IMHO a much nicer solution.

**Comment by ares** [comment]
Hm, there were many failures since puppet server won't start in case `/usr/share/puppet` (parent directory) does not exist. So I added creation to server install manifest, if we want to rely that this directory exists in configuration, it make sense to ensure it's there.

**Comment by ares** [comment]
oh, this now fails again since we already define sharedir directory, great... suggestion how to do it cleaner than just changing the sharedir path for p4 and add /etc paths to common dirs welcome

**Comment by ares** [comment]
can't believe my eyes, tests are finally green, @ekohl would you mind taking another look?

File: manifests/params.pp [hunk 1]
```diff
@@ -110,7 +110,7 @@
       $rundir                     = '/var/run/puppetlabs'
       $ssldir                     = '/etc/puppetlabs/puppet/ssl'
       $vardir                     = '/opt/puppetlabs/puppet/cache'
-      $sharedir                   = '/opt/puppetlabs/puppet'
+      $sharedir                   = '/opt/puppetlabs/puppet/modules'
```
ekohl: I think this is incorrect and shouldn't be changed.

manifests/server/install.pp
```diff
@@ -24,6 +24,12 @@
     }
   }
 
+  # Create Foreman share dir which does not depend on Puppet version
```
ekohl: Reminds me of https://github.com/theforeman/puppet-puppet/issues/67.

manifests/server/install.pp
```diff
@@ -24,6 +24,12 @@
...
+  exec { 'mkdir -p /usr/share/puppet/modules':
```
ekohl: Could you move this to ```manifests/server/config.pp```?

manifests/params.pp
ares: Oops, sorry to many versions that I've tried were the cause

manifests/server/install.pp
ares: which seems was later refactoring and then lost when introducing puppet 4 support

ares:
@ekohl I made the changes, I hope it's better now, could you please take another look?

mmoll:
While I don't like it, merging for the sake of compatibility. 👴

mmoll:
merged, díky @ares!

PR #556, opened 2017-10-11
Avoid duplicate declaration issues when server_additonal_settings and additional_settings contain same key

treydock:
In the past I've set `bucketdir` in `[main]` to a value that differs from the value in `[master]` and now I get duplicate declaration with 8.0.3.

This is what I get in unit tests for my roles module:

```
Evaluation Error: Error while evaluating a Resource Statement, Duplicate declaration: Puppet::Config::Additional_settings[bucketdir] is already declared in file /Users/tdockendorf/puppet/osc-puppetmaster-conf/modules/puppet/manifests/config.pp:69; cannot redeclare at /Users/tdockendorf/puppet/osc-puppetmaster-conf/modules/puppet/manifests/server/config.pp:120 at /Users/tdockendorf/puppet/osc-puppetmaster-conf/modules/puppet/manifests/server/config.pp:120:3
```


mmoll:
Could you delete the comment above that section as well? This was re-implemented like this because we needed to stay Puppet 3.x compatible back then.

treydock:
@mmoll Should I also replicate this style change for agent?  Not an issue for me but I am a fan of consistency.

mmoll:
@treydock that would be nice 👍

treydock:
@mmoll changes made.  Use of `puppet::config::additional_settings` defined type completely removed in favor of using iterations.

manifests/agent/config.pp
```diff
@@ -36,12 +36,10 @@
...
 
-  # we need to store this in a variable, because older puppet doesn't
-  # like resource{function(): ... }
-  $additional_settings_keys = keys($::puppet::agent_additional_settings)
-  puppet::config::additional_settings{ $additional_settings_keys:
-    hash     => $::puppet::agent_additional_settings,
-    resource => '::puppet::config::agent',
+  $::puppet::agent_additional_settings.each |$key,$value| {
+    puppet::config::agent {
+      $key: value => $value,
```
ekohl: Nit: typically we place the name of the resource at the same line as the resource. So ```$key``` should be on the same line as ```puppet::config::agent```.

treydock:
@ekohl Made the changes

mmoll:
merged, thanks @treydock!

Issue #472, opened 2017-01-23
Specifying puppetserver version a bit confusing.


claflico:
Trying to migrate an existing puppet setup from using jlamber121 module to this module so that I can upgrade versions. Am currently running the following versions on CentOS 7:

puppetlabs-release-pc1-1.0.0-2.el7.noarch
puppet-agent-1.4.2-1.el7.x86_64
puppetexplorer-2.0.0-1.noarch
puppetdb-4.0.2-1.el7.noarch
puppetserver-2.3.2-1.el7.noarch
puppetdb-termini-4.0.2-1.el7.noarch

Trying to specify via hiera: 
puppet::version: '1.4.2'
puppet:server_puppetserver_version: '2.3.2'

results in puppet trying to install 
/bin/yum -d 0 -e 0 -y downgrade puppetserver-1.4.2

After digging into the manifests and testing different variables I found that I have to specify it twice to get everything to work properly:
puppet::server_puppetserver_version: '2.3.2'
puppet::server_version: '2.3.2'

mmoll:
This is, IMHO, a documentation issue and really a problem. As most people just use the latest version of Puppet/Puppetserver it's not often visible. The thing is that version and server_version are package (RPM/DEB) version, while server_puppetserver_version is the application version of Puppetserver to generate the configs correctly for the specific version.

preisbeck:
Just stumbled over this myself. After working with the module myself for several weeks now, I never noticed the second version declaration.

mmoll:
It can't be done in the module, as e.g. Debian and FreeBSD do have subversions in their packaging.

ekohl:
I'm not sure how to proceed with this. Should we extend the README a bit?

PR #560, opened 2017-10-20
Fixes 21401

jdshewey:
; Since the requirement of puppet 4 via the RPM, requring puppet 3 needed to be changed to the puppet 4 equivalent.

### DIFF
--- a/manifests/server/install.pp
+++ b/manifests/server/install.pp
@@ -33,7 +33,7 @@ class puppet::server::install {
                                       default => ['puppetmaster-common', 'puppetmaster'],
                                     },
         /^(FreeBSD|DragonFly)$/ => [],
-        default                 => ['puppet-server'],
+        default                 => ['puppetserver'],
       },
       'puppetserver' => 'puppetserver',
     }
--- a/metadata.json
+++ b/metadata.json
@@ -12,7 +12,7 @@
     "foreman",
     "puppet",
     "puppetmaster",
-    "puppet-server"
+    "puppetserver"
   ],
   "dependencies": [
     {
--- a/spec/classes/puppet_server_spec.rb
+++ b/spec/classes/puppet_server_spec.rb
@@ -10,7 +10,7 @@ describe 'puppet::server' do
         ssldir = '/var/lib/puppet/ssl'
       end
 
-      server_package = 'puppet-server'
+      server_package = 'puppetserver'
       if facts[:osfamily] == 'Debian'
         server_package = 'puppetmaster'
         if facts[:puppetversion].to_f > 4.0


ekohl:
I think this is correct and reminds me I need to finish https://github.com/theforeman/puppet-puppet/pull/541

ekohl:
Another detail: if you use ```#21401``` instead of ```21401``` our automation will pick it up and relate one to the other.

manifests/server/install.pp
```diff
@@ -33,7 +33,7 @@
                                       default => ['puppetmaster-common', 'puppetmaster'],
                                     },
         /^(FreeBSD|DragonFly)$/ => [],
-        default                 => ['puppet-server'],
+        default                 => ['puppetserver'],
       },
       'puppetserver' => 'puppetserver',
```
mmoll: shouldn't this be used automatically as `$::puppet::server::implementation` should be set accordingly?

jdshewey: I believe this IS where $::puppet::server::implementation is defined:

$server_package_default = $::puppet::server::implementation ? {
      'master'       => $::osfamily ? {
        'Debian'                => $puppet4 ? {
                                      true    => ['puppet-master'],
                                      default => ['puppetmaster-common', 'puppetmaster'],
                                    },
        /^(FreeBSD|DragonFly)$/ => [],
        default                 => ['puppet-server'],
      },
      'puppetserver' => 'puppetserver',

mmoll: nope.
`$server_package_default` is deduced from `$::puppet::server::implementation`, which in turn should be `puppetserver`, if Puppet 4 AIO packages are used.

```puppet
    $server_package_default = $::puppet::server::implementation ? {
      'master'       => $::osfamily ? {
[...]
      'puppetserver' => 'puppetserver',
    }
```

jdshewey: Right. IF puppet 4 AIO packages are used BEFORE installing the foreman-installer.rpm (which is not noted in the quick start guide.) If puppet 3 packages are used, the installer fails because the forman-installer.rpm requires puppet v4's puppet-agent.rpm package, but cascades no dependency for puppetserver.rpm (v4), so this package doesn't install if you do a `yum -y install katello` . This means that puppet-server.rpm installation isn't attempted until after you run `foreman-installer --scenario katello` at which point this puppet module defaults to and tries to install puppet-server.rpm (v3) causing a dependency conflict with puppet-agent.rpm. if you have Puppet 3 installed, the foreman-installer.rpm fails to install because it can't find the puppet-agent.rpm dependency. Therefore the only way (and proper way) to fix this is to remove all support for v3, which isn't going to work anyway.

Go ahead and give a fresh installation (following the quick start guide exactly) a try if you aren't sure what I'm blathering on about - you can't miss it.

mmoll: Sounds like an RPM issue outside the scope of the puppet-puppet module then?

mmoll: looking at the redmine issue: 1.16 and 1.17 indeed do not work with Puppet 3.x, while 1.15 should.

jdshewey: So, it sounds like this needs to be merged for 1.16 and 1.17, and another bug filed against 1.15 to have the RPM dependencies fixed. If it's just working with a spec file, I can probably commit that update, but if the spec files for packaging are generated out of ruby, I'm sure to get lost pretty quickly...

mmoll: see https://github.com/theforeman/foreman-packaging, the RPMs are generated from their respectibe branches i.e. rpm/1.15, rpm/1.16, rpm/develop (the later for 1.17/nightly)

mmoll: Regarding "this needs to be merged for 1.16 and 1.17": I still think the current puppet-puppet code is fine as-is by the explanation above.

mmoll:
Maybe @ekohl or @ehelms do have some more insights here as I don't really use RPMs currently, but I think the nightly systest (if not blocked otherwise) is fine.

manifests/server/install.pp
ehelms: Sorry, I am having a hard time following this. Here is what I know.

For 1.15, puppet 3 and puppet 4 are supported. The foreman-installer package, in 1.15., properly handles that (https://github.com/theforeman/foreman-packaging/blob/rpm/1.15/foreman-installer/foreman-installer.spec#L19).

For 1.16 and 1.17, only Puppet 4 is supported. 

I think perhaps this puppet module (which is treated as providing stand alone puppet management support and is also used by the installer) is still striving to manage both Puppet 3 and Puppet 4/5.

The quickstart guide is out of date and should be updated. Please file a docs bug or a PR to help us fix it up! Note that the quickstart guide does not apply to Katello which I noticed you mentioned in there. In the case of Katello, this guide is the only guide we support (https://theforeman.org/plugins/katello/nightly/installation/index.html). And we are aware that this varies from Foreman installs and are working to bring these two in line.

Does this help @jdshewey ? I'd like to work to ensure there are as few holes for users to fall into as possible with installs.

jdshewey: Some. I see this requires puppet 3:

    Requires:   curl
    Requires:   %{?scl_prefix}puppet >= 3.6.0

But I'm not seeing how this allows for the installation of puppet-agent, the new name of puppet in version 4. Secondly, I do not see where this installer requires the "puppet-server" rpm (v3) or "puppetserver" (v4) rpm, which are also required dependencies. From the perspective of the RPM, this is the problem, unless there is some code I missed. This code is used by all scenarios for the installer, correct?

As for Katello, am I correct in inferring from your link that only the nightly unstable build is supported? Or can Katello be installed from 1.15?

jdshewey: Also, Doesn't the nightly systest use vagrant or ansible without actually walking through installing the RPMs and walking through forman-install --scenario katello?

jdshewey: Also, @mmoll if you think it is fine as is, why don't you try giving the install a go for yourself. It most decidedly runs into issues.

jdshewey: This appears to have been reported by someone else [in January as well](https://github.com/theforeman/puppet-puppet/issues/472)

mmoll: The referenced issue has nothing to do with the choice which puppet master/server package is chosen for install, only about the RPM/DEB version of the respective package and the puppetserver configfile format.

mmoll:
I tested everything on a clean (as in "freshly OS install") system and everything is working fine. puppet-puppet does install puppetserver with Puppet 4 or 5 AIO packages installed, so this PR is not correct.

Furthermore the check for AIO Pupept packages is working. If I do have Puppet 3 installed and try to install foreman-installer (nightly version in this case):
```
[...]
Error: Package: 1:foreman-installer-1.17.0-0.develop.201710260324gitb296be2.el7.noarch (foreman)
           Requires: puppet-agent >= 1.9.0
[...]
```

For 1.15 the situation might be different there, however, the behaviour described in the redmine issue is not happening on a clean system.

My rough guess is that Puppet 3 was installed on that system and foreman-installer executed with it (thus, filling `/etc/foreman-installer/scenarios.d/foreman-answers.yaml` with the Puppet 3.x values) and then Puppet got updated to 4.x, but the answers are still for Puppet 3.x.

The easiest solution is to start over with a *really* clean system.

Issue #564, opened 2017-11-21
Cannot set `codedir` in `puppet.conf`


exptom:
I am using the latest version of the module (8.0.4) with Puppet 4.8 (the version packaged in debian stable).

I am configuring the module using hiera and have specified the codedir as follows:
`puppet::codedir: '/etc/puppet/code'`

However the `codedir` option is never added to the `puppet.conf` file. All other options that I have set using hiera are added to the `puppet.conf` file.
Am I doing something wrong or is this a bug?

ekohl:
We do use it in ```templates/server/puppetserver/conf.d/puppetserver.conf.erb``` but that's ```puppetserver``` which Debian doesn't ship. AFAIK it does ship the (deprecated) ```puppet-server``` which I'm guessing we don't really test with.

I'm not exactly sure which is the best section for this but I'm guessing main so ```puppet module install``` will honor it too. Can you test if adding the following works:
```puppet
puppet::config::main { 'codedir':
  $::puppet::codedir,
}
```

I'd also note that ```/etc/puppet/code``` should be the default on Debian.

exptom:
Please excuse my ignorance. Where exactly should I be adding that line to test?
In one of the manifests of this module?

ekohl:
It shouldn't matter really. Could also be in your profile module or you could set it via ```$additional_settings``` since those should end up in the ```[main]``` section too.

exptom:
I've added it to additional_settings and that has worked. Thanks :)

ekohl:
I've submitted https://github.com/theforeman/puppet-puppet/pull/566 to do this out of the box. We have a parameter for it and we should respect it. When that's merged you will get a duplicate definition but should be easy to fix.

exptom:
@ekohl Thank you for your help with this

PR #566, opened 2017-11-29
Set the codedir in puppet.conf

ekohl:
In GH-564 it was pointed out this is not propagated to the actual puppet.conf even though we have a setting for it.


mmoll:
merged, bedankt @ekohl!

PR #567, opened 2017-11-30
Prepare a 8.1.0 release

ekohl:
I think this should be 8.1.0 but I'd consider a 9.0 release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,23 @@
 # Changelog
 
+## 8.1.0
+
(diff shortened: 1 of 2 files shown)


ekohl:
We're having some issues with the ```parallel_tests``` gem and I think our concurrency is too high in some cases. A worker is killed but doesn't report back.

akofink:
Ok, doesn't seem like a blocker. :wink:

mmoll:
#566 can also get added now

Issue #571, opened 2018-01-03
Default instance/memory values can create unstable configurations with many processors


kasimon:
The default instance number/memory configuration for servers with more than 2GB memory

https://github.com/theforeman/puppet-puppet/blob/6735410e6346b616c7875e23af7638b8d6a9c01f/manifests/params.pp#L411

can result in unsuitable configurations for servers with more than 4 processors. According to the [official tuning guide](https://puppet.com/docs/puppetserver/5.1/tuning_guide.html), each server instance should have 512MB memory available. Since the java memory is limited by default at 2GB, but the number of instances unconditionally equals the number of processors, the puppetserver process will run with not enough memory and may quit working at load peaks.

My suggestions would be:
* Either limit instances too: `default_instances =  min(number_processors, 4)`
* or calculate possible instances from number of processors and available memory: `default_instances = min (number_processors, (memory_in_mb / 512) -1)`
* or at least log a warning when more than instances will be configured than fitting into the configured memory

ekohl:
> Either limit instances too: default_instances = min(number_processors, 4)

https://github.com/theforeman/puppet-katello/blob/986a17ea3007f7e5c6ff7c8e9fdb74bef55507c6/manifests/params.pp#L32 implements a similar change and I'd accept a patch that implements this. We don't implement real automatic tuning anyway.

PR #572, opened 2018-01-03
Make performance tuning defaults more safe

kasimon:
Only use 2G for puppetserver if at least 3G available. Don't use more than four instances, as 512MB memory per instance are recommended. 

Closes: #571

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -408,10 +408,10 @@ class puppet::params {
   } else {
     $mem_in_mb = 0 + $::memorysize_mb
   }
-  if $mem_in_mb >= 2048 {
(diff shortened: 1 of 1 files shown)


kasimon:
Unfortunately I have no idea what failure travis-ci is complaining about.

ekohl:
I think we had that for a while and I haven't been able to pin it down. Maybe we need to lower the number of parallel tests again. You can ignore it for now.

@mmoll any objections to merging this?

mmoll:
looks fine to me, merged, danke @kasimon!

kasimon:
Gern :)

PR #575, opened 2018-01-17
restart httpd when changing cert

sean797:
Currently if you `rm -f $::puppet::server::ssl_cert` and run foreman-installer it will regenerate certs but wont restart the service.

### DIFF
--- a/manifests/server.pp
+++ b/manifests/server.pp
@@ -487,11 +487,9 @@ class puppet::server(
   if $implementation == 'master' {
     $pm_service   = !$passenger and $service_fallback
     $ps_service   = undef
-    $rack_service = $passenger
   } elsif $implementation == 'puppetserver' {
     $pm_service   = undef
     $ps_service   = true
-    $rack_service = false
   }
 
   class { '::puppet::server::install': }
@@ -501,7 +499,7 @@ class puppet::server(
     httpd_service => $httpd_service,
     puppetmaster  => $pm_service,
     puppetserver  => $ps_service,
-    rack          => $rack_service,
+    rack          => $passenger,
   }
   -> Class['puppet::server']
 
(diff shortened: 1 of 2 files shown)


ekohl:
Any idea why the ```server.pp``` isn't taking care of this?
https://github.com/theforeman/puppet-puppet/blob/08f9ad96a16989aa6276809c6e40c909d3a257ce/manifests/server.pp#L498-L499

sean797:
Because when httpd uses the same certs Puppet does and these values are set:

https://github.com/theforeman/puppet-puppet/blob/08f9ad96a16989aa6276809c6e40c909d3a257ce/manifests/server.pp#L491-L495

`::puppet::server::service` doesn't restart httpd:
https://github.com/theforeman/puppet-puppet/blob/08f9ad96a16989aa6276809c6e40c909d3a257ce/manifests/server/service.pp#L50-L63

I'll have a look at fixing that tomorrow.

ekohl:
I'm not sure I follow this. First of all, this is only a problem when using the passenger based implementation, correct? Is it a problem that Apache isn't reloading (i.e. still serving the old certs) or is just reloading the app sufficient?

sean797:
> Is it a problem that Apache isn't reloading (i.e. still serving the old certs)

Yes

manifests/server.pp
```diff
@@ -487,11 +487,9 @@
   if $implementation == 'master' {
     $pm_service   = !$passenger and $service_fallback
     $ps_service   = undef
-    $rack_service = $passenger
   } elsif $implementation == 'puppetserver' {
     $pm_service   = undef
     $ps_service   = true
-    $rack_service = false
```
sean797: I'm not sure why this was false to start with? I don't see how `$implementation == 'puppetserver'` and using passenger are mutually exclusive.

I guess if Puppet and Apache aren't sharing the same certs then I guess this change could potentially restart, Apache when something changed in Puppet also.

ekohl:
I need some time to revisit this. Not sure I will have any until after cfgmgmtcamp.

mmoll:
what's the status here?

ekohl:
In light of https://github.com/theforeman/puppet-puppet/pull/650 I'm closing this.

PR #582, opened 2018-03-02
Allowing the package_source to be an Httpurl

MAXxATTAXx:
Allow httpurl arguments for the pacakge reosurce.

In previous releases this value was not checked and when used with providers like chocolatey, it can support httpurl. The module should not restrict this parameter type so heavily.


rodolfosalgado:
This is awesome, I had to do the same thing for me...

MAXxATTAXx:
Not sure where that test shuld go. puppet_agent_install_spec.rb?

ekohl:
Since that doesn't actually check the parameter type I think it needs to go in ```spec/classes/puppet_init_spec.rb```

ekohl:
Thanks!

PR #584, opened 2018-03-07
Use the correct Stdlib::HTTPUrl

ekohl:
96a514149a35defcfbffefe085a0a9bebf920873 allowed an additional type but it's not matching the case of the type. Somehow the tests do pass so Puppet isn't that particular about it but it looks like Kafo is.


mmoll:
I re-ran The Puppet 5 test and it came back green, also this patch fixes the problem for me, too :+1:

mmoll:
merged, bedankt @ekohl!

PR #585, opened 2018-03-08
WIP: Add systemd_randomizeddelaysec

jcharaoui:
Allow configuration of the RandomizedDelaySec parameter of the systemd
timer, which allows randomizing Puppet runs in a way that lets the
administrator see exact moment of the next Puppet run. There's no
incompatibility between Puppet's internal splay and this, but both
probably shouldn't be used together.

This probably still needs adjustments to existing tests.

### DIFF
--- a/spec/classes/puppet_agent_service_systemd_spec.rb
+++ b/spec/classes/puppet_agent_service_systemd_spec.rb
@@ -68,6 +68,9 @@ describe 'puppet::agent::service::systemd' do
             should contain_file('/etc/systemd/system/puppet-run.timer').
               with_content(/.*OnCalendar\=\*-\*-\* \*\:10,40:00.*/)
 
+            should contain_file('/etc/systemd/system/puppet-run.timer').
+              with_content(/.*RandomizedDelaySec\=0.*/)
+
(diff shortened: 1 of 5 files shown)


ekohl:
@jcharaoui this is marked as WIP:. Does that mean it's not ready for review?

jcharaoui:
The functionality is ready for review. I marked it WIP because I didn't write a test case for it. Should be pretty simple but unfortunately I'm unfamiliar with puppet module tests.

spec/classes/puppet_agent_service_systemd_spec.rb
```diff
@@ -68,6 +68,9 @@
             should contain_file('/etc/systemd/system/puppet-run.timer').
               with_content(/.*OnCalendar\=\*-\*-\* \*\:10,40:00.*/)
 
+            should contain_file('/etc/systemd/system/puppet-run.timer').
+              with_content(/.*RandomizedDelaySec\=0.*/)
```
ekohl: I think this can even be ```/^RandomizedDelaySec\=0$/``` since it checks lines. Not sure though.

mmoll:
merged, merci @jcharaoui!

PR #587, opened 2018-04-04
Allow reports to be disabled

sanyu:
This PR adds support for disabling reports


sanyu:
@mmoll I've added some tests, but the bot added the "Need testing" label again, is there anything else I'm missing?

mmoll:
@sanyu That's fine, it means rather "testing by a human", on a real system or similar. :panda_face:

ekohl:
Thanks!

@mmoll I realized too late I didn't squash the commits :(

PR #590, opened 2018-04-19
adding support for Amazon Linux

RobReus:
This has been tested on the latest Amazon Linux AMI in agent mode. 

Fixes:
`Error: /Stage[main]/Puppet::Agent::Service::Systemd/Service[puppet-run.timer]: Provider systemd is not functional on this host`

I have added "2017" to the regex of $agent_restart_command and $unavailable_runmodes. Without it, the code will try to stop the systemd timer which will fail because systemd is not being used on Amazon Linux (yet).


RobReus:
@mmoll so far there's no 2018 version yet. When it comes out, we'll just adjust the regex to: `/^(2|5|6|201(7|8))$/`. Ain't it a beauty :P

PR #600, opened 2018-06-16
make foreman puppet module dir optional

bastelfreak:


bastelfreak:
This is currently work in progress. I'm not sure yet if the tests pass and I also need to adjust the docs. @ekohl @mmoll feedback is highly appreciated :)

----

I noticed that the module creates this directory. I never saw a foreman installation that uses this so I want to make it optional (or get rid of the exec at all). If there are usecases for the directory I will change the default from false to true.

manifests/server/config.pp
```diff
@@ -125,9 +125,11 @@
   }
 
   # Create Foreman share dir which does not depend on Puppet version
-  exec { 'mkdir -p /usr/share/puppet/modules':
-    creates => '/usr/share/puppet/modules',
-    path    => ['/usr/bin', '/bin'],
+  if ($puppet::server::create_foreman_modules_directory == true) {
```
ekohl: I think we can avoid a parameter:
```puppet
if '/usr/share/puppet/modules' in $puppet::server::common_modules_path {}
```

bastelfreak:
@ekohl I updated the code

manifests/server/config.pp
```diff
@@ -124,10 +124,12 @@
     mode   => '0750',
   }
 
-  # Create Foreman share dir which does not depend on Puppet version
-  exec { 'mkdir -p /usr/share/puppet/modules':
-    creates => '/usr/share/puppet/modules',
-    path    => ['/usr/bin', '/bin'],
+  if ('/usr/share/puppet/modules' in $puppet::server::common_modules_path) {
```
ekohl: the parenthesis aren't needed

mmoll:
merged, danke @bastelfreak!

PR #603, opened 2018-06-25
Add ability to configure custom_trusted_oid_mapping.yaml

### LINKED ISSUE #602: /etc/puppetlabs/puppet/custom_trusted_oid_mapping.yaml should be manageable
In setups which use trusted facts, /etc/puppetlabs/puppet/custom_trusted_oid_mapping.yaml needs to be configured to map the OID to a usable variable name.

See https://puppet.com/docs/puppet/5.3/ssl_attributes_extensions.html#extension-requests-permanent-certificate-data for more information

baurmatt:
This commit adds support to configure the
$confdir/custom_trusted_oid_mapping.yaml as described in
https://puppet.com/docs/puppet/5.5/config_file_oid_map.html.

Fixes GH-602

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -473,6 +473,9 @@
 #                                           can query the certificate-status endpoint
 #                                           Defaults to [ '127.0.0.1', '::1', $::ipaddress ]
 #
+# $server_custom_trusted_oid_mapping::      A hash of custom trusted oid mappings. Defaults to undef
+#                                           Example: { 1.3.6.1.4.1.34380.1.2.1.1 => { shortname => 'myshortname' } }
+#
 # $server_admin_api_whitelist::             The whitelist of clients that
 #                                           can query the puppet-admin-api endpoint
 #                                           Defaults to [ '127.0.0.1', '::1', $::ipaddress ]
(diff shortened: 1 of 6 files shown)


manifests/init.pp
```diff
@@ -182,7 +182,7 @@
 # $environment::                            Default environment of the Puppet agent
 #
 # $agent_additional_settings::              A hash of additional agent settings.
-#                                           Example: {stringify_facts => true}
+#                                           Example: {stringify_facts                          => true}
```
ekohl: Could you revert this change and other whitespace only changes?

types/custom_trusted_oid_mapping.pp
```diff
@@ -0,0 +1,4 @@
+type Puppet::Custom_trusted_oid_mapping = Hash[String, Struct[{
```
ekohl: We have a bug in our installer that it currently doesn't handle multi line types because it does incorrect manual parsing. We should be fixing it but in the short term I wonder if we should put it on a single line.

spec/classes/puppet_server_config_spec.rb
```diff
@@ -832,6 +836,29 @@
...
+                },
+                '1.3.6.1.4.1.34380.1.2.1.2' => {
+                  shortname => 'myothershortname',
+                },
+              }
+           }"
+        end
+
+        it 'should have a configured custom_trusted_oid_mapping.yaml' do
+          should contain_file("#{confdir}/custom_trusted_oid_mapping.yaml").
```
ekohl: We have a `verify_exact_contents` helper which might be easier to use.

manifests/init.pp
baurmatt: Done! Sorry! :/ Are the whitespace changes in server.pp ok?

types/custom_trusted_oid_mapping.pp
baurmatt: Sure!

manifests/init.pp
ekohl: Yes, indenting is no problem.

spec/classes/puppet_server_config_spec.rb
baurmatt: This is really helpful, thanks! :)

baurmatt:
Fixed review comments, thanks! :)

ekohl:
Thanks!

PR #605, opened 2018-07-07
Support Ubuntu/bionic, drop Debian 7 & Fedora 25

mmoll:


manifests/params.pp
```diff
@@ -330,7 +330,7 @@
     'Debian' : {
       $agent_restart_command = "/usr/sbin/service ${service_name} reload"
       if (
-        ($::operatingsystem == 'Debian') and (versioncmp($::operatingsystemrelease, '8.0') >= 0) or
+        ($::operatingsystem == 'Debian') or
         ($::operatingsystem == 'Ubuntu') and (versioncmp($::operatingsystemrelease, '15.04') >= 0)
```
ekohl: I know these braces were already here, but they're weird. Technically they're correct but they don't add anything either.

spec/classes/puppet_agent_service_spec.rb
```diff
@@ -23,7 +23,7 @@
           should contain_class('puppet::agent::service::cron').with(:enabled => false)
         end
         case os
-        when /\Adebian-(8|9)/, /\A(redhat|centos|scientific)-7/, /\Afedora-/, /\Aubuntu-16/, /\Aarchlinux-/
+        when /\Adebian-(8|9)/, /\A(redhat|centos|scientific)-7/, /\Afedora-/, /\Aubuntu-(16|18)/, /\Aarchlinux-/
```
ekohl: If we've dropped Debian 7 then it can be simplified to `/\Adebian-/`.

manifests/params.pp
```diff
@@ -329,10 +329,10 @@
   case $::osfamily {
     'Debian' : {
       $agent_restart_command = "/usr/sbin/service ${service_name} reload"
-      if (
-        ($::operatingsystem == 'Debian') and (versioncmp($::operatingsystemrelease, '8.0') >= 0) or
-        ($::operatingsystem == 'Ubuntu') and (versioncmp($::operatingsystemrelease, '15.04') >= 0)
-      ) {
+      if
+      $::operatingsystem == 'Debian' or
```
ekohl: Indenting looks a bit weird here. Would it be good to keep parenthesis around the entire statement? `if (..) {`

mmoll: Done. I did put the whole condition into one long line as it would look more strange because of the formatting rules we have in place otherwise.

PR #609, opened 2018-07-18
allow unauthenticated service endpoint

carazzim0:
Hi there,

since the latest update today we could not connect to the endpoint which we use for monitoring purposes.

Please have a look and let me know if there's something missing.

Thanks in advance
Sebastian


mmoll:
@ekohl could GH-615 go in first?

PR #610, opened 2018-07-23
Simplify puppet::agent::daemon

ekohl:
With Puppet 4 data types we can simplify the complicated case statement to a if/else.


mmoll:
merged, bedankt @ekohl!

Issue #611, opened 2018-07-25
puppet_ca_proxy doesn't pass a valid client cert to upstream ca


gonoph:
When puppet_ca_proxy is defined, apache issues this error message when trying to utilize proxy:

```log
Proxy client certificate callback: (:8140) downstream server wanted client certificate but none are configured
```

Recommended changes to [passenger.pp](/theforeman/puppet-puppet/tree/master/manifests/server/passenger.pp) :
1. create `/path/to/combined.pem` from `$ssl_cert` and `$ssl_cert_key`
2. add `SSLProxyMachineCertificateFile /path/to/combined.pem`after `ProxyPassMatch` as custom fragment.

bkearney:
@gonoph is this tied to an upstream issue? Ping @tbrisker for visibility on answer.

mmoll:
@gonoph would you open a PR to the 10.1-stable branch? Otherwise I'd close this one due to GH-650

mmoll:
closing...

PR #613, opened 2018-07-25
puppet_ca_proxy - include cert/key combo into apache vhosts

gonoph:
Change passenger.pp to combine the ssl_cert and ssl_cert_key into a combined
cert file, which can then be used for SSLProxyMachineCertificateFile
directive. Add the new directive to allow the ProxyPassMatch directive to
work with some protected URLs on the master CA. (ex:
certificate_revocation_list).

Fixes GH-611

### DIFF
--- a/manifests/server/combine_certs.pp
+++ b/manifests/server/combine_certs.pp
@@ -5,17 +5,17 @@ define puppet::server::combine_certs(
   $cert     = $::puppet::server::ssl_cert,
   $key      = $::puppet::server::ssl_cert_key,
 ) {
-    $dir = dirname($combined)
+  $dir = dirname($combined)
 
-    $file_cert = file($cert)
-    $file_key  = file($key)
+  $file_cert = file($cert)
+  $file_key  = file($key)
 
-    file { $dir:
-      ensure => directory,
-    }
-
-    file { $combined:
-      ensure  => file,
-      content => "${file_cert}${file_key}",
-    }
+  file { $dir:
+    ensure => directory,
   }
+
+  file { $combined:
+    ensure  => file,
+    content => "${file_cert}${file_key}",
+  }
+}
--- a/manifests/server/combine_certs.pp
+++ b/manifests/server/combine_certs.pp
@@ -1,0 +1,21 @@
+# combine the public x509 certificate and private key into one file
+#
+define puppet::server::combine_certs(
+  $combined = $::puppet::server::passenger::ssl_combined,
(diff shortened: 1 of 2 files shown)


manifests/server/passenger.pp
```diff
@@ -57,10 +60,16 @@
   ]
 
   if $puppet_ca_proxy and $puppet_ca_proxy != '' {
+    puppet::server::combine_certs { $ssl_combined:
```
ekohl: Because the passenger setup is deprecated (IIRC Puppet 6 will remove it), I'd prefer to inline the resources here. It's not many

manifests/server/combine_certs.pp
```diff
@@ -0,0 +1,21 @@
+# combine the public x509 certificate and private key into one file
+#
+define puppet::server::combine_certs(
+  $combined = undef,
+  $cert     = undef,
+  $key      = undef,
+) {
+  $dir = dirname($combined)
+
+  $file_cert = file($cert)
```
ekohl: This is probably not going to work. Here you assume the file is served as a static file from inside this module. Another problem is that this is done server side while the passenger setup often just assumes local certificates exist. Worst case you're going to end up with a puppet server side certificate (SSLProxyMachineCertificateFile) on the puppet agent side. That can be a security issue.

You'll likely need to use https://github.com/puppetlabs/puppetlabs-concat to combine the files.

gonoph: I was testing on Red Hat's Satellite 6.3 which supports puppet 3 and 4. puppetlabs-concat requires > puppet 4.7.0. I was hoping to use something that was compatible with both. I guess that whatever module creates the certificates as part of the install, should probably create the concatenated file that SSLProxyMachineCertificateFile requires?

ekohl: There are versions of puppetlabs-concat that allow older versions but have the same API. We already rely on it heavily. The thing is that Puppet 3 is really deprecated and running with passenger is as well. With Puppet 6 around the corner I'm more leaning to removing the support than fixing any issues left.

Satellite 6.4 will drop Puppet 3 support as well and require you to update to Puppet 4 while you're on Satellite 6.3.

It is true that we don't have similar proxy code for use with puppetserver though so maybe the question is how we can facilitate that if needed.

ekohl:
I'm going to close this since we're moving away from the passenger setup which follows Puppetlabs. Soon I'll spend time to formally deprecate parts when I'm going to look at supporting Puppet 6.

PR #623, opened 2018-08-24
Make `splaylimit` & `runinterval` types consistent

alexjfisher:
Previously, `splaylimit` defaulted to a String and wouldn't accept
Integers.  `runinterval` defaulted to an Integer but wouldn't accept
Strings.


ekohl:
Thanks!

PR #626, opened 2018-08-27
removing redundant > character

attachmentgenie:
leaving it in breaks your puppet server setup


mmoll:
merged, bedankt @attachmentgenie!

PR #627, opened 2018-08-28
Use `contain` instead of `anchor` pattern

alexjfisher:


mmoll:
merged, thanks @alexjfisher!

alexjfisher:
@ekohl classes aren’t implicitly contained when you use `class {}`

https://puppet.com/docs/puppet/5.5/lang_containment.html#the-contain-function

> To contain classes that are declared with the resource-like declaration syntax, use the contain function after declaring the class

ekohl:
@alexjfisher Thanks, that's great to get confirmed.

PR #628, opened 2018-09-05
Remove redundant symlink

ekohl:
With recent versions of puppetlabs_spec_helper this is done automatically


mmoll:
merged, bedankt @ekohl!

PR #629, opened 2018-09-14
Pin networking.ip fact in specs

ekohl:
puppet-extlib ip_to_cron switched to the networking.ip fact. This pins the ip if networking is present to maintain compatibility with Facter 2.


mmoll:
I guess the expectation needs updating, too?

ekohl:
I ran it locally and it worked. @alexjfisher what do you think of a fallback to the old fact in the function?

ekohl:
Closing in favor of https://github.com/theforeman/puppet-puppet/pull/636

Issue #630, opened 2018-09-19
Add shared gem path for Puppetserver >= 5.3


baurmatt:
https://github.com/theforeman/puppet-puppet/blob/73f7fe6cb344b1f00ff15e8779097e55bb0bf274/templates/server/puppetserver/conf.d/puppetserver.conf.erb#L19

This needs to be extend for Puppetserver >= 5.3 due to the introduced [shared gem folder](https://puppet.com/docs/puppetserver/5.3/release_notes.html#new-features-2) `/opt/puppetlabs/puppet/lib/ruby/vendor_gems`.

I will provide a PR.

PR #631, opened 2018-09-19
Add shared gem path for Puppetserver >= 5.3

baurmatt:
Since 5.3.0 Puppet Server can use some gems shipped by puppet-agent. For
this the gem-path settings of puppetserver.conf needs to be extend by
/opt/puppetlabs/puppet/lib/ruby/vendor_gems. This is the default as
shipped by Puppetlabs.

Fixes GH-630.

### DIFF
--- a/manifests/server/puppetserver.pp
+++ b/manifests/server/puppetserver.pp
@@ -140,6 +140,7 @@ class puppet::server::puppetserver (
   $jvm_cmd = strip(join(flatten($jvm_cmd_arr), ' '))
 
   if $::osfamily == 'FreeBSD' {
+    $server_gem_paths = [ '${jruby-puppet.gem-home}', "\"${server_puppetserver_vardir}/vendored-jruby-gems\"", ] # lint:ignore:single_quote_string_with_variables
     augeas { 'puppet::server::puppetserver::jvm':
       context => '/files/etc/rc.conf',
       changes => [ "set puppetserver_java_opts '\"${jvm_cmd}\"'" ],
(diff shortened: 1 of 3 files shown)


baurmatt:
Tests failures aren't related. Seems like master is broken as well.

mmoll:
The hard-coded path will probably break on e.g. FreeBSD, so either a variable for the whole path should get introduced or (if that's possible) it should get deduced from a fact...

mmoll:
actually, there's already `server_puppetserver_vardir`

baurmatt:
Good point! :) But to be honest, I have no idea what the correct path would be for FreeBSD nor do I have a FreeBSD around to take a look. Could it be `/usr/local/share/puppet/lib/ruby/vendor_gems`?

mmoll:
puppetserver6 came into the FreeBSD ports collection and that dir is simply not included:
```diff
-    gem-path: [${jruby-puppet.gem-home}, "/opt/puppetlabs/server/data/puppetserver/vendored-jruby-gems", "/opt/puppetlabs/puppet/lib/ruby/vendor_gems"]
+    gem-path: [${jruby-puppet.gem-home}, "/var/puppet/server/data/puppetserver/vendored-jruby-gems"]
```

mmoll:
Also, I'd rather would use >= 6.0 as version constraint, at least for now.

baurmatt:
Updated the PR so that the `gem-path` setting for FreeBSD doesn't include the path to `vendor_gems`

mmoll:
> FreeBSD: I think that's an error of the package maintainer. 

While I have no information about the brackgorund, the default for FreeBSD's puppetservers is not to include that path, so we should just mirror that.

> 6.0 >=: Can you explain why? The gem-path setting was updated by Puppetlabs with 5.3, this module should reflect that.

Sorry, mixed this up with another PR. 5.3 is correct. :+1:

ekohl:
Thanks!

PR #634, opened 2018-09-25
Fixes #25036 - ensure proxy registration happens before puppet

evgeni:


evgeni:
At least https://github.com/theforeman/puppet-foreman_proxy/pull/456 was not enough in @ehelms and my tests.

And [you can't collect `Class`es](https://puppet.com/docs/puppet/5.5/lang_collectors.html) ;)

ekohl:
Of course life isn't as easy as you'd want to ...

PR #635, opened 2018-09-26
Contain classes

ekohl:
This is the current way of containing classes


ekohl:
You're right. I partially did it but now at least with classes we're consistently not using them. Changing all the variables is another (big) patch so I'm leaving that out here.

mmoll:
@ekohl please rebase :)

ekohl:
Rebased

mmoll:
merged, bedankt @ekohl!

PR #636, opened 2018-09-27
Use facter 3 facts during tests

alexjfisher:
By default, rspec-puppet-facts uses the facterdb facts that match the
version of the facter gem installed.  Starting with facter 3, facter
was no longer shipped as a gem (it was rewritten in C++).

This commit sets `facterversion` to `3.0.0` so that tests run with
more modern (structured) facts stubbed.

Puppet has been shipping facter 3 since puppet 4.2.0.  This module
supports puppet >= 4.6.1, (and puppet 4 is EOL at the end of 2018).

puppet-extlib's `ip_to_cron` function (used in
puppet::agent::service::cron) has recently moved to using the
`networking` structured fact not available in facter 2.

### DIFF
--- a/spec/classes/puppet_agent_service_cron_spec.rb
+++ b/spec/classes/puppet_agent_service_cron_spec.rb
@@ -1,4 +1,5 @@
 require 'spec_helper'
+require 'deep_merge'
 
 describe 'puppet::agent::service::cron' do
   on_os_under_test.each do |os, facts|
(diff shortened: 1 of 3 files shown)


ekohl:
You missed the systemd spec, see https://github.com/theforeman/puppet-puppet/pull/629/files#diff-7795d1e9a0843253185fbd796347dda7

PR #640, opened 2018-10-15
Test for the correct line endings on Windows

ekohl:


mmoll:
`spec/classes/puppet_config_spec.rb` probably also needs a modification.

ekohl:
This is quite ugly but IMHO less ugly than the alternative of including `\r?\n`. Note that `$` also doesn't work because there's a `\r` there. Not sure if there's a better way.

mmoll:
it's :green_heart:

mmoll:
merged, bedankt @ekohl!

PR #642, opened 2018-10-18
Add Puppet 6 support

ekohl:
Generated with https://github.com/theforeman/foreman-installer-modulesync/pull/92

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -1,10 +1,19 @@
 fixtures:
   repositories:
     apache:   'https://github.com/puppetlabs/puppetlabs-apache.git'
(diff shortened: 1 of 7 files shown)


ekohl:
Note to self: we should drop the debian8 beaker tests in favor of debian9. Going to let this continue though to see what's going to happen.

ekohl:
This should now be ready to go. I didn't add Ubuntu acceptance tests because they were failing with a timeout and I haven't had time to investigate. That shouldn't hold back this PR though because it was also timing out on Puppet 5.

mmoll:
merged, bedankt @ekohl!

PR #643, opened 2018-10-23
Update CA related files for Puppetserver 5.3.6

mmoll:


mmoll:
As always, puppetserver got broken with a patch update...

manifests/params.pp
```diff
@@ -449,7 +449,9 @@
   $server_use_legacy_auth_conf      = false
 
   # For Puppetserver, certain configuration parameters are version specific. We assume a particular version here.
-  if versioncmp($::puppetversion, '5.3.0') >= 0 {
+  if versioncmp($::puppetversion, '5.5.7') >= 0 {
```
ekohl: Should this be `5.3.6`?

mmoll: AFAICT no :cry:

mmoll:
@ekohl would you prefer this to go in first, or GH-649 and GH-650?

ekohl:
It's fine to merge this first. Is it still WIP like the title says?

mmoll:
I'll add some tests then it should be ready. I have no idea, if anything in `ca.conf` should be set differently...

mmoll:
This should be ready now, I took the Puppet Labs defaults for the CA settings.

mmoll:
@ekohl gentle ping :point_up:
I even think this could still go into the next 10.x release, at least I think there should be no behavioural change.

ekohl:
Thanks! We just kicked off the pipeline for 1.20.0 but we're planning on a relatively quick 1.20.1 release anyway.

PR #645, opened 2018-10-26
Refactor testing and some puppet classes

ekohl:
With rspec-puppet you're not supposed to test individual classes that should be private. This also makes it more efficient because we had quite a few duplicated cases.

By enforcing $enabled a boolean we can use a simple if/else instead of a case statement. We can also use pick() to remove another case statement.

I'll also get to the server side but it might be easier to remove passenger support first.


mmoll:
merged, bedankt @ekohl!

PR #647, opened 2018-10-30
Add support for managing jetty threads

bastelfreak:
This is based on https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/blob/master/doc/jetty-config.md

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -537,6 +537,32 @@
 # $server_parser::                          Sets the parser to use. Valid options are 'current' or 'future'.
 #                                           Defaults to 'current'.
 #
+# $server_acceptor_threads::                This sets the number of threads that the webserver will dedicate to accepting
+#                                           socket connections for unencrypted HTTP traffic. If not provided, the webserver
+#                                           defaults to the number of virtual cores on the host divided by 8, with a minimum
+#                                           of 1 and maximum of 4.
+#
+# $server_selector_threads::                This sets the number of selectors that the webserver will dedicate to processing
+#                                           events on connected sockets for unencrypted HTTPS traffic. If not provided,
+#                                           the webserver defaults to the minimum of: virtual cores on the host divided by 2
+#                                           or max-threads divided by 16, with a minimum of 1.
+#
+# $server_max_threads::                     This sets the maximum number of threads assigned to responding to HTTP and/or
+#                                           HTTPS requests for a single webserver, effectively changing how many
+#                                           concurrent requests can be made at one time. If not provided, the
+#                                           webserver defaults to 200.
+#
+# $server_ssl_acceptor_threads::            This sets the number of threads that the webserver will dedicate to accepting
+#                                           socket connections for encrypted HTTPS traffic. If not provided, defaults to
+#                                           the number of virtual cores on the host divided by 8, with a minimum of 1 and maximum of 4.
+#
+# $server_ssl_selector_threads::            This sets the number of selectors that the webserver will dedicate to processing
+#                                           events on connected sockets for encrypted HTTPS traffic. Defaults to the number of
+#                                           virtual cores on the host divided by 2, with a minimum of 1 and maximum of 4.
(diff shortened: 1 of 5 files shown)


templates/server/puppetserver/conf.d/webserver.conf.erb
```diff
@@ -27,4 +27,19 @@ webserver: {
         <%= cipher_suite %>,
     <%- end -%>
     ]
+<%- if @server_acceptor_threads -%>
```
ekohl: For them to be in the local scope you need to "import" them in the `puppet::server::puppetserver` class. We also "import" them in `puppet::server`.

mmoll:
@bastelfreak this needs a rebase :)

bastelfreak:
@mmoll @ekohl can you review this please? I think the code is fine, but something is fishy with the acceptance tests on centos 6.

mmoll:
A test with these parameters set would be nice, looks fine visually.

bastelfreak:
I'm not sure yet why the new tests fail. I will have a look again later. Maybe somebody else has an idea.

spec/classes/puppet_server_config_spec.rb
```diff
@@ -714,6 +714,53 @@
         end
       end
 
+      describe 'alow jetty specific server threads' do
```
ekohl: Could you move these tests to `spec/classes/puppet_server_puppetserver_spec.rb` where we have all the tests just for puppetserver?

manifests/init.pp
```diff
@@ -536,6 +536,32 @@
 #
 # $server_parser::                          Sets the parser to use. Valid options are 'current' or 'future'.
 #                                           Defaults to 'current'.
```
mmoll: one comment line missing :point_up:

mmoll:
@bastelfreak one tiny thing, otherwise IMHO ready to go :)

manifests/init.pp
bastelfreak: fixed

spec/classes/puppet_server_puppetserver_spec.rb
```diff
@@ -586,6 +586,53 @@
         let(:params) { super().merge(server_puppetserver_version: '2.1.0') }
         it { should raise_error(Puppet::Error, /puppetserver <2.2 is not supported by this module version/) }
       end
+
+      describe 'allow jetty specific server threads' do
+        context 'with thread config' do
+          let :params do
+            {
+              server:                      true,
```
ekohl: Could you inherit from the parent?
```ruby
let(:params) do
  super().merge(
    server_selector_threads:     1,
    server_acceptor_threads:     2,
    server_ssl_selector_threads: 3,
    server_ssl_acceptor_threads: 4,
    server_max_threads:          5 
  )
end
```

spec/classes/puppet_server_puppetserver_spec.rb
```diff
@@ -586,6 +586,53 @@
...
+            {
+              server:                  true,
+              server_implementation:   'puppetserver',
+              server_ca:               true,
+              server_puppetserver_dir: '/etc/custom/puppetserver',
+              server_jruby_gem_home:   '/opt/puppetlabs/server/data/puppetserver/jruby-gems'
+            }
+          end
+
+          it { is_expected.to compile.with_all_deps }
```
ekohl: Could you move this to the `with default parameters` block (starts at line 29)? Perhaps merge it in the existing check for this file at line 73.

spec/classes/puppet_server_puppetserver_spec.rb
bastelfreak: done

spec/classes/puppet_server_puppetserver_spec.rb
bastelfreak: done

ekohl:
Thanks!

PR #648, opened 2018-11-01
Refactor puppetserver spec testing

ekohl:
This is in preparation of dropping support for puppetserver < 2.7

### DIFF
--- a/spec/classes/puppet_server_puppetserver_spec.rb
+++ b/spec/classes/puppet_server_puppetserver_spec.rb
@@ -1,424 +1,245 @@
 require 'spec_helper'
 
-describe 'puppet::server::puppetserver' do
+describe 'puppet' do
   on_os_under_test.each do |os, facts|
-    next if facts[:osfamily] == 'windows'
-    next if facts[:osfamily] == 'Archlinux'
+    next if unsupported_puppetmaster_osfamily(facts[:osfamily])
     context "on #{os}" do
-      let :pre_condition do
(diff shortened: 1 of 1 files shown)


ekohl:
@mmoll the centos 6 acceptance test has become unstable but I think that doesn't block this PR.

spec/classes/puppet_server_puppetserver_spec.rb
```diff
@@ -1,903 +1,589 @@
...
+        }
+      end
 
       describe 'with default parameters' do
-        let(:params) do
-          default_params.merge(:server_puppetserver_dir => '/etc/custom/puppetserver')
-        end
         it { should contain_file('/etc/custom/puppetserver/bootstrap.cfg') }
-        it { should contain_file_line('ca_enabled').with_ensure('present') }
-        it { should contain_file_line('ca_disabled'). with_ensure('absent') }
```
mmoll: What's the reason for removing these two ensures?

mmoll:
on small question, :+1: otherwise

spec/classes/puppet_server_puppetserver_spec.rb
ekohl: Oh, that was my bad. Initially I changed this to `should_not` because I removed the version pinning as well but then you test with the latest version which doesn't have them. Will add them again.

mmoll:
merged, bedankt @ekohl!

PR #650, opened 2018-11-02
Remove passenger support

ekohl:


ekohl:
When the tests pass I think this is also ready to go.

ekohl:
I've split off https://github.com/theforeman/puppet-puppet/pull/652 so we can at least get that in.

mmoll:
merged, bedankt @ekohl!

PR #651, opened 2018-11-03
adjust puppetversion to puppetserver_version mapping

mmoll:


mmoll:
Both of thoese versions got relased at 2018-03-20

PR #652, opened 2018-11-16
Refactor puppet server spec testing

ekohl:


spec/classes/puppet_server_spec.rb
```diff
@@ -1,181 +1,754 @@
...
+describe 'puppet' do
   on_os_under_test.each do |os, facts|
-    next if unsupported_puppetmaster_osfamily(facts[:osfamily])
-    context "on #{os}" do
-      if facts[:osfamily] == 'FreeBSD'
-        ssldir = '/var/puppet/ssl'
-      else
-        ssldir = '/etc/puppetlabs/puppet/ssl'
+    context "on #{os}", unless: unsupported_puppetmaster_osfamily(facts[:osfamily]) do
+      # TODO: only used in 1 test so not before each?
```
mmoll: still TODO? ;)

spec/classes/puppet_server_spec.rb
```diff
@@ -1,181 +1,754 @@
...
+        it 'should set up the ENC' do
+          should contain_class('foreman::puppetmaster')
+            .with_foreman_url('https://foo.example.com')
+            .with_receive_facts(true)
+            .with_puppet_home(puppetserver_vardir)
+            .with_puppet_etcdir(etcdir)
+            .with_timeout(60)
+          # Since this is managed inside the foreman module it does not
+          # make sense to test it here
+          # with_puppet_basedir('/usr/lib/ruby/site_ruby/1.9/puppet').
```
mmoll: leftover?

spec/classes/puppet_server_spec.rb
ekohl: It was, but moved it to a `before :context` in that particular case.

spec/classes/puppet_server_spec.rb
ekohl: Updated to verify it passes in the correct dir.

PR #653, opened 2018-11-16
Remove dynamic and config environments

ekohl:
Since Puppet 4 only directory environments are supported so this was mostly dead code.


ekohl:
Rephrased. I did consider keeping the env defined type but chose not to in favor of recommending people to store these files in git and deploy either with our git repo (should still work I think) or tools like r10k.

mmoll:
@ekohl needs a rebase and fixing

ekohl:
Rebased

mmoll:
merged, bedankt @ekohl!

PR #654, opened 2018-11-18
Drop Debian 8 support

ekohl:


mmoll:
I'm unsure if I really like this change.

While the "normal" support for Debian/jessie ended 06/2018, it still has semi-official LTS until 06/2020.

Is there a near-future change that's hard to implement with Debian/jessie?

ekohl:
I was debating this. One reason that triggered it was that there are only PC1 packages available. However, I noticed Debian 9 has no PC1 packages and our acceptance tests don't cover Puppet 5 yet so I'm closing this for now.

PR #658, opened 2018-12-25
Modulesync: refactor acceptance tests + gem updates

ekohl:


spec/support/aio.rb
```diff
@@ -0,0 +1,9 @@
+aio = on_supported_os.reject do |os, facts|
+  ['FreeBSD', 'DragonFly', 'Windows'].include?(facts[:operatingsystem])
```
mmoll: Archlinux should get added here, see 75d0369

ekohl: Debatable. It's not really AIO because it uses the system ruby and OS packages, but follows some paths from the AIO packaging. cc @bastelfreak

mmoll: Our params.pp says `$aio_package = ($::osfamily == 'Windows' or $::rubysitedir =~ /\/opt\/puppetlabs\/puppet/)`, following that logic it's not. ;)
`

ekohl: Right, this was reject, not filter :)

bastelfreak: hey,
we build Puppet against system ruby 2.5.1 at the moment, so `$fact['rubysitedir']` resolves to `/usr/lib/ruby/site_ruby/2.5.0`. Besides that we try to follow the paths in the AIO packages as good as possible.

ekohl: Updated. I think it should now cover Arch. I do wonder: who owns `/etc/puppetlabs/puppet` in packaging? In AIO it's root (which I agree with) but our current module sets it to puppet.

Issue #383, opened 2016-04-01
Additional case for managing puppet-agent service


szemlyanoy:
Hey all,

Seems in current state puppet-agent module can be either controlled or turned off entirely. 
My case: I want to ignore any control of service resource in case I apply catalog from any non-production envrironment, because I sometimes stop puppet-agent to test changes from my environment and want to keep them for some while there.
Of course I can do like:
`case $::environment {
    'production': {class { 'puppet': runmode => service, }}
    default: { class {'puppet': runmode => none, }}
  }`

but this will end up with service stop/disabling(!) literally every time while it's more convenient to control it manually.

My proposal: add new parameter $puppet::agent_service_control with default 'true'  
Thoughts?

Sergey

ekohl:
I'd recommend ```puppet agent --disable``` which means you can keep the service running and it'll do as you expect it to.

szemlyanoy:
this trick just disables any puppet runs from any environment. Lock file just does it's magic

ekohl:
Then you want that variable to control whether ```puppet::agent::service``` is included at all?

szemlyanoy:
Exactly, it might be handy for some cases

szemlyanoy:
if no one mind I can proceed with it...?

mmoll:
@szemlyanoy what's the status here?

ekohl:
@szemlyanoy Any update on this?

ekohl:
Closing for inactivity. A PR is still welcome.

Rathios:
I ran into a case like this too. I wanted to try using this module to manage my agents, but I'd also like to be able to leave the service unstarted after Puppet runs so I can test undisturbed while I do --noop runs etc.

Sent a PR for review. I'm not proficient at rspec tests, so please let me know if something doesn't look right.

PR #659, opened 2018-12-25
Add unmanaged Puppet Agent runmode

Rathios:
Fixes GH-383

### DIFF
--- a/manifests/agent/service/cron.pp
+++ b/manifests/agent/service/cron.pp
@@ -1,7 +1,7 @@
 class puppet::agent::service::cron (
   Boolean $enabled = false,
 ) {
-  unless 'unmanaged' in $::puppet::runmode or 'cron' in $::puppet::unavailable_runmodes {
+  unless $::puppet::runmode == 'unmanaged' or 'cron' in $::puppet::unavailable_runmodes {
     if $enabled {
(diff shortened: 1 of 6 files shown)


ekohl:
How does this differ from runmode `none`?

Rathios:
Left a comment here too: https://github.com/theforeman/puppet-puppet/issues/383

Bottom line is I would like to be able to leave the service stopped while testing changes, undisturbed by the service. In that context I would also like to be able to enable the service without starting it, which I can do with a profile in the control-repository, but it's not currently possible to manually manage the service yourself as long as this module is used to install the agent.

manifests/agent/service/cron.pp
```diff
@@ -1,7 +1,7 @@
 class puppet::agent::service::cron (
   Boolean $enabled = false,
 ) {
-  unless 'cron' in $::puppet::unavailable_runmodes {
+  unless 'unmanaged' in $::puppet::runmode or 'cron' in $::puppet::unavailable_runmodes {
```
ekohl: runmode is a string, so you can compare it with `$::puppet::runmode == 'unmanaged'`

mmoll:
I'm OK with it after fixing :+1:

Rathios:
Ha, you're right. The consistency OCD missed the fact that $::puppet::unavailable_runmodes is NOT a string. Things make sense now. Fixed.

ekohl:
Thanks!

PR #664, opened 2018-12-28
Config file updates and FreeBSD refactorings

mmoll:


mmoll:
The package naming is not 100% consistent, so server and agent need to get handled differently:
```
puppet4-4.10.12                Configuration management framework written in Ruby
puppet5-5.5.8                  Configuration management framework written in Ruby
puppet6-6.1.0                  Configuration management framework written in Ruby

puppetserver-2.8.0_1           Puppet Server running in the JVM
puppetserver5-5.3.6            Puppet Server running in the JVM
puppetserver6-6.1.0            Puppet Server running in the JVM
```

manifests/params.pp
```diff
@@ -272,9 +272,18 @@
   # - puppetdb for puppetdb
   $server_storeconfigs_backend = undef
 
-  $server_ssl_dir  = $ssldir
-  $server_package     = undef
-  $server_version     = undef
+  if ($::osfamily =~ /(FreeBSD|DragonFly)/) {
```
ekohl: I'd leave this `if` off. It doesn't hurt to always define `$puppet_major`.

mmoll: done.

PR #665, opened 2018-12-28
Remove $deb_naio_package & puppet-common install

ekohl:
* 7a1b62e Remove $deb_naio_package variable
* d22a852 Remove Debian package special case
This was an ancient workaround where puppet didn't properly depend on puppet-common. On Debian Jessie puppet already depends on puppet-common and everything newer has made puppet-common a dummy transitional package. This means we can remove the special case and fall back to the general else case.


mmoll:
merged, bedankt @ekohl!

PR #671, opened 2019-01-16
Implement $server_max_open_files

### LINKED ISSUE #670: Make max open files configurable for puppetserver
Larger setups tend to use more file descriptors as the default configuration of the OS.

Puppet itself suggest to increase them if needed --> https://puppet.com/docs/pe/2017.3/config_ulimit.html

As we currently do not manage the init script (and I definitely like to keep it that way) the best way is probably to just use the `$config` file for it.

Will send a PR soon :)

baurmatt:
This commit implements the $server_max_open_files parameter which allows
the user to configure the max open file descriptor limit for
Puppetserver.

Fixes #670.

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -13,6 +13,7 @@ fixtures:
     inifile:  'https://github.com/puppetlabs/puppetlabs-inifile.git'
     puppetdb: 'https://github.com/puppetlabs/puppetlabs-puppetdb.git'
     stdlib:   'https://github.com/puppetlabs/puppetlabs-stdlib.git'
+    systemd:  'https://github.com/camptocamp/puppet-systemd.git'
     yumrepo_core:
       repo: 'https://github.com/puppetlabs/puppetlabs-yumrepo_core'
       puppet_version: '>= 6.0.0'
--- a/manifests/server/puppetserver.pp
+++ b/manifests/server/puppetserver.pp
@@ -126,6 +126,7 @@ class puppet::server::puppetserver (
   $ca_allow_sans                          = $::puppet::server::ca_allow_sans,
   $ca_allow_auth_extensions               = $::puppet::server::ca_allow_auth_extensions,
   $ca_enable_infra_crl                    = $::puppet::server::ca_enable_infra_crl,
+  $max_open_files                         = $::puppet::server::max_open_files,
 ) {
   include ::puppet::server
 
@@ -190,6 +191,26 @@ class puppet::server::puppetserver (
         incl    => $config,
         context => "/files${config}",
         changes => $jruby_jar_changes,
+      }
+    }
+
+    $ensure_max_open_files = $max_open_files ? {
+      undef   => 'absent',
+      default => 'present',
+    }
+    if $facts['service_provider'] == 'systemd' {
+      systemd::dropin_file { 'puppetserver.service-limits.conf':
+        ensure   => $ensure_max_open_files,
+        filename => 'limits.conf',
+        unit     => 'puppetserver.service',
+        content  => "[Service]\nLimitNOFILE=${max_open_files}\n",
+      }
+    } else {
+      file_line { 'puppet::server::puppetserver::max_open_files':
+        ensure => $ensure_max_open_files,
+        path   => $config,
+        line   => "ulimit -n ${max_open_files}",
+        match  => '^ulimit\ -n',
       }
     }
   }
--- a/spec/acceptance/puppetserver_config_spec.rb
+++ b/spec/acceptance/puppetserver_config_spec.rb
@@ -1,0 +1,46 @@
+require 'spec_helper_acceptance'
+
+describe 'Puppetserver config options', unless: ENV['BEAKER_PUPPET_COLLECTION'] == 'pc1' && fact('lsbdistcodename') == 'stretch' do
+  before(:context) do
+    if fact('lsbdistcodename') == 'jessie' && ENV['BEAKER_PUPPET_COLLECTION'] != 'pc1'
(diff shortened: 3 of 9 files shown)


manifests/server/puppetserver.pp
```diff
@@ -192,6 +193,26 @@
         changes => $jruby_jar_changes,
       }
     }
+
+    $ensure_max_open_files = $max_open_files ? {
+      undef   => 'absent',
+      default => 'present',
+    }
+    if $::service_provider == 'systemd' {
```
alexjfisher: These days, is it better to use `$facts['service_provider']` rather than the topscope variable?

manifests/server/puppetserver.pp
```diff
@@ -192,6 +193,26 @@
...
+      systemd::dropin_file { 'puppetserver.service-limits.conf':
+        ensure   => $ensure_max_open_files,
+        filename => 'limits.conf',
+        unit     => 'puppetserver.service',
+        content  => "[Service]\nLimitNOFILE=${max_open_files}",
```
alexjfisher: I think the last line of a file should also end with a newline character.  But I guess here it doesn't really matter.

spec/spec_helper.rb
```diff
@@ -13,6 +13,7 @@
                                                                              # Original fact sources:
 add_custom_fact :puppet_environmentpath, '/etc/puppetlabs/code/environments' # puppetlabs-stdlib
 add_custom_fact :root_home, '/root'                                          # puppetlabs-stdlib
+add_custom_fact :service_provider, 'redhat'                                  # puppetlabs-stdlib
```
ekohl: I'd say `systemd` is a more common default if I look at `metadata.json` but if you really want a correct implementation (from https://github.com/voxpupuli/modulesync_config/pull/431):

```ruby
# Rough conversion of grepping in the puppet source:
# grep defaultfor lib/puppet/provider/service/*.rb
add_custom_fact :service_provider, ->(os, facts) do
  case facts[:osfamily].downcase
  when 'archlinux'
    'systemd'
  when 'darwin'
    'launchd'
  when 'debian'
    if facts[:operatingsystem] == 'Ubuntu'
      facts[:operatingsystemmajrelease].to_i >= 15 ? 'systemd' : 'upstart'
    elsif facts[:operatingsystem] == 'Debian' && facts[:operatingsystemmajrelease].to_i >= 8
      'systemd'
    else
      'debian'
    end
  when 'freebsd'
    'freebsd'
  when 'gentoo'
    'openrc'
  when 'openbsd'
    'openbsd'
  when 'redhat'
    facts[:operatingsystemrelease].to_i >= 7 ? 'systemd' : 'redhat'
  when 'suse'
    facts[:operatingsystemmajrelease].to_i >= 12 ? 'systemd' : 'redhat'
  when 'windows'
    'windows'
  else
    'init'
  end
end
```

.fixtures.yml
```diff
@@ -13,6 +13,7 @@ fixtures:
     inifile:  'https://github.com/puppetlabs/puppetlabs-inifile.git'
     puppetdb: 'https://github.com/puppetlabs/puppetlabs-puppetdb.git'
     stdlib:   'https://github.com/puppetlabs/puppetlabs-stdlib.git'
+    systemd:  'git://github.com/camptocamp/puppet-systemd.git'
```
ekohl: Could you use https:// here?

manifests/server/puppetserver.pp
baurmatt: Changed, thanks! :)

manifests/server/puppetserver.pp
baurmatt: Changed, thanks! :)

spec/spec_helper.rb
baurmatt: Oh wow, this is amazing! Switched to this way more complete implementation.

spec/classes/puppet_server_puppetserver_spec.rb
```diff
@@ -404,6 +404,40 @@
...
+            it do
+              should contain_systemd__dropin_file('puppetserver.service-limits.conf')
+                .with_ensure('present')
+                .with_filename('limits.conf')
+                .with_unit('puppetserver.service')
+                .with_content("[Service]\nLimitNOFILE=32143")
+            end
+          end
+
+          context 'on non systemd based systems' do
```
ekohl: This will now probably fail for some distros

baurmatt: Yeah :( Fixed it! :)

baurmatt:
> Could you add an acceptance test as well?

Done! :)

> Since you're adding systemd as a soft dependency, could you describe this in README.md?

Not really sure where to add it. Can you point me to a good place for it? :)

spec/acceptance/puppetserver_config_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+        server_jvm_min_heap_size => '256m',
+        server_max_open_files    => 32143,
+      }
+      MANIFEST
+    end
+
+    it_behaves_like 'a idempotent resource'
+
+    # pgrep -f java.*puppetserver would be better. But i cannot get it to work. Shellwords.escape() seems to break something
+    describe command("grep '^Max open files' /proc/`cat /var/run/puppetlabs/puppetserver/puppetserver.pid`/limits"), :sudo => true do
```
ekohl: You may be able to parse this via `systemctl status puppetserver` as well. Both the main pid as the limit. On the other hand, this is a real test on the result so I'm fine with this.

ekohl:
Thanks!

PR #682, opened 2019-02-28
Fixes #26199 - Fix dependency cycle when using git_repo

ekohl:
Previously this resulted in a dependency cycle. The tests didn't compile so didn't catch it.

    dependency cycles found: (File[/opt/puppetlabs/puppet/cache] => Class[Puppet::Server::Install] => Class[Puppet::Server::Config] => File[/opt/puppetlabs/puppet] => File[/opt/puppetlabs/puppet/cache])

a61e010e8d24f2e44ab1ebba0cc4382b05ce9d37 introduced this by managing $sharedir in puppet::server::config. $vardir was managed in puppet::server::install. Since sharedir is actually /opt/puppetlabs/puppet and vardir is the cache directory within that, it created a cycle.

### DIFF
--- a/manifests/server/config.pp
+++ b/manifests/server/config.pp
@@ -232,10 +232,17 @@ class puppet::server::config inherits puppet::config {
   }
 
   if $::puppet::server::git_repo {
-    # need to chown the $vardir before puppet does it, or else
-    # we can't write puppet.git/ on the first run
-
     include ::git
+
+    if $::puppet::server::manage_user {
+      Class['git'] -> User[$::puppet::server::user]
+    }
+
+    file { $::puppet::vardir:
+      ensure => directory,
+      owner  => 'root',
+      group  => 'root',
+    }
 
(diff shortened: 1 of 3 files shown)


manifests/server/config.pp
```diff
@@ -232,18 +232,25 @@
...
-
     include ::git
 
+    if $::puppet::server::manage_user {
+      Class['git'] -> User[$::puppet::server::user]
+    }
+
+    file { $::puppet::vardir:
+      ensure => directory,
+      owner  => 'root',
```
ekohl: Note I am changing this. In the old code it was owned by puppet:puppet but now it's root:root. This matches with the puppet-agent package (at least 5.5.10).

mmoll:
tests fail :broken_heart:

ekohl:
:green_heart:

mmoll:
merged, bedankt @ekohl!

PR #683, opened 2019-03-01
Handle the pluginsync setting deprecation

ekohl:
In Puppet 4 the pluginsync setting was deprecated. At least in Puppet 5 it generates a deprecation warning and in Puppet 6 it was removed.  Because the default was already true, we only emit it if it's non-default to avoid the deprecation warning. In Puppet 6 we can detect an invalid setting so we hard fail.

### DIFF
--- a/manifests/agent/config.pp
+++ b/manifests/agent/config.pp
@@ -6,7 +6,6 @@ class puppet::agent::config inherits puppet::config {
     'localconfig':       value => '$vardir/localconfig';
     'default_schedules': value => false;
     'report':            value => $::puppet::report;
-    'pluginsync':        value => $::puppet::pluginsync;
     'masterport':        value => $::puppet::port;
     'environment':       value => $::puppet::environment;
     'listen':            value => $::puppet::listen;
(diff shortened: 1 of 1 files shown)


mmoll:
I'm fine with this as-is but asking myself if it would be good to dump a `warn` on Puppet 5... or is that the same kind of visibility as the one generated by Puppet itself?

ekohl:
The reason I noticed this was because I saw this in my logs:

```
Mar  1 14:02:40 foreman puppet-agent[18582]: Config file /etc/puppetlabs/puppet/puppet.conf changed; triggering re-parse of all config files.
Mar  1 14:02:40 foreman puppet-agent[18582]: Setting 'pluginsync' is deprecated.
Mar  1 14:02:40 foreman puppet-agent[18582]:   (location: /opt/puppetlabs/puppet/lib/ruby/vendor_ruby/puppet/defaults.rb:1879:in `block in <module:Puppet>')
Mar  1 14:02:55 foreman puppet-agent[18582]: Config file /etc/puppetlabs/puppet/puppet.conf changed; triggering re-parse of all config files.
```

mmoll:
@ekohl So this is only visible in the logfile? Or also in the Puppet Report in Foreman (I tihnk a puppet `warn` would show up there and also change host status)?

ekohl:
That's from syslog, not the puppet report. I can see about that.

ekohl:
@mmoll do you think including this in a patch release is good enough? In the next major version drop it.

mmoll:
@ekohl sure, let's do this... We'll drop Ubuntu/trusty and Puppet 4.x in the next round of module releases anyway, I presume...

mmoll:
merged, bedankt @ekohl!

PR #684, opened 2019-03-16
drop EOL OSes

mmoll:


manifests/params.pp
```diff
@@ -301,7 +301,7 @@
   case $::osfamily {
     'Debian' : {
       $agent_restart_command = "/usr/sbin/service ${service_name} reload"
-      if ($::operatingsystem == 'Debian' or $::operatingsystem == 'Ubuntu' and versioncmp($::operatingsystemrelease, '15.04') >= 0) {
+      if ($::operatingsystem == 'Debian' or $::operatingsystem == 'Ubuntu') {
```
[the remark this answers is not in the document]
mmoll: true.

spec/spec_helper.rb
```diff
@@ -22,13 +22,7 @@
   when 'darwin'
     'launchd'
   when 'debian'
-    if facts[:operatingsystem] == 'Ubuntu'
```
ekohl: I'm slightly tempted to keep this in since it matches the Puppet built in values and externalize this somewhere, but probably fine to drop it here.

mmoll: we need to handle this in msync later anyway.

PR #686, opened 2019-03-19
add 2018 to osreleasemajor for agent_restart_command 

cabrinha:
Amazon Linux now has a 2018 release date on their AMIs:

```
# cat /etc/issue

Amazon Linux AMI release 2018.03
Kernel \r on an \m

# facter -p os
{
  architecture => "x86_64",
  family => "RedHat",
  hardware => "x86_64",
  name => "Amazon",
  release => {
    full => "2018.03",
    major => "2018",
    minor => "03"
  },
  selinux => {
    enabled => false
  }
}
```

So, when not specifying $agent_restart_command, we get this:

```
Error: /Stage[main]/Puppet::Agent::Service::Systemd/Service[puppet-run.timer]: Provider systemd is not functional on this host
```

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -307,7 +307,7 @@ class puppet::params {
       # PSBM is a CentOS 6 based distribution
       # it reports its $osreleasemajor as 2, not 6.
       # thats why we're matching for '2' in both parts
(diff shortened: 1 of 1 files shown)


ekohl:
Thanks!

PR #688, opened 2019-03-25
remove Debian 8 support

mmoll:


mmoll:
As it seems the Debian project removed some stuff regarding jessie, especially the removal of the -backports repo does hurt us, so remove it alltogether, regardless of its unofficial LTS.

PR #689, opened 2019-03-26
add Puppetserver 6.3 auth.conf update

mmoll:


mmoll:
It would be super nice if GH-685 could get in first to avoid rebase hell when adding a test here.

mmoll:
rebased and added a test, this should be ready now

PR #690, opened 2019-03-29
Configurable minute and hour for cronjob and systemd.timer

mauricemeyer:
This PR enables you to configure the minute and hour at which your Puppet Cronjob/systemd timer runs.

This is useful for advanced scenarios where you want to limit when the Puppet agent runs.

The change is fully backwards compatible as it does not change any defaults


mmoll:
merged, danke @mauricemeyer!

PR #691, opened 2019-04-09
Adding ability to choose between plain text or json format log file

bt-lemery:
Adds ability to specify whether puppetserver should output log file in plain text format, or json for easier injestion in to ELK via filebeat.


manifests/params.pp
```diff
@@ -58,95 +58,100 @@
...
+      $ssldir                        = "${dir_prefix}/etc/ssl"
+      $vardir                        = "${dir_prefix}/var"
+      $sharedir                      = "${dir_prefix}/share"
+      $bindir                        = "${dir_prefix}/bin"
+      $root_group                    = undef
+      $server_puppetserver_dir       = undef
+      $server_puppetserver_vardir    = undef
+      $server_puppetserver_rundir    = undef
+      $server_puppetserver_logdir    = undef
+      $server_puppetserver_logformat = undef
```
ekohl: Any reason not to default set it unconditionally to plain? You're skipping this on Windows anyway.

manifests/server/puppetserver.pp
```diff
@@ -233,6 +234,13 @@
     }
   }
 
+  unless $::osfamily == 'Windows' {
```
ekohl: What's the reasoning to skip it on Windows? AFAIK we don't support the server on Windows anyway.

manifests/params.pp
bt-lemery: Sure, will do.

manifests/server/puppetserver.pp
bt-lemery: I don't support it on Windows, either :)  I'll remove the 'unless'.

manifests/server/puppetserver.pp
```diff
@@ -227,6 +228,17 @@
...
+  file { "${server_puppetserver_dir}/logback.xml":
+    ensure  => file,
+    content => template('puppet/server/puppetserver/logback.xml.erb'),
+  }
+
+  if versioncmp($server_puppetserver_version, '5.3.6') >= 0 {
+    $ca_conf_ensure = present
+  } else {
+    $ca_conf_ensure = absent
+  }
```
mmoll: I guess the rebase/merge went wrong here...

manifests/server.pp
```diff
@@ -355,6 +357,7 @@
   Stdlib::Absolutepath $puppetserver_vardir = $::puppet::server_puppetserver_vardir,
   Optional[Stdlib::Absolutepath] $puppetserver_rundir = $::puppet::server_puppetserver_rundir,
   Optional[Stdlib::Absolutepath] $puppetserver_logdir = $::puppet::server_puppetserver_logdir,
+  Optional[String] $server_puppetserver_logformat = $puppet::server_puppetserver_logformat,
```
ekohl:
```suggestion
  Enum['plain', 'json'] $server_puppetserver_logformat = $puppet::server_puppetserver_logformat,
```

manifests/init.pp
```diff
@@ -375,6 +375,9 @@
 #
 # $server_puppetserver_logdir::             The path of the puppetserver log dir
 #
+# $server_puppetserver_logformat::          Format for puppetserver log files, either plain text or json.  Defaults
```
ekohl: This parameter also needs to be added in this file. You can also use the data type `Enum['plain', 'text']` and leave it out of the docs. We also don't include the defaults.

manifests/server/puppetserver.pp
ekohl: Could you have a look here? This still needs to be removed

templates/server/puppetserver/logback.xml.erb
```diff
@@ -0,0 +1,76 @@
...
+    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+        <encoder>
+            <pattern>%d %-5p [%t] [%c{2}] %m%n</pattern>
+        </encoder>
+    </appender>
+
+<% if @server_puppetserver_logformat == 'plain' -%>
+    <appender name="F1" class="ch.qos.logback.core.rolling.RollingFileAppender">
+        <!-- TODO: this path should not be hard-coded -->
+        <file>/var/log/puppetlabs/puppetserver/puppetserver.log</file>
```
ekohl: We have a variable for the logdir. Can you check if the following works?

```suggestion
        <file><%= @server_puppetserver_logdir %>/puppetserver.log</file>
```

templates/server/puppetserver/logback.xml.erb
```diff
@@ -0,0 +1,76 @@
...
+            <maxHistory>90</maxHistory>
+            <totalSizeCap>1GB</totalSizeCap>
+        </rollingPolicy>
+        <encoder>
+            <pattern>%d %-5p [%t] [%c{2}] %m%n</pattern>
+        </encoder>
+    </appender>
+<% end -%>
+
+<% if @server_puppetserver_logformat == 'json' -%>
```
ekohl: If the intention is to support only one format, you can change it to `elsif`. The code suggests you want to support both so in that case you may need to change the data type to allow both.

ekohl:
Closing for inactivity. If you're still interested in this, please submit a new PR.

PR #693, opened 2019-04-16
Use current timeout management variables

ekohl:
`configtimeout` was deprecated since Puppet 4 and replaced with `http_connection_timeout` and `http_read_timeout`.


mmoll:
merged, bedankt @ekohl!

PR #694, opened 2019-04-17
Rely on the service_provider fact

ekohl:
This avoids hardcoding the service provider and greatly simplifies the code.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -288,37 +288,30 @@ class puppet::params {
   $systemd_unit_name = 'puppet-run'
   # Mechanisms to manage and reload/restart the agent
   # If supported on the OS, reloading is prefered since it does not kill a currently active puppet run
-  case $facts['os']['family'] {
-    'Debian' : {
-      $agent_restart_command = "/usr/sbin/service ${service_name} reload"
-      $unavailable_runmodes = []
+  if $facts['service_provider'] == 'systemd' {
+    $agent_restart_command = "/bin/systemctl reload-or-restart ${service_name}"
+    $unavailable_runmodes = $facts['os']['family'] ? {
+      'Archlinux' => ['cron'],
+      default     => [],
     }
-    'Redhat' : {
-      # PSBM is a CentOS 6 based distribution
-      # it reports its $osreleasemajor as 2, not 6.
-      # thats why we're matching for '2' in both parts
-      # Amazon Linux is like RHEL6 but reports its osreleasemajor as 2017 or 2018.
-      $agent_restart_command = $facts['os']['release']['major'] ? {
-        /^(2|5|6|2017|2018)$/ => "/sbin/service ${service_name} reload",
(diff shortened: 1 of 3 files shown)


mmoll:
@ekohl tests fail :broken_heart:

mmoll:
is this a 12.0.0 candidate?

ekohl:
No, it's a branch I had locally that I thought I should finish at some point. I'd rather get 12.0.0 out now and merge this later.

mmoll:
what's the status here?

ekohl:
Rebased and updated the tests

mmoll:
test failures related?

ekohl:
I'm not entirely sure how to explain the test failures. It's also annoying we don't have EL7 tests now so that should be fixed first by using an older image.

ekohl:
So that does appear to be the problem:
```
2021-03-29T15:04:07.1177819Z debian9-64.example.com 15:04:07$ facter --json "service_provider"
2021-03-29T15:04:07.2509850Z   {
2021-03-29T15:04:07.2510867Z     "service_provider": ""
2021-03-29T15:04:07.2517137Z   }
```

ekohl:
Thinking more about this, it may well be that `facter` doesn't include the custom Puppet facts.

ekohl:
After such a long time I finally found out the problem and it was really stupid: I used `/usr/bin/systemctl` but on Debian it's `/bin/systemctl`. Somehow Puppet didn't raise an error about the command not being found and it fell back to a real restart instead of a reload. Let's see if CI agrees.

ekohl:
This finally passes. The other failures are unrelated.

manifests/params.pp
```diff
@@ -288,37 +288,30 @@
   $systemd_unit_name = 'puppet-run'
   # Mechanisms to manage and reload/restart the agent
   # If supported on the OS, reloading is prefered since it does not kill a currently active puppet run
-  case $facts['os']['family'] {
-    'Debian' : {
-      $agent_restart_command = "/usr/sbin/service ${service_name} reload"
-      $unavailable_runmodes = []
+  if $facts['service_provider'] == 'systemd' {
+    $agent_restart_command = "/bin/systemctl reload-or-restart ${service_name}"
```
wbclark, line 292: Nice, TIL

wbclark (commented):
What if any is the relationship between these two commits? Is the change to `server_jvm_max_heap_size` better suited to a separate PR?

ekohl:
It's something I found while testing this manually. Technically it could be a separate PR, but I think there's little value in that. It is only for the acceptance tests so it's not like users need to know about it via the changelog. Combining them saves CI resources.

wbclark:
> It's something I found while testing this manually. Technically it could be a separate PR, but I think there's little value in that. It is only for the acceptance tests so it's not like users need to know about it via the changelog. Combining them saves CI resources.

Ah, got it, TY

wbclark (approved):

PR #702, opened 2019-06-19
Support for 'versioned code' in the puppet server configuration

DesideriusE:


mmoll:
continuation in GH-726

PR #703, opened 2019-06-19
Revert/systemd dropin

DesideriusE:


ekohl:
We chose to use this because it also correctly handles the `systemd daemon-reload`. If you think we should revert this, it needs some argumentation but for now I'm going to close this.

DesideriusE:
I understand why you'd want to use a module that takes care of such implementation details for you.

Rather, my point is that the inclusion of this particular module lays claim on the '::systemd' module namespace in puppet. This creates a conflict at sites where the ::systemd namespace is already occupied, thus preventing the use of (or update to) this new version of the 'theforeman/puppet' module.

Admittedly, this is not the 'theforeman' module's fault, but rather a shortcoming in puppet in which there's currently no concept of a 'private module' namespace that would let one safely use modules as libraries within the containing module only.

As it stands however, module writers should be weary to use any non-standard 'library modules', especially if a module is to be positioned as general-purpose functionality, potentially useful for a large audience (as I think holds true for the 'theforeman/puppet' module), because its usefulness can be severely impaired by module namespace conflicts.


Best regards,

--Paul kranenburg

________________________________

PR #707, opened 2019-10-04
Ensure config file ends with a new line

baurmatt:
The puppet.conf file doesn't end with a newline which makes it hard to `cat`. This ensure that the file will end with a newline.

### DIFF
--- a/manifests/config/entry.pp
+++ b/manifests/config/entry.pp
@@ -30,7 +30,7 @@ define puppet::config::entry (
   # they make sure that '1_main ' is ordered before '1_main_*'
   ensure_resource('concat::fragment', "puppet.conf_${section}", {
       target  => "${::puppet::dir}/puppet.conf",
-      content => "\n\n[${section}]",
+      content => "\n[${section}]",
       order   => "${sectionorder}_${section} ",
   })
 
@@ -39,7 +39,7 @@ define puppet::config::entry (
   if (!defined(Concat::Fragment["puppet.conf_${section}_${key}"])){
     concat::fragment{"puppet.conf_${section}_${key}":
       target  => "${::puppet::dir}/puppet.conf",
-      content => "\n    ${key} = ${_value}",
+      content => "    ${key} = ${_value}",
       order   => "${sectionorder}_${section}_${key} ",
     }
   } else {
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -82,15 +82,17 @@ class puppet::config(
   -> case $::osfamily {
     'Windows': {
       concat { "${puppet_dir}/puppet.conf":
-        mode  => '0674',
+        mode           => '0674',
+        ensure_newline => true,
       }
     }
 
     default: {
       concat { "${puppet_dir}/puppet.conf":
-        owner => 'root',
-        group => $::puppet::params::root_group,
-        mode  => '0644',
+        owner          => 'root',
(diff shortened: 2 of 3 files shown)


ekohl:
Last time I tried this, it didn't do what I wanted. From the reference:
> ensure_newline
> Specifies whether to add a line break at the end of each fragment that doesn't already end in one.

We also add newlines in places and this resulted in a hard to read file. Is this still the case?

baurmatt:
Ha, right! :) `puppet::config::entry` adds a newline at the beginning of the fragment and the `ensure_newline` parameter function only checks (and adds) newlines at the end. I've fixed it and the diff now looks good:

```
Notice: /Stage[main]/Puppet::Config/Concat[/etc/puppetlabs/puppet/puppet.conf]/File[/etc/puppetlabs/puppet/puppet.conf]/content:
--- /etc/puppetlabs/puppet/puppet.conf	2019-04-23 09:19:27.831205678 +0200
+++ /tmp/puppet-file20191024-11597-17womp5	2019-10-24 09:48:23.947771487 +0200
@@ -1,5 +1,4 @@

-
 [main]
     certname = test1234
     codedir = /etc/puppetlabs/code
@@ -27,4 +26,4 @@
     runinterval = 1800
     splay = false
     splaylimit = 1800
-    usecacheonfailure = true
\ No newline at end of file
+    usecacheonfailure = true

Notice: /Stage[main]/Puppet::Config/Concat[/etc/puppetlabs/puppet/puppet.conf]/File[/etc/puppetlabs/puppet/puppet.conf]/content: current_value '{md5}80c36d9d03e380e9db04d16088aa31c8', should be '{md5}5ee4e42185caa115c3ddd0ba2d022dba' (noop)
```

alexjfisher:
I remember @bastelfreak trying to make a similar change and us deciding that it wasn't this easy.
```puppet
puppet::config::entry {'example_setting 1':
    key     => 'example_setting',
    value   => '1',
    section => 'example_section',
  }
puppet::config::entry {'example_setting 2':
    key     => 'example_setting',
    value   => '2',
    section => 'example_section',
}
```
would have previously generated...
```
[example_section]
    example_setting = 1,2
```

now it generates
```
[example_section]
    example_setting = 1
,2
```

manifests/config/entry.pp
```diff
@@ -39,7 +39,7 @@
   if (!defined(Concat::Fragment["puppet.conf_${section}_${key}"])){
     concat::fragment{"puppet.conf_${section}_${key}":
```
alexjfisher: This fragment will always get a newline now.  This breaks adding multiple values to the same key via different instances of `puppet::config::entry` (the resources added in the `else` below).

alexjfisher:
Here's the IRC logs. https://gist.github.com/alexjfisher/30fe5623f620c5af5666a93407e9e284

baurmatt:
Adding multiple values to the same key with different `puppet::config::entry`'s doesn't sound like an indented feature to be honest.

alexjfisher:
It looks very deliberate to me.  The fragment in the `else` would never trigger otherwise.  On the other hand, I wouldn't be that surprised if the feature had no users??  Ripping it out might be an option and I think would be preferable to leaving it in but broken.

baurmatt:
Right! :( Removing it does indeed sound like a better option than just leaving it in.

@ekohl What do you think?

PR #709, opened 2019-10-17
WIP/RFC: Add `run_agent` task

alexjfisher:


tasks/run_agent.rb
```diff
@@ -0,0 +1,160 @@
...
+  def agent_locked?
+    File.exist?(agent_catalog_run_lockfile)
+  end
+
+  def agent_disabled?
+    File.exist?(agent_disabled_lockfile)
+  end
+
+  def disabled_reason
+    file = File.open agent_disabled_lockfile
```
ekohl: Isn't this equivalent to `JSON.load(File.read(agent_disabled_lockfile))`?

tasks/run_agent.rb
```diff
@@ -0,0 +1,160 @@
+#!/opt/puppetlabs/puppet/bin/ruby
```
alexjfisher: This seems to be the most sensible default.  Non AIO users can override this in their bolt configuration with [`interpreters`](https://puppet.com/docs/bolt/latest/bolt_configuration_options.html#global-configuration-options).

tasks/run_agent.rb
```diff
@@ -0,0 +1,160 @@
...
+
+require 'open3'
+require 'etc'
+require 'json'
+require 'yaml'
+require_relative '../../ruby_task_helper/files/task_helper.rb'
```
alexjfisher: Should I add [`puppetlabs/ruby_task_helper`](https://forge.puppet.com/puppetlabs/ruby_task_helper) to the metadata.json?
As far as the module is concerned, it's a 'soft' dependency.  But I also don't envisage a module namespace clash, so adding it to the metadata might be more helpful to users.

tasks/run_agent.rb
```diff
@@ -0,0 +1,160 @@
...
+
+class RunAgentTask < TaskHelper
```
alexjfisher: No idea what to use for the class name.  The [docs](https://forge.puppet.com/puppetlabs/ruby_task_helper#usage) aren't especially helpful.

tasks/run_agent.rb
```diff
@@ -0,0 +1,160 @@
...
+    waited = 0
+    while agent_locked? && waited < @wait_time
+      sleep 1
+      waited += 1
+    end
+
+    if agent_locked?
+      raise TaskHelper::Error.new(
+        'Lockfile still exists after waiting',
+        'theforeman-puppet/lockfile_timeout_expired'
```
alexjfisher: [> kind - A standard string for machines to handle. You may share kinds between your modules or namespace kinds per module.](https://puppet.com/docs/bolt/latest/writing_tasks.html#returning-errors-in-tasks)

Not really sure if this is a good choice or how important it is.

tasks/run_agent.rb
```diff
@@ -0,0 +1,160 @@
...
+      )
+    end
+
+    waited
```
alexjfisher: I don't think I'm actually using this returned value at the moment.  I suppose I could bung it into the response??

tasks/run_agent.rb
```diff
@@ -0,0 +1,160 @@
...
+      )
+    end
+  end
+
+  def run_puppet
+    cmd = [@puppet_bin, 'agent', '-t', '--detailed-exitcodes']
+    cmd << if @noop
+             '--noop'
+           else
+             '--no-noop'
```
alexjfisher: This is to override the value in the config file.  I should probably make this clearer in whatever docs I write or in the task metadata.

tasks/run_agent.rb
```diff
@@ -0,0 +1,160 @@
...
+           end
+
+    stdout, stderr, status = Open3.capture3(*cmd)
+
+    case status.exitstatus
+    when 0
+      {
+        result: 'The run succeeded with no changes or failures; the system was already in the desired state (or --noop was used).',
```
alexjfisher: Most of this text was taken from the puppet docs.  I added the bit about `noop` myself.  Maybe I should open a PR to fix this upstream ;)

runejuhl:
Quoting myself from https://gist.github.com/alexjfisher/753c61e09e891c9c9336ba0522b581a4#gistcomment-3057265

> If you plan on using this not just for development you should probably avoid using `-t` (`--test`):
> 
> > --test
> > 
> >     Enable the most common options used for testing. These are 'onetime', 'verbose', 'no-daemonize', 'no-usecacheonfailure', 'detailed-exitcodes', 'no-splay', and 'show_diff'.
> https://puppet.com/docs/puppet/latest/man/agent.html
> 
> Specifically you don't want the `no-usecacheonfailure`:
> 
> > usecacheonfailure
> > 
> > Whether to use the cached configuration when the remote configuration will not compile. This option is useful for testing new configurations, where you want to fix the broken configuration rather than reverting to a known-good one.
> > 
> >     Default: true
> https://puppet.com/docs/puppet/latest/configuration.html#usecacheonfailure

That is, if you use `-t`/`--test` you implicitly set `no-usecacheonfailure`, which usually isn't what you want for a production run.

tasks/run_agent.rb
```diff
@@ -0,0 +1,160 @@
...
+      raise TaskHelper::Error.new(
+        'Puppet agent needs to run as root',
+        'theforeman-puppet/bad_euid',
+        euid: Process.euid
+      )
+    end
+  end
+
+  def run_puppet
+    cmd = [@puppet_bin, 'agent', '-t', '--detailed-exitcodes']
```
runejuhl: Sorry, with https://github.com/theforeman/puppet-puppet/pull/709#issuecomment-544475964 I was of course referring to this line.

alexjfisher: For a manual run (as opposed to a normal scheduled run), I think I'd personally never want to accidentally be using the cache.  I do need a way of making this and other options configurable though.

alexjfisher: `--detailed-exitcodes` is redundant here though (already included in `-t`).

runejuhl: Yes, agreed, having a simple toggle for `--test` (which doesn't use cache) and `--no-daemonize --onetime --no-splay --detailed-exitcodes` (which would use cache) would be nice.

alexjfisher: Can you override options set with `--test` by setting conflicting arguments?
eg does
```
puppet agent -t --usecacheonfailure
```
work?

alexjfisher: answering my own question... Apparently not.

runejuhl: Seems like it:

~~~
$ puppet apply --noop -e 'file { "/tmp/omgtest": ensure => file, }'
Notice: Compiled catalog for managua.lan in environment production in 0.01 seconds
Notice: /Stage[main]/Main/File[/tmp/omgtest]/ensure: current_value 'absent', should be 'file' (noop)
Notice: Class[Main]: Would have triggered 'refresh' from 1 event
Notice: Stage[main]: Would have triggered 'refresh' from 1 event
Notice: Applied catalog in 0.02 seconds

$ puppet apply --noop -e 'file { "/tmp/omgtest": ensure => file, }' --no-noop
Notice: Compiled catalog for managua.lan in environment production in 0.01 seconds
Notice: /Stage[main]/Main/File[/tmp/omgtest]/ensure: created
Notice: Applied catalog in 0.01 seconds
~~~

And yes, it'd be much better to use the default `--test`.

runejuhl: Seems to work just fine with `puppet apply`. Last argument takes precedence:

~~~
$ puppet apply --no-noop -e 'file { "/tmp/omgtest": ensure => file, }' --noop
Notice: Compiled catalog for managua.lan in environment production in 0.01 seconds
Notice: /Stage[main]/Main/File[/tmp/omgtest]/ensure: current_value 'absent', should be 'file' (noop)
Notice: Class[Main]: Would have triggered 'refresh' from 1 event
Notice: Stage[main]: Would have triggered 'refresh' from 1 event
Notice: Applied catalog in 0.01 seconds
~~~

alexjfisher: I went for `puppet agent -t --splay` and the run *didn't* wait a random amount of time.

tasks/run_agent.rb
alexjfisher: yes. Updated

tasks/run_agent.rb
alexjfisher: @runejuhl I've added a `puppet_settings` variable.  It's an optional hash where you can override the default settings.

ekohl:
@alexjfisher what do we want to do with this? I don't have bolt or the capability to maintain this. Would you be interested in becoming a co-maintainer for the module?

PR #715, opened 2019-11-26
Stop acceptance tests on EL7

ekohl:
Because of https://github.com/moby/moby/issues/38749 we can't start puppetserver under systemd. That makes all our acceptance tests useless.


mmoll:
merged, bedankt @ekohl!

Issue #714, opened 2019-11-22
puppetserver cipher suites must be upgraded


fschaer:
Hi,

I got a very hard time upgrading my puppetdb from 5.2 to 6.7.3-1.

According to documentation, if not specified, puppetdb will use "only non-DHE ciphers".
See : https://puppet.com/docs/puppetdb/latest/configure.html#cipher-suites

When starting in debug mode, I found that it in fact only accepts these keys on CentOS 7.7.1908 + jdk 1.8 : 

[...]

PR #717, opened 2019-12-03
Update cipher suites

fschaer:
This change is needed as puppetlabs is allowing only (non ?)-DH ciphers in puppetdb6, which in turn causes connectivity failures between puppet server and puppetdb if no common cipher is found. Unfortunately, changing the puppet server ciphers is not enough, setting the puppetdb ciphers is also necessary to make sure all communications are ok - this was added in the README in the puppetdb integration section as this change has to be done outside this module.

Fixes GH-714


noqqe:
I can reproduce the behavior. PuppetDB >6.7.2 does not support the ciphers that are being configured by the module.

mmoll:
-> GH-721

PR #719, opened 2019-12-04
Move parameters to advanced

ekohl:
Most parameters are actually not touched by most users. This makes it clear.


mmoll:
merged, bedankt @ekohl!

PR #720, opened 2020-01-15
Add server_multithreaded parameter

alexjfisher:
Puppet Server 6.8 has just been released with an experimental new
feature.

> This release adds a new JRuby pool architecture that maintains a
> single JRuby instance through which requests to Puppet Server are run
> concurrently. In this mode, the server's memory footprint is
> significantly lighter, because it no longer needs to run multiple JRuby
> instances. Toggle this behavior by setting the
> `jruby-puppet.multithreaded` to `true`.

This commit adds support for this new setting.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -387,6 +387,8 @@
 #                                           503 responses returned when max-queued-requests is enabled. (Puppetserver 5.x only)
 #                                           Defaults to 1800 for Puppetserver >= 5.0
 #
+# $server_multithreaded::                   Use multithreaded jruby. (Puppetserver >= 6.8 only).  Defaults to false.
+#
 # $server_idle_timeout::                    How long the server will wait for a response on an existing connection
 #
 # $server_connect_timeout::                 How long the server will wait for a response to a connection attempt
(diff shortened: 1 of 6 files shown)


ekohl:
Looking closer, it would be better to use the `real_puppetserver_version` variable as determined here:
https://github.com/theforeman/puppet-puppet/blob/0de3202b7b4ca41213656496dee27af8c26c28aa/manifests/server.pp#L475-L483
The reason is that the version number might be `undef` (default).

alexjfisher:
I think this is already the case? (Although the module can be a bit confusing to follow).  The template function is called from puppetserver.pp and uses `$server_puppetserver_version`.  This defaults to `$::puppet::server::real_puppetserver_version`.

https://github.com/theforeman/puppet-puppet/blob/0de3202b7b4ca41213656496dee27af8c26c28aa/manifests/server/puppetserver.pp#L108

ekohl:
You're right on that part. However, maybe it's good to add another condition that if the puppetversion is at least 6.12.0 then the puppetserver version is 6.8.0?

alexjfisher:
> You're right on that part. However, maybe it's good to add another condition that if the puppetversion is at least 6.12.0 then the puppetserver version is 6.8.0?

That could very easily not be true though.  The current code is conservative.  If your puppet version is 6.X, then so is your puppetserver.  Much more than that would probably be too great an assumption.

ekohl:
yes, in general it's a complicated situation.

PR #721, opened 2020-01-15
Update cipher suites

mmoll:
This change is needed as puppetlabs is allowing only (non ?)-DH ciphers in puppetdb6, which in turn causes connectivity failures between puppet server and puppetdb if no common cipher is found. Unfortunately, changing the puppet server ciphers is not enough, setting the puppetdb ciphers is also necessary to make sure all communications are ok - this was added in the README in the puppetdb integration section as this change has to be done outside this module.

Fixes GH-714

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,12 @@ Please see the notes about using puppetlabs/puppetdb 5.x with older versions of 
 newer releases of the module and set the values via hiera or an extra include of `puppetdb::globals` with
 `puppetdb_version` defined.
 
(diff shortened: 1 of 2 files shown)


mmoll:
@fschaer @noqqe could you check this?

fschaer:
Cant test for now cause I have to re-install a dev puppetdb, will try asap.
But from what I remember, the RSA keys were the only ones that I found to be working with puppetdb

ekohl:
@mmoll not entirely sure this is backwards incompatible, but I always consider changing defaults a good reason. Let me know if you disagree.

PR #728, opened 2020-03-20
Add server_ca_client_self_delete to CA Servers

neilfromit:
Add the server_ca_client_self_delete boolean parameter to CA servers.
When true, allow puppet clients to delete their own certificate from
the Puppet CA.

### DIFF
--- a/spec/classes/puppet_server_puppetserver_spec.rb
+++ b/spec/classes/puppet_server_puppetserver_spec.rb
@@ -280,6 +280,14 @@ describe 'puppet' do
         end
       end
 
+      describe 'ca_client_self_delete' do
+        context 'when set' do
+          let(:params) { super().merge(server_ca_client_self_delete: true)}
+          it { should contain_file(auth_conf)
+            .with_content(%r{^(\ *)name: "Allow nodes to delete their own certificates",$}) }
+        end
+      end
+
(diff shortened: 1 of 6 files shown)


neilfromit:
Just to expand a little more, the use case for this feature for me was to allow for puppet agent certificate self renewal. I'm using the [reidmv/puppet-module-puppet_certificate](https://github.com/reidmv/puppet-module-puppet_certificate) in particular to achieve this. Ideally auth.conf should be a little less static but this was the simpler commit for now. I've included a basic spec test.

spec/classes/puppet_server_puppetserver_spec.rb
```diff
@@ -280,6 +280,14 @@
         end
       end
 
+      describe 'ca_client_self_delete' do
+        context 'when set' do
+          let(:params) { super().merge(server_ca_client_self_delete: true)}
+          it { should contain_file(auth_conf)
+            .with_content(%r{^(\ *)name: "Allow nodes to delete their own certificates",$}) }
```
mmoll: Could you also check (around line 109) that it's not added by default?

mmoll:
merged, thanks @neilfromit!

neilfromit:
Awesome! Thanks for reviewing it.

PR #729, opened 2020-03-27
add Ubuntu integration tests

mmoll:


mmoll:
Ubuntu 16.04 suffers from a systemd bug in docker containers, so it's not added.

PR #731, opened 2020-03-30
Add server_trusted_external_command parameter

### LINKED ISSUE #730: Add trusted_external_command parameter
Puppet 6.11.0 introduced a new feature called "trusted_external_command". It's basically a external node classifer where the result get saved into PuppetDB.

https://puppet.com/docs/puppet/latest/release_notes_puppet.html#experimental-feature:-catalog-compilation-with-external-trusted-data-from-third-parties

https://puppet.com/docs/puppet/latest/configuration.html#trustedexternalcommand

I will provide a PR.

baurmatt:
Make it possible to configure the 'trusted_external_command' parameter.
https://puppet.com/docs/puppet/latest/configuration.html#trustedexternalcommand

Fixes #730.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -227,6 +227,9 @@
 #
 # $server_external_nodes::                  External nodes classifier executable
 #
+# $server_trusted_external_command::        The external trusted facts script to use.
+#                                           (Puppet >= 6.11 only).
+#
 # $server_git_repo::                        Use git repository as a source of modules
 #
 # $server_environments_owner::              The owner of the environments directory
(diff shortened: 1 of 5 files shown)


mmoll:
:+1: code wise, however, I think it would be nice to limit its use to new enough Puppet versions.

baurmatt:
As the defaut ist `false` I'm not sure if that is required given that the documentation clearly states that this feature is for Puppet > 6.11. What exactly do you have in mind? :) Just fail if `server_trusted_external_command` is given for < 6.11? :)

mmoll:
> Just fail if server_trusted_external_command is given for < 6.11? :)

It might be a bit harsh, but I guess it's the right thing, as just going over it on older clients might give a false sense of the state the environment is put into.

mmoll:
merged, danke @baurmatt!

PR #732, opened 2020-03-31
Refactor PuppetDB integration

ekohl:
Since Puppet 4.0 the active record backend is gone and the only one is PuppetDB. This is the default value for backend is puppetdb so we can simply expose the storeconfigs boolean.

The PuppetDB parameters aren't for the puppetserver itself but rather the puppetdb config in a way that's compatible with this module. Separating it into a class makes it a bit clearer that it really needs an additional Puppet module.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,10 +102,10 @@ wrapper classes or even your ENC (if it supports param classes). For example:
 
     # Want to integrate with an existing PuppetDB?
     class { '::puppet':
-      server                      => true,
-      server_puppetdb_host        => 'mypuppetdb.example.com',
-      server_reports              => 'puppetdb,foreman',
-      server_storeconfigs_backend => 'puppetdb',
+      server               => true,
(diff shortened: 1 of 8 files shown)


ekohl:
I've added an additional commit that refactors the PuppetDB config to a separate class.

oranenj:
Seems good to me, though I'd change the PR title to something less specific since it's not only about storeconfigs anymore. Something like "Refactor PuppetDB integration" perhaps?

ekohl:
Looks like I forgot to set facts in the new test file. Updated.

mmoll:
still :broken_heart:

ekohl:
Blocked on https://github.com/theforeman/puppet-puppet/pull/735

mmoll:
tests fail :broken_heart:

ekohl:
Not my best work, but it's green now.

mmoll:
merged, bedankt @ekohl!

PR #735, opened 2020-04-16
Modulesync: make voxpupuli helpers work

ekohl:
46c40319775e2293a9e92556fe68755fa348f92d was merged too soon. This makes it work, with the exception of removal of Debian 10 + Puppet 5 (for which no packages are present). That's the only manual change left.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,6 +36,7 @@ junit/
 pkg/
 coverage/
 .yardoc/
+REFERENCE.md
 
 ## InteliJ / RubyMine
 .idea
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -280,7 +280,7 @@
       `$server_http` parameter.
 * Other features:
     * Support native puppetserver package on FreeBSD
-    * Allow disabling crl when `server_ca => true` 
+    * Allow disabling crl when `server_ca => true`
     * Add SLES AIO agent support
     * Add support for Parallels PSBM
 * Other changes and fixes:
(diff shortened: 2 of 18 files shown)


ekohl:
Looks like we're now using the modern fact and our override no longer works:
https://github.com/theforeman/puppet-puppet/blob/46c40319775e2293a9e92556fe68755fa348f92d/spec/support/aio.rb#L5

ekohl:
Includes https://github.com/theforeman/puppet-puppet/pull/737.

ekohl:
Locally all the tests are green now. Hopefully Travis agrees.

ekohl:
The failure is now in puppet-foreman. Looks like that combination is untested. I've been considering pulling the puppetmaster parts out to their own module so we can properly test it.

ekohl:
https://github.com/theforeman/puppet-puppetserver_foreman/pull/1 contains fixes. To unblock master, I'll submit the same fixes to puppet-foreman now but we should port over.

ekohl:
https://github.com/theforeman/puppet-foreman/pull/832 is the short term fix.

ekohl:
New attempt: https://github.com/theforeman/puppet-foreman/pull/834

ekohl:
:green_heart: now. Note it should be merged fast forward so https://github.com/theforeman/puppet-puppet/pull/737 is also merged. This means the Github UI isn't the right tool.

mmoll:
merged, bedankt @ekohl!

PR #736, opened 2020-04-17
Drop Foreman API version parameters

ekohl:
The latest theforeman-foreman has removed these and for older versions they have the only possible value as a default already.


ekohl:
Needs https://github.com/theforeman/puppet-puppet/pull/735 to pass, but that one needs this. It needs a separate entry in the changelog so I propose we merge this even though it's broken.

mmoll:
merged, bedankt @ekohl!

PR #741, opened 2020-05-05
Revert "Convert the systemd timer service to camptocamp/systemd"

ekohl:
This reverts commit 4302973e8f47c77ed35b2a399c1ccef2d503b400, except the dependency bits. It's also updated to be lint clean.

The reason to revert this is that the integrated daemon reload in systemd::timer leads to dependency cycles in the Katello scenario.


ehelms:
Thanks for this!

ekohl:
Me neither, but I think puppet-systemd needs a better way to deal with it. I started https://github.com/camptocamp/puppet-systemd/pull/148 which will only be a solution when we drop Puppet 5. The alternative is a solution like https://github.com/camptocamp/puppet-systemd/blob/1e1560015321b0bd3d51b15a22231e98b8ec8898/manifests/dropin_file.pp#L59. However, I did this now to unblock katello nightly.

PR #745, opened 2020-05-13
Add Fedora 31, drop Fedora 26

ekohl:
Fedora 26 is EOL. Fedora 32 is not added since there are no packages available.

### DIFF
--- a/.sync.yml
+++ b/.sync.yml
@@ -4,6 +4,7 @@
     - centos6-64
     - debian9-64
     - debian10-64
+    - fedora31-64
     - ubuntu1804-64
   env:
     global:
--- a/.travis.yml
+++ b/.travis.yml
@@ -70,6 +70,28 @@ matrix:
     - rvm: 2.5.1
       env:
         - BEAKER_PUPPET_COLLECTION=puppet5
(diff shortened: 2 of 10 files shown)


ekohl:
Looks like cron is not available, like in the ubuntu images. I've submitted https://github.com/puppetlabs/puppetlabs-stdlib/pull/1108 which can make our code dynamic rather than having to hardcode available providers. For now I've just ensured the package is installed in our acceptance tests.

mmoll:
As it seems, there are no puppetserver RPMs available for F31 currently.

ekohl:
Not entirely sure how to deal with this now. On the one hand, the agent works but on the other, it's not entirely there. We have `unsupported_puppetmaster_osfamily` but Fedora is not its own family.

mmoll:
Maybe we can just add a `puppetserver_supported` variable? Then the Ubuntu/focal PR also could go in and we can enable it per OS, once new puppetserver builds are available...

ekohl:
So something like this. I added a trivial agent test case to at least test something

mmoll:
merged, bedankt @ekohl!

PR #747, opened 2020-05-27
Don't re-create existing CA certs

zipkid:
When starting a new puppetserver as CA with existing CA files, the current code tries to perform 'puppetserver ca setup' because the puppetserver's cert does not exist yet.
However, that command fails because the CA cert files exist and can, correctly, not be overwritten.

This PR changes checking for the CA cert, instead of the server's own cert.


ekohl:
Thanks! I've taken the liberty to squash and modify the commit message slightly.

PR #749, opened 2020-06-24
Move to puppetserver_foreman

ekohl:
The Foreman integration bits are being split out to their own module. This changes over the integration.

Depends on https://github.com/theforeman/puppet-puppetserver_foreman/pull/1


ekohl:
I believe this is now ready to merge.

ehelms (approved):
Are the test failures expected?

ekohl:
Yes:
* CentOS 6 fails because the yum repos are gone
* Debian 10 + Puppet 5 fails because there are no Puppet 5 packages on Debian 10
* Same for Ubuntu 20.04 + Puppet 5

PR #753, opened 2020-07-21
allow puppet server to retrieve v4 catalog

vchepkov:
I have replicated configuration which is supplied by PE

Default 'deny' rule doesn't match it's comment and is redundant, due to global "puppetlabs deny all" rule

### DIFF
--- a/templates/server/puppetserver/conf.d/auth.conf.erb
+++ b/templates/server/puppetserver/conf.d/auth.conf.erb
@@ -19,12 +19,17 @@ authorization: {
 <%- if scope.function_versioncmp([@server_puppetserver_version, '6.3.0']) >= 0 -%>
         {
             # Allow services to retrieve catalogs on behalf of others
+            allow: [
+<%- @server_admin_api_whitelist.each do |client| -%>
+                "<%= client %>",
+<%- end -%>
+            ]
             match-request: {
                 path: "^/puppet/v4/catalog/?$"
+                query-params: {}
                 type: regex
                 method: post
             }
-            deny: "*"
             sort-order: 500
             name: "puppetlabs v4 catalog for services"
         },


vchepkov:
It's unrelated, but it does allow catalog_diff to work

ekohl:
Ok thanks! I'll let the tests continue to run before merging.

ekohl:
I just realized this was still open, but merged https://github.com/theforeman/puppet-puppet/commit/ed78b93173bb982683e38e1face0c1900172580c so now there's a merge conflict.

vchepkov:
I will rebase

alexjfisher:
@vchepkov Do you need to rebase?  I used `server_puppetserver_trusted_agents` as that was the whitelist being allowed in the v3 endpoint.

vchepkov:
Hmm, that's a real conflict. my change was to allow server_admin_api_whitelist, merged change was to allow server_trusted_agents. I still think adding deny rule, as code is doing is redundant, especially with the comment that says 'allow', but if that's what was accepted, I can close this PR, let me know what you think

vchepkov:
IMHO, code should be changed to add allow rule if `server_puppetserver_trusted_agents` is defined and version supports it and don't add it at all otherwise. But my change seems outdated, so I will close this request. Thanks

alexjfisher:
I think the `deny` was taken from the upstream default config.  I agree it doesn't look necessary, but I thought it better to leave it be.  This way, the majority of users won't see any change (and won't have their servers restarted).

The puppetserver admin API is perhaps poorly named.  'admin' might imply elevated/special privileges, but in reality the admin api only lets you clear the environment cache and recycle jruby instances.  Nothing destructive and nothing that leaks secrets.  Users of the `server_admin_api_whitelist` parameter might have been very surprised/alarmed to see that those certs could now fetch any catalog.

PR #755, opened 2020-08-04
Release 14.1.0

wbclark:


wbclark:
> I just merged #756 so would you mind updating to reflect that? It may be a good thing to also check #753. I realized that it's very similar to #756 and checking the diffs now.

Done.

I updated the date as well.

PR #615, opened 2018-08-13
Revert "use puppetlabs-{hocon, puppet_authorization} for auth.conf

mmoll:
This reverts commit 6cb7115a55be299e9326f5385a8a14cc11e69bb3.

closes GH-595

PR #756, opened 2020-08-05
Use server_trusted_agents in v4 catalog endpoint

alexjfisher:


templates/server/puppetserver/conf.d/auth.conf.erb
```diff
@@ -12,7 +12,7 @@ authorization: {
                 type: regex
                 method: [get, post]
             }
-            allow: <%= @server_trusted_agents << '$1' %>
+            allow: <%= @server_trusted_agents + ['$1'] %>
```
alexjfisher, line 15: It's not just in this template that `server_trusted_agents` was being mutated, but in the catalog itself. So much for puppet variables being immutable! I guess in irb, you can accidentally abuse this. ;)

templates/server/puppetserver/conf.d/auth.conf.erb
```diff
@@ -24,7 +24,11 @@ authorization: {
                 type: regex
                 method: post
             }
+<%- if @server_trusted_agents.empty? -%>
             deny: "*"
+<%- else -%>
+            allow: <%= @server_trusted_agents %>
```
alexjfisher, line 30: I also would have liked to introduce a new parameter so that certificate extensions could have been used as well as just a list of certnames. irb instead of the old way (removed in https://github.com/theforeman/puppet-puppet/pull/615) makes it more difficult though.  If I just append a hash to the `server_trusted_agents` array the outputted string isn't valid hocon.  Valid configuration would look a bit like...

```
allow: [
        "cert1",
        "cert2",
        {
            extensions: {
                pp_authorization: "catalog"
            }
        }
    ]
```
I might try to figure out a nice way of generating this in a separate PR.

ekohl:
I guessed this is an enhancement, or is it a bugfix?

alexjfisher:
It should have probably been part of https://github.com/theforeman/puppet-puppet/pull/689??  Not sure whether I'd call that a bug though?

alexjfisher:
It doesn't fix a regression, so probably `enhancement` is fine.

templates/server/puppetserver/conf.d/auth.conf.erb
vchepkov: FYI, PE rule has additional restriction for v4 catalog `query-params: {}`, I presume for security reasons

here is the code snippet:

```
        {
            "allow": [
                "puppet.chepkov.lan",
                {
                    "rbac": {
                        "permission": "puppetserver:compile_catalog:*"
                    }
                }
            ],
            "match-request": {
                "method": "post",
                "path": "^/puppet/v4/catalog/?$",
                "query-params": {},
                "type": "regex"
            },
            "name": "puppetlabs v4 catalog",
            "sort-order": 500
        }
    ,
```

templates/server/puppetserver/conf.d/auth.conf.erb
alexjfisher: oops.  That was me! https://github.com/theforeman/puppet-puppet/commit/4f07b5fac070f1c9ee8762df3852787f1d0558f3#diff-6775a3530d5d8059f4175382a1af402cR12

templates/server/puppetserver/conf.d/auth.conf.erb
alexjfisher: It wouldn't hurt.  Don't think it does much either though.  The body of the post is json containing all the input data.  I think any query parameters would just be ignored? https://puppet.com/docs/puppetserver/latest/puppet-api/v4/catalog.html

templates/server/puppetserver/conf.d/auth.conf.erb
alexjfisher: Also, 'naming things is hard'.  Maybe I should have called it `trusted_compile_clients` or something instead.

PR #758, opened 2020-08-07
Add `server_puppetserver_trusted_certificate_extensions`

alexjfisher:
Accepts an array of hashes to allow for more complicated setups.
Typical use would probably just be a single array element such as...

```puppet
class { 'puppet':
  ...
  server_puppetserver_trusted_certificate_extensions => [{ 'pp_authorization' => 'catalog' }],
  ...
}
```

### DIFF
--- a/manifests/server.pp
+++ b/manifests/server.pp
@@ -285,6 +285,8 @@
 #
 # $puppetserver_trusted_agents::       Certificate names of agents that are allowed to fetch *all* catalogs. Defaults to empty array
 #
+# $puppetserver_trusted_certificate_extensions:: An array of hashes of certificate extensions and values.
+#                                      Example: [{ 'pp_authorization' => 'catalog' }]
 #
 # $ca_allow_sans::                     Allow CA to sign certificate requests that have Subject Alternative Names
 #                                      Defaults to false
@@ -431,6 +433,7 @@ class puppet::server(
   Boolean $puppetserver_experimental = $puppet::server_puppetserver_experimental,
   Optional[String[1]] $puppetserver_auth_template = $puppet::server_puppetserver_auth_template,
   Array[String] $puppetserver_trusted_agents = $puppet::server_puppetserver_trusted_agents,
+  Array[Hash] $puppetserver_trusted_certificate_extensions = $puppet::server_puppetserver_trusted_certificate_extensions,
   Optional[Enum['off', 'jit', 'force']] $compile_mode = $puppet::server_compile_mode,
   Optional[Integer[1]] $selector_threads = $puppet::server_selector_threads,
   Optional[Integer[1]] $acceptor_threads = $puppet::server_acceptor_threads,
--- a/spec/classes/puppet_server_puppetserver_spec.rb
+++ b/spec/classes/puppet_server_puppetserver_spec.rb
@@ -1,4 +1,5 @@
 require 'spec_helper'
+require 'hocon'
 
 describe 'puppet' do
   on_supported_os.each do |os, facts|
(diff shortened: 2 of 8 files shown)


spec/classes/puppet_server_puppetserver_spec.rb
```diff
@@ -276,10 +277,33 @@
...
-      describe 'server_trusted_agents' do
-        context 'when set' do
+      describe 'puppetlabs v3 catalog from agents rule' do
+        context 'when server_trusted_agents set' do
           let(:params) { super().merge(server_puppetserver_trusted_agents: ['jenkins', 'octocatalog-diff']) }
-          it { should contain_file(auth_conf).with_content(/^            allow: \["jenkins", "octocatalog-diff", "\$1"\]$/) }
+          it do
+            content = catalogue.resource('file', auth_conf).send(:parameters)[:content]
+            rules = Hocon.parse(content)['authorization']['rules']
+            rule = rules.find {|rule| rule['name'] == 'puppetlabs v3 catalog from agents' }
```
ekohl, line 286: Looks like you can extract this to a `let(:rule)` block and reuse it between the 3 tests

spec/classes/puppet_server_puppetserver_spec.rb
```diff
@@ -276,10 +277,33 @@
...
+            expect(rule['allow']).to eq(['jenkins', 'octocatalog-diff', '$1'])
+          end
+        end
+        context 'with server_trusted_certificate_extensions' do
```
ekohl, line 290: I always like to visually separate context blocks with an empty line (GH renders an ugly diff)
```suggestion

        context 'with server_trusted_certificate_extensions' do
```

templates/server/puppetserver/conf.d/auth.conf.erb
```diff
@@ -12,7 +12,13 @@ authorization: {
...
                 method: [get, post]
             }
-            allow: <%= @server_trusted_agents + ['$1'] %>
+<%= scope.call_function(
+  'to_json_pretty',
+  [
+    {'allow' => @server_trusted_agents + ['$1'] + @server_trusted_certificate_extensions.map { |extension| { 'extensions' => extension } } }
+  ]
+).lines.to_a[1..-2].map{ |line| "          #{line}"}.join
+-%>
```
ekohl, line 21: We migrated away from puppetlabs-hocon because puppetlabs-hocon because back then the support wasn't really clear. Looking at how they still update it I do wonder if we should reconsider. However, probably better to merge it now.

alexjfisher:
Can this get a review? (I was messaged on Puppet Community slack today.  @chadh is planning PRs in the same area).

manifests/server.pp
```diff
@@ -291,6 +291,8 @@
 #
 # $puppetserver_trusted_agents::       Certificate names of agents that are allowed to fetch *all* catalogs. Defaults to empty array
 #
+# $puppetserver_trusted_certificate_extensions:: An array of hashes of certificate extensions and values. Defaults to array.
```
ekohl, line 294: Is the `Defaults to array` really that useful? I see it's also for trusted_agents, but I have my doubts about the usefulness.

ekohl (commented):
Mostly looks good, some small improvements in the tests.

manifests/server.pp
alexjfisher: I made it an array to allow full flexibility.  Not sure anyone really wants to do https://github.com/theforeman/puppet-puppet/pull/758/files#diff-60ae41fd0a31977447947f59940ee9a4R475 but if it was just a plain hash, that wouldn't be possible.

templates/server/puppetserver/conf.d/auth.conf.erb
alexjfisher: I think it'd be much easier for users to add their own configs in their profiles if it still did use puppetlabs-hocon

manifests/server.pp
ekohl: I meant the literal `Defaults to array.` in the documentation string.

alexjfisher: oh... right.  Good point.

spec/classes/puppet_server_puppetserver_spec.rb
alexjfisher: and I guess `rules` and `content` too.

spec/classes/puppet_server_puppetserver_spec.rb
```diff
@@ -468,7 +486,30 @@
           end
           context 'with server_trusted_agents' do
             let(:params) { super().merge(server_puppetserver_trusted_agents: ['jenkins', 'octocatalog-diff']) }
-            it { should contain_file(auth_conf).with_content(%r{^(\ *)allow: \["jenkins", "octocatalog-diff"\]\n(\ *)sort-order: 500\n(\ *)name: "puppetlabs v4 catalog for services"}) }
+            it do
+              content = catalogue.resource('file', auth_conf).send(:parameters)[:content]
```
ekohl, line 490: Could you apply the same let trick here as above?

alexjfisher: oops.  Of course.

ekohl (approved):

Issue #759, opened 2020-08-14
Appending values with puppet::config::entry is broken


dnlsng:
This commit, e87d8f2a7a008e2d92f5e41999fee83117edec25, breaks the ability to add multiple puppet::config::entry resources for the same key. By setting ensure_newline to true on the concat resource, a new line is added at the end of every fragment which removes the ability to append values. The appended values appear on the next line which results in a puppet config file which is not valid.

One way to resolve this would be to revert this commit and then add a concat fragment at the end of the file with just a single new line character.

ekohl:
This was indeed mentioned @ https://github.com/theforeman/puppet-puppet/pull/707#issuecomment-546929749 but I never got around to reverting it.

A patch would be very welcome.

dnlsng:
Sorry, I've been busy and only just had a chance to get around to a patch.

PR #762, opened 2020-09-10
Fix appending puppet config entry

dnlsng:
This fixes issue #759 by reverting e87d8f2a7a008e2d92f5e41999fee83117edec25 and then adding a newline to the end of the puppet.conf file.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -73,6 +73,17 @@ class puppet::config(
     puppet::config::main { $key: value => $value }
   }
 
+  # Ensure there is a newline at the end of puppet.conf
+  $newline = $facts['os']['family'] ? {
+    'Windows' => "\r\n",
+    default   => "\n",
+  }
+  concat::fragment{'puppet.conf_eof_newline':
+    target  => "${puppet::dir}/puppet.conf",
+    content => $newline,
+    order   => '999_eof_newline',
+  }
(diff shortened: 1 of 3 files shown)


manifests/config.pp
```diff
@@ -73,6 +73,17 @@
     puppet::config::main { $key: value => $value }
   }
 
+  # Ensure there is a newline at the end of puppet.conf
+  $newline = $facts['os']['family'] ? {
+    'Windows' => "\r\n",
```
ekohl, line 78: This feels a bit odd since the other parts only use `\n` and not `\r\n`.

manifests/config/entry.pp
```diff
@@ -30,7 +30,7 @@
   # they make sure that '1_main ' is ordered before '1_main_*'
   ensure_resource('concat::fragment', "puppet.conf_${section}", {
       target  => "${puppet::dir}/puppet.conf",
-      content => "\n[${section}]",
+      content => "\n\n[${section}]",
```
ekohl, line 33: The only downside of this is that you end up with 2 empty lines at the start of the file. I'm not sure how to best deal with that.

ekohl:
https://github.com/theforeman/puppet-puppet/pull/782 was merged and resolves the issue.

PR #768, opened 2020-10-06
Remove `systemd` from puppet::agent::service case

alexjfisher:
Looks like #498 had a breaking change when it added an Enum for
`runmode` that didn't include `systemd` (only `systemd.timer`)

But that was a long time ago, so just tidying the code in
`puppet::agent::service`, instead of ammending the `Enum` in `init.pp`.


ekohl (approved):

alexjfisher:
Test failures should be fixed in https://github.com/theforeman/puppet-puppet/pull/767

Issue #752, opened 2020-07-13
setup fileserver and add access rights to /etc/puppetlabs/puppetserver/conf.d/auth.conf


tepecat:
Hey,
I would like setup a fileserver and set access rights, https://puppet.com/docs/puppet/latest/file_serving.html

Atm I did not find a possibility to use this module to configure /etc/puppetlabs/puppetserver/conf.d/auth.conf to add an access  rule.

Did I not find the right parameter to configure auth.conf or is it atm not possible to add own config to auth.conf?

Best regards.
thomas.

ekohl:
There is a parameter `auth_template` if you want full flexibility. Is that good enough?

vchepkov:
It would be convenient to just add a rule. For example, this module is very useful
https://github.com/camptocamp/puppet-catalog-diff and one can't easily modify auth.conf to allow necessary access

ekohl:
Since I don't use that, I don't have time to write a patch but will certainly accept a PR.

tepecat:
thank you for your suggestions, I will try auth_template and puppet-catalog-diff.

tepecat:
auth_template  is connected to ${puppet_dir}/auth.conf (see manifests/config.pp )and it is only useable, if use-legacy-auth-conf is true.
I am using use-legacy-auth-conf = false, so auth_template is not useable for new auth.conf (puppetserver/conf.d/auth.conf)

I add needed rule for my fileserver into templates/server/puppetserver/conf.d/auth.conf.erb now. Perhaps in a later version of this module it is possible to add additional rules into new auth.conf.

Thank you for your help!

ekohl:
Ah, you're right. It does appear to be hardcoded:
https://github.com/theforeman/puppet-puppet/blob/f63fdaa5d3ad2850483d15fb752d2171de18f4b8/manifests/server/puppetserver.pp#L247-L250

Sadly, I don't have time to dive into this but a PR is welcome.

chadh:
My use case is different, but relevant.  I need to update the allowed hosts for catalog, node, and report (we use a different node_name_fact than certname).  I think allowing the module user to provide a template for the puppetserver auth.conf is probably a good workaround for now.  Ideally maybe we would have types for building that file, but I know I don't have the cycles to implement that right now.

PR #769, opened 2020-10-07
adds param for user-provided puppetserver auth.conf

chadh:
There are a lot of special use cases for auth.conf, and rather than 100% manage that file, this PR gives the module user a way to provide their own.  Ideally we would use an epp template and also provide a hash of arguments for the template , but I tried to just match the existing pattern in this PR.  This potentially address #752 .

### DIFF
--- a/manifests/server/puppetserver.pp
+++ b/manifests/server/puppetserver.pp
@@ -120,6 +120,7 @@ class puppet::server::puppetserver (
   $metrics_graphite_interval              = $puppet::server::metrics_graphite_interval,
   $metrics_allowed                        = $puppet::server::metrics_allowed,
   $server_experimental                    = $puppet::server::puppetserver_experimental,
+  $server_auth_template                   = $puppet::server::puppetserver_auth_template,
   $server_trusted_agents                  = $puppet::server::puppetserver_trusted_agents,
   $allow_header_cert_info                 = $puppet::server::allow_header_cert_info,
   $compile_mode                           = $puppet::server::compile_mode,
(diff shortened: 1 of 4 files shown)


manifests/server/puppetserver.pp
```diff
@@ -244,9 +245,15 @@
...
-    content => template('puppet/server/puppetserver/conf.d/auth.conf.erb'),
+  if $server_auth_template {
+    file { "${server_puppetserver_dir}/conf.d/auth.conf":
+      ensure => file,
+      content => template($server_auth_template),
+    }
+    file { "${server_puppetserver_dir}/conf.d/auth.conf":
+      ensure  => file,
+      content => template('puppet/server/puppetserver/conf.d/auth.conf.erb'),
+    }
```
ekohl, line 256: I think you're missing an `else` but I'd prefer `pick($server_auth_template, 'puppet/server/puppetserver/conf.d/auth.conf.erb')`.

manifests/server/puppetserver.pp
```diff
@@ -244,9 +245,10 @@
     content => template('puppet/server/puppetserver/conf.d/puppetserver.conf.erb'),
   }
 
+  $auth_template = pick($server_auth_template, 'puppet/server/puppetserver/conf.d/auth.conf.erb')
   file { "${server_puppetserver_dir}/conf.d/auth.conf":
-    ensure  => file,
-    content => template('puppet/server/puppetserver/conf.d/auth.conf.erb'),
+    ensure => file,
```
ekohl, line 250: I think lint will complain about this
```suggestion
    ensure  => file,
```

manifests/server/puppetserver.pp
chadh: sorry, that was super sloppy of me.

ekohl (approved):
Thanks! I do indeed consider this a solution to #752. It may not be the prettiest since ERB templates and their context are fragile, but it at least provides flexibility.

PR #770, opened 2020-10-20
Update documentation for change from puppet master to server

tbrisker:
Puppet has been using `server` naming since 5.x, and the puppet master
implementation is long deprecated and unsupported. Additionally,
`compile masters` have been renamed to `compilers` in 6.x. Updating
our documentation to use the current nomenclature.
No parameters were renamed to preserve backwards compatibility.
Additionally, to minimize changes no changes were made with regards to
capitalization or refering to `puppetserver` vs. `puppet server`

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,9 @@
 [![Puppet Forge](https://img.shields.io/puppetforge/v/theforeman/puppet.svg)](https://forge.puppetlabs.com/theforeman/puppet)
 [![Build Status](https://travis-ci.org/theforeman/puppet-puppet.svg?branch=master)](https://travis-ci.org/theforeman/puppet-puppet)
 
-# Puppet module for installing the Puppet agent and master
+# Puppet module for installing the Puppet agent and server
 
-Installs and configures the Puppet agent and optionally a Puppet master (when
+Installs and configures the Puppet agent and optionally a Puppet server (when
 `server` is true).  Part of the [Foreman installer](https://github.com/theforeman/foreman-installer)
 or to be used as a Puppet module.
 
@@ -13,7 +13,7 @@ If you know you'll be installing an earlier or specific version, you will
 need to override `server_puppetserver_version`. More information in the Puppet
 Server section below.
 
-Many puppet.conf options for agents, masters and other are parameterized, with
+Many puppet.conf options for agents, servers and other are parameterized, with
 class documentation provided at the top of the manifests. In addition, there
 are hash parameters for each configuration section that can be used to supply
 any options that are not explicitly supported.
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -80,7 +80,7 @@ wrapper classes or even your ENC (if it supports param classes). For example:
     # Agent and cron (or daemon):
     class { '::puppet': runmode => 'cron' }
```
ekohl, line 81: Unrelated, but I should do a follow up that also removes `::`. That was a Puppet 3/4 compatibility thing we no longer need to care about.

manifests/server.pp
```diff
@@ -31,32 +31,32 @@
...
+# $port::                              Puppet server port
 #
 # $ca::                                Provide puppet CA
 #
 # $ca_crl_filepath::                   Path to ca_crl file
 #
-# $ca_crl_sync::                       Sync the puppet ca crl to compile masters. Requires compile masters to
-#                                      be agents of the CA master (MOM) defaults to false
+# $ca_crl_sync::                       Sync the puppet ca crl to compilers. Requires compilers to
+#                                      be agents of the CA server (MOM) defaults to false
```
ekohl, line 49: MOM here refers to Master of Masters. I'm not sure what the new term for that is.

manifests/init.pp
```diff
@@ -164,7 +164,7 @@
 #
 # $agent_noop::                             Run the agent in noop mode.
 #
-# $puppetmaster::                           Hostname of your puppetmaster (server
+# $puppetmaster::                           Hostname of your puppetserver (server
```
ekohl, line 167: We may also need to update this, but I don't really know what a good name would be. `agent_server` is also not that clear and just `server` is already taken.

manifests/server.pp
tbrisker: The 2019.8 docs state: 
> "We've unified infrastructure terminology across all installation types. We now call compile masters compilers to reflect their role: compiling catalogs. Similarly, we call the master a master, whether or not your installation includes compilers. In high availability installations, the node that replicates data from the master is simply a replica. R.I.P. MoM, master of masters, and primary master replica."

but tbh i'm not sure i understand what is supposed to replace MoM? simply Master? how does that fit with the change from master to server everywhere?

manifests/init.pp
tbrisker: maybe `agent_server_hostname`?

manifests/server.pp
ekohl: @binford2k not sure who could answer this. Would you mind taking a look or redirecting us to some source that documents it?

binford2k: Apologies for missing this earlier. We're using "primary server" where we used to use MoM. So in HA you then have replicas of the primary. Compilers remain the same.

PR #771, opened 2020-10-21
Adds resubmit_facts config for puppet.conf

crazymind1337:
This adds the possibility to configure `resubmit_facts` via a parameter.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -187,6 +187,8 @@
 #
 # $agent_restart_command::                  The command which gets excuted on puppet service restart
 #
+# $resubmit_facts::                         Rerun the facts at the end of the puppet run.
+#
 # $client_package::                         Install a custom package to provide
 #                                           the puppet client
 #
@@ -455,7 +457,7 @@
 #
 # $server_puppetserver_experimental::       For Puppetserver 5, enable the /puppet/experimental route? Defaults to true
 #
-# $server_puppetserver_auth_template::      Template for generating /etc/puppetlabs/puppetserver/conf.d/auth.conf 
+# $server_puppetserver_auth_template::      Template for generating /etc/puppetlabs/puppetserver/conf.d/auth.conf
 #
 # $server_puppetserver_trusted_agents::     Certificate names of puppet agents that are allowed to fetch *all* catalogs
 #                                           Defaults to [] and all agents are only allowed to fetch their own catalogs.
@@ -600,6 +602,7 @@ class puppet (
   Hash[String, Data] $additional_settings = $puppet::params::additional_settings,
   Hash[String, Data] $agent_additional_settings = $puppet::params::agent_additional_settings,
   Optional[String] $agent_restart_command = $puppet::params::agent_restart_command,
+  Boolean $resubmit_facts = $puppet::params::resubmit_facts,
   String $classfile = $puppet::params::classfile,
   String $hiera_config = $puppet::params::hiera_config,
   String $auth_template = $puppet::params::auth_template,
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -21,6 +21,7 @@ class puppet::params {
   $systemd_cmd         = undef
 
   $agent_noop          = false
+  $resubmit_facts      = false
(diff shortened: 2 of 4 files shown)


manifests/params.pp
```diff
@@ -3,37 +3,38 @@
...
-  $prerun_command      = undef
-  $postrun_command     = undef
-  $server_compile_mode = undef
-  $dns_alt_names       = []
-  $use_srv_records     = false
+  $cron_cmd             = undef
+  $systemd_cmd          = undef
+
+  $agent_noop           = false
+  $resubmit_facts = false
```
ekohl, line 24: Please keep this aligned

manifests/params.pp
```diff
@@ -3,37 +3,38 @@
...
-  $ca_server           = undef
-  $ca_port             = undef
-  $ca_crl_filepath     = undef
-  $server_crl_enable   = undef
-  $prerun_command      = undef
-  $postrun_command     = undef
-  $server_compile_mode = undef
-  $dns_alt_names       = []
-  $use_srv_records     = false
+  $cron_cmd             = undef
```
ekohl, line 20: There's no need to realign

manifests/params.pp
wbclark, line 24: Thanks for opening this PR @crazymind1337 !

Could you fix the alignment for the new parameter here please?

Also, I don't see any reason why the alignment for the other params needs to change.

manifests/init.pp
```diff
@@ -600,6 +600,7 @@
   Hash[String, Data] $additional_settings = $puppet::params::additional_settings,
   Hash[String, Data] $agent_additional_settings = $puppet::params::agent_additional_settings,
   Optional[String] $agent_restart_command = $puppet::params::agent_restart_command,
+  Boolean $resubmit_facts = $puppet::params::resubmit_facts,
```
ekohl, line 603: Would need to be documented

ekohl (changes requested):
I wonder if this is really such a common thing. We try to only have all the common parameters in `init.pp`. Are you not able to keep this in your own manifest?

crazymind1337:
Wow that has been fast :D I will fix the mentioned issues.

manifests/params.pp
crazymind1337: I renamed the variable from previous `$agent_resubmit_facts` via `sed` and forgot to fix the alignment :(

crazymind1337: I renamed the variable from previous `$agent_resubmit_facts` via `sed` and forgot to fix the alignment :(

manifests/params.pp
crazymind1337: I renamed the variable from previous `$agent_resubmit_facts` via `sed` and forgot to fix the alignment :(

manifests/params.pp
crazymind1337: Fixed.

manifests/params.pp
crazymind1337: Fixed.

manifests/init.pp
crazymind1337: Fixed.

manifests/params.pp
crazymind1337: Fixed.

crazymind1337:
Do you still need some changes?

ekohl:
It's been busy with the Foreman 2.3.0 RC1 preparations, so a bit delayed on the response.

> I wonder if this is really such a common thing. We try to only have all the common parameters in `init.pp`. Are you not able to keep this in your own manifest?

I'd still like to get this question answered.

crazymind1337:
> I wonder if this is really such a common thing. We try to only have all the common parameters in `init.pp`. Are you not able to keep this in your own manifest?

It is a new feature for since puppet 6.11.0 (?). Yes it also could be done in our manifests.

ekohl:
> Yes it also could be done in our manifests.

I think this use case isn't that common and I think doing it in your manifests makes more sense.

crazymind1337:
Ok

PR #777, opened 2021-02-08
Add `puppet::package_install_options` variable to allow you to pass flags to package resource

vollmerk:
We needed this internally to add the `--diasableexclude=all` flag to the yum install, as puppet* is in the base "exclude=" statement to prevent unintentional upgrade of the puppet agent, by overzealous sysadmins who `yum -y update` I would expect this could be useful in any number of other situations where installation of the package requires specific options being passed to the installer. 

As far as I can tell all supported platforms default package management system supports the `install_options` flag.

### DIFF
--- a/manifests/agent/install.pp
+++ b/manifests/agent/install.pp
@@ -5,13 +5,15 @@ class puppet::agent::install(
   $package_name = $puppet::client_package,
   $package_version = $puppet::version,
   $package_provider = $puppet::package_provider,
+  $package_install_options = $puppet::package_install_options,
   $package_source = $puppet::package_source,
 ) {
   if $manage_packages == true or $manage_packages == 'agent' {
     package { $package_name:
-      ensure   => $package_version,
-      provider => $package_provider,
-      source   => $package_source,
+      ensure          => $package_version,
+      provider        => $package_provider,
+      install_options => $package_install_options,
+      source          => $package_source,
     }
   }
 }
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -140,6 +140,10 @@
 #                                           Defaults to undef. If 'windows' or 'msi' are
 #                                           used as the provider then this setting is
 #                                           required.
+# $package_install_options::                Flags that should be passed to the package manager
+#                                           during installation. Defaults to undef. May be
+#                                           a string or a hash, see Puppet Package resource
+#                                           documentation for the provider matching your package manager
 #
 # $unavailable_runmodes::                   Runmodes that are not available for the
 #                                           current system. This module will not try
@@ -563,6 +567,7 @@ class puppet (
   Optional[String] $dir_owner = $puppet::params::dir_owner,
   Optional[String] $dir_group = $puppet::params::dir_group,
   Optional[String] $package_provider = $puppet::params::package_provider,
+  Optional[Variant[String,Hash]] $package_install_options = $puppet::params::package_install_options,
   Optional[Variant[Stdlib::Absolutepath, Stdlib::HTTPUrl]] $package_source = $puppet::params::package_source,
   Integer[0, 65535] $port = $puppet::params::port,
   Boolean $pluginsync = $puppet::params::pluginsync,
--- a/spec/classes/puppet_agent_spec.rb
+++ b/spec/classes/puppet_agent_spec.rb
@@ -61,6 +61,7 @@ describe 'puppet' do
             .with_ensure('present')
             .with_provider(package_provider)
             .with_source(nil)
+            .with_install_options(nil)
         end
 
         # config
--- a/spec/classes/puppet_init_spec.rb
+++ b/spec/classes/puppet_init_spec.rb
@@ -35,7 +35,10 @@ describe 'puppet' do
         it { should_not contain_class('puppet::server') }
         it { should contain_file(puppet_directory).with_ensure('directory') }
         it { should contain_concat(puppet_concat) }
-        it { should contain_package(puppet_package).with_ensure('present') }
+        it { should contain_package(puppet_package)
+          .with_ensure('present')
+          .with_install_options(nil)
+        }
       end
 
       describe 'with server => true', :unless => unsupported_puppetmaster_osfamily(facts[:osfamily]) do
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -190,6 +190,7 @@ class puppet::params {
   }
 
   $package_source = undef
+  $package_install_options = undef
 
   # Need your own config templates? Specify here:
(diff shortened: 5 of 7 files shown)


manifests/init.pp
```diff
@@ -140,6 +140,10 @@
 #                                           Defaults to undef. If 'windows' or 'msi' are
 #                                           used as the provider then this setting is
 #                                           required.
+# $package_install_options::                Flags that should be passed to the installer
+#                                           during installation. Defaults to undef. May be
+#                                           a string or a hash, see Puppet Package resource
+#                                           documentation for more information
```
wbclark, line 146:
```suggestion
# $package_install_options::                Flags that should be passed to the package manager
#                                           during installation. Defaults to undef. May be
#                                           a string or a hash, see Puppet Package resource
#                                           documentation for the provider matching your package manager
```

I think this is a little clearer for users. In practice these modules are only supported on a few OSes so the actual number of package providers we support is small. Users would more quickly get an idea of what this does by going straight to the documentation for the `yum` Package provider or equivalent

wbclark (commented):
I'm looking for a way to add a simple test that no additional command line arguments are passed to the package manager with default options, to ensure we maintain that behavior going forward.

Looking at the spec tests I see for the agent we have a test for the `package` class at https://github.com/theforeman/puppet-puppet/blob/master/spec/classes/puppet_agent_spec.rb#L59-L64

I think we could add `.with_install_options(nil)` on a newline to that test. Could you try it out please?

wbclark:
For the server there is also a spec test for the `package` class at https://github.com/theforeman/puppet-puppet/blob/master/spec/classes/puppet_server_spec.rb#L69

In that case extending the same line so that it reads `it { should contain_package(puppetserver_pkg).with_install_options(nil) }` should suffice

wbclark:
Finally there is the main init spec with its `package` test for default params at https://github.com/theforeman/puppet-puppet/blob/master/spec/classes/puppet_init_spec.rb#L38

We can add coverage here by replacing that line with:

```
        it { should contain_package(puppet_package)
          .with_ensure('present')
          .with_install_options(nil)
        }
```

manifests/agent/install.pp
```diff
@@ -5,13 +5,15 @@
...
 ) {
   if $manage_packages == true or $manage_packages == 'agent' {
     package { $package_name:
-      ensure   => $package_version,
-      provider => $package_provider,
-      source   => $package_source,
+      ensure            => $package_version,
+      provider          => $package_provider,
+      install_options   => $package_install_options,
+      source            => $package_source,
```
wbclark, line 16:
```suggestion
      ensure          => $package_version,
      provider        => $package_provider,
      install_options => $package_install_options,
      source          => $package_source,
```

To fix this test breakage:

```
manifests/agent/install.pp:13:arrow_alignment:WARNING:indentation of => is not properly aligned (expected in column 23, but found it in column 25)
manifests/agent/install.pp:14:arrow_alignment:WARNING:indentation of => is not properly aligned (expected in column 23, but found it in column 25)
manifests/agent/install.pp:15:arrow_alignment:WARNING:indentation of => is not properly aligned (expected in column 23, but found it in column 25)
manifests/agent/install.pp:16:arrow_alignment:WARNING:indentation of => is not properly aligned (expected in column 23, but found it in column 25)
```

wbclark:
Thanks @vollmerk for the submission!

I gave a bit of feedback about simple tests to ensure this remains undefined by default going forward.

Please take a look and let me know if you have any question or comment.

Looking forward to hearing from you,

vollmerk:
> Thanks @vollmerk for the submission!
> 
> I gave a bit of feedback about simple tests to ensure this remains undefined by default going forward.
> 
> Please take a look and let me know if you have any question or comment.
> 
> Looking forward to hearing from you,

Implemented your suggestions, thanks. Let me know if I missed anything.

spec/classes/puppet_agent_spec.rb
```diff
@@ -61,6 +61,7 @@
             .with_ensure('present')
             .with_provider(package_provider)
             .with_source(nil)
+	    .with_install_options(nil)
```
wbclark, line 64:
```suggestion
            .with_source(nil)
            .with_install_options(nil)
```

It looks correct in my editor but GH shows the alignment as off. Converting tabs to spaces here should fix that :)

spec/classes/puppet_init_spec.rb
```diff
@@ -35,7 +35,10 @@
         it { should_not contain_class('puppet::server') }
         it { should contain_file(puppet_directory).with_ensure('directory') }
         it { should contain_concat(puppet_concat) }
-        it { should contain_package(puppet_package).with_ensure('present') }
+        it { should contain_package(puppet_package)
+	  .with_ensure('present')
+	  .with_install_options(nil)
+	}
```
wbclark, line 41:
```suggestion
        it { should contain_package(puppet_package)
          .with_ensure('present')
          .with_install_options(nil)
        }
```

Same issue as above

wbclark (commented):
Thanks @vollmerk 

Overall it looks good to me. I left some minor feedback on the indentation, I think we're good to move forward once that is fixed up

The failures on the EL6 acceptance tests look to me like they are unrelated to this PR. I will investigate what could be causing that on our end so that we can be confident everything should be nice and :green_apple:

vollmerk:
> Thanks @vollmerk
> 
> Overall it looks good to me. I left some minor feedback on the indentation, I think we're good to move forward once that is fixed up
> 
> The failures on the EL6 acceptance tests look to me like they are unrelated to this PR. I will investigate what could be causing that on our end so that we can be confident everything should be nice and 🍏

Added the white-space fixes, and re-squashed the commits...

wbclark:
The reason for the remaining test failure is the unsurprising fact that there no longer exists any centos6 image at https://vagrantcloud.com/centos

vollmerk:
> The reason for the remaining test failure is the unsurprising fact that there no longer exists any centos6 image at https://vagrantcloud.com/centos

Let me know if you need me to make any more adjustments on this relating to those checks, or if ya'll will take care of them in a different pull req

wbclark:
Hi @vollmerk , I don't think anything further is required on your end.

I opened https://github.com/voxpupuli/beaker-hostgenerator/pull/203 just now to resolve the upstream issue which was breaking the EL6 acceptance tests.

I'll use that to test against your branch locally in a moment.

wbclark (approved):
As expected, EL6 acceptance tests are passing when I run them locally with a working EL6 image

wbclark:
Thanks @vollmerk !

Issue #780, opened 2021-03-22
`puppet::server_puppetserver_metrics: false` causes `puppetserver` to fail


kBite:
Initially I wanted to disable the Ruby profiler. With release 8.0.0 `$server_enable_ruby_profiler` was merged with `$server_puppetserver_metrics`. But when configuring `puppet::server_puppetserver_metrics: false` the module removed `metrics.conf`.

Restoring that file allows `puppetserver` to start again.

As a workaround I modified this module _not_ to remove `metrics.conf`, but I'm sure that's not the best way to solve this - just the easiest for me.

Using
- puppet-puppet **14.2.0**
- puppetserver **6.15.1**

puppetserver
```
class { 'puppet':
  agent                                  => true,
  autosign_entries                       => $autosign_entries,
  ca_server                              => $facts['fqdn'],
  environment                            => $facts['environment'],
  puppetmaster                           => $facts['fqdn'],
  server                                 => true,
  server_allow_header_cert_info          => true,
  server_ca                              => true,
  server_ca_allow_auth_extensions        => true,
  server_ca_allow_sans                   => true,
  server_ca_client_self_delete           => true,
  server_check_for_updates               => false,
  server_common_modules_path             => '',
  server_environment_class_cache_enabled => true,
  server_foreman                         => true,
  server_foreman_url                     => "https://${facts['fqdn']}",
  server_http                            => true,
  server_ip                              => 'localhost',
  server_jvm_max_heap_size               => "${cpu_count_2x}G",
  server_jvm_min_heap_size               => "${cpu_count_1x}G",
  server_max_active_instances            => $cpu_count_2x,
  server_max_queued_requests             => $cpu_count_4x,
  server_max_requests_per_instance       => 1000000,
  server_metrics_graphite_enable         => false,
  server_metrics_server_id               => 'puppet.local',
  server_multithreaded                   => true,
  server_port                            => 8141,
  server_puppetserver_experimental       => false,
  server_puppetserver_metrics            => false,
  server_puppetserver_version            => $version,
  server_reports                         => 'puppetdb,foreman',
  server_ruby_load_paths                 => [
    '/opt/puppetlabs/puppet/lib/ruby/vendor_ruby',
    '/opt/puppetlabs/puppet/cache/lib'
  ],
  server_storeconfigs                    => true,
  server_strict_variables                => true,
  show_diff                              => true,
  splay                                  => true,
  server_jvm_extra_args                  => [
[...]

PR #779, opened 2021-03-08
Support Puppet 7

ekohl:
https://puppet.com/docs/puppet/7.4/server/release_notes.html

* [x] Handle cadir move to `/etc/puppetlabs/puppetserver/ca`
* [x] Handle auth.conf removal
* [x] Handle JSON fact cache instead of YAML (as used by Foreman's ENC)
* [x] Look at any legacy Puppet 3 endpoints
* [x] Align default ciphers

PR #781, opened 2021-03-24
(#780) puppetserver: don't purge metrics.conf / Use dedicated parameter for jruby profiler 

bastelfreak:
This introduces a new parameter to disable the profiling indepently from
the metrics. Previously this was enabled by default. However Puppet Inc.
disables it in their standard configuration and the profiler causes a
significate memory and CPU overhead so I guess we should stick with the
upstream default.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -437,8 +437,8 @@
 #
 # $server_puppetserver_jruby9k::            For Puppetserver 5, use JRuby 9k? Defaults to false
 #
-# $server_puppetserver_metrics::            Enable metrics (Puppetserver 5.x only)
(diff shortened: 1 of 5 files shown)


bastelfreak:
that's my (currently untested) approach to fix #780 . Let me know what you think / if you like two PRs for this.

ekohl:
Other than the failing unit tests, I think this looks good. I already started to change the odd `Optional[Boolean]` in https://github.com/theforeman/puppet-puppet/pull/779 anyway.

ekohl (approved):
Looks good to me, let's see if the test agree.

ekohl (approved):
Test failures look unrelated, merging.

PR #782, opened 2021-03-25
Fix config append

### LINKED ISSUE #759: Appending values with puppet::config::entry is broken

SimonPe:
This should fix #759.
I start by reverting #707 and re-implementing a fix for it. the actual fix was actually quiet easy, just by adding a concat snippet at the end of each config block with a newline.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -73,6 +73,12 @@ class puppet::config(
     puppet::config::main { $key: value => $value }
   }
 
+  concat::fragment { 'puppet.conf_comment':
+    target  => "${puppet_dir}/puppet.conf",
+    content => '# file managed by puppet',
+    order   => '0_comment',
+  }
+
   file { $puppet_dir:
     ensure => directory,
     owner  => $puppet::dir_owner,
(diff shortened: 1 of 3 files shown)


manifests/config.pp
```diff
@@ -73,6 +73,12 @@
     puppet::config::main { $key: value => $value }
   }
 
+  concat::fragment { 'puppet.conf_comment':
+    target  => "${puppet_dir}/puppet.conf",
+    content => '# file managed by puppet',
```
ekohl, line 78: This is a nice solution to the otherwise empty line.

ekohl (approved):
Tested on me own infra and this works. Thanks!

PR #785, opened 2021-04-26
Drop EL6 support

wbclark:


manifests/params.pp
```diff
@@ -303,17 +303,11 @@
...
-      # Amazon Linux is like RHEL6 but reports its osreleasemajor as 2017 or 2018.
       $agent_restart_command = $facts['os']['release']['major'] ? {
-        /^(2|5|6|2017|2018)$/ => "/sbin/service ${service_name} reload",
         '7'       => "/usr/bin/systemctl reload-or-restart ${service_name}",
         default   => undef,
       }
       $unavailable_runmodes = $facts['os']['release']['major'] ? {
-        /^(2|5|6|2017|2018)$/ => ['systemd.timer'],
         default   => [],
       }
```
ekohl, line 312: If you have just a single result, this is equivalent
```suggestion
      $unavailable_runmodes = []
```

ekohl (changes requested):

alexjfisher:
Considering Puppet Inc are releasing Puppet 7 packages for EL6 and this module hasn't had a version out yet with Puppet 7 support, I'd be in favour of keeping EL6 support (for agents) for the time being.

If this goes through, I won't be able to upgrade to Puppet 7 until every last CentOS 6 box has been decommissioned?

Or perhaps it is just going to be a case of having to manually set a few parameters and EL6 *might* continue to work for a few more releases even if not officially supported??

ehelms:
I think this popped up in part because tests are failing when trying to do a new release -- https://github.com/theforeman/puppet-puppet/pull/787 

So the question I have is, if we keep EL6 support do we just ignore those tests for now? I don't want to break your use case @alexjfisher we just need to define our strategy so we can get a new release out in the next day or two.

ekohl:
I would be fine with ignoring failing tests for now.

ehelms:
Now that we have a new release, should we consider pulling this in? If not, where should we align this change to in the roadmap of this module?

wbclark:
> Now that we have a new release, should we consider pulling this in? If not, where should we align this change to in the roadmap of this module?

It was brought up above that it would be nice to have Puppet 7 support before dropping EL6. Puppet 7 is packaged for EL6 and that gives a window to complete the upgrade to Puppet 7 first before decommissioning EL6.

For that, at least https://github.com/theforeman/puppet-puppet/pull/788 and https://github.com/theforeman/puppet-puppet/pull/779 are needed. @alexjfisher are you able to help with reviews in this area ?

ekohl:
Closing in favor of https://github.com/theforeman/puppet-puppet/pull/823.

PR #786, opened 2021-04-26
El8 agent restart cmd

wbclark:
To make things easier later on, I based this on https://github.com/theforeman/puppet-puppet/pull/785 which is touching cases of the same param

If for whatever reason it is desired to do one but not the other, let me know and I can make the necessary changes


manifests/params.pp
```diff
@@ -303,17 +303,11 @@
       $unavailable_runmodes = []
     }
     'Redhat' : {
-      # PSBM is a CentOS 6 based distribution
-      # it reports its $osreleasemajor as 2, not 6.
-      # thats why we're matching for '2' in both parts
-      # Amazon Linux is like RHEL6 but reports its osreleasemajor as 2017 or 2018.
       $agent_restart_command = $facts['os']['release']['major'] ? {
```
ekohl, line 306: I'd go a step further and assume systemd (once EL6 has been dropped). That means I'd just remove the whole case statement and reduce it to `$agent_restart_command = "/usr/bin/systemctl reload-or-restart ${service_name}"`

ekohl (changes requested):

ekohl:
Since 317f0b0ebfa212182b23b4326655cdf8c139ca8a this is now already done.

PR #787, opened 2021-04-26
Release 15.0.0

wbclark:


wbclark:
While we are doing a major release, it would be nice to merge https://github.com/theforeman/puppet-puppet/pull/785 which is also a breaking change

At that point it would become convenient to get https://github.com/theforeman/puppet-puppet/pull/786 as well

Please send a ping if any rebases are needed :)

ehelms (approved):

PR #788, opened 2021-05-03
Allow Puppet 7 compatible versions of mods

ekohl:


timdeluxe:
@ekohl Why did you not raise systemd dependency?

ekohl:
Because that's really a breaking change and I'm still looking at that. The title was a bit misleading, but it's on my radar.

timdeluxe:
Okay, we need that to fit in dependencies. I will fork, raise by myself, but not raise a PR, because it is on your radar. However i can report if we experience any issues due to that, if you are interested.

ekohl:
That should be https://github.com/theforeman/puppet-puppet/pull/790. I'll probably prepare the PRs for all other modules prior to merging it and verify our installer still works completely.

PR #790, opened 2021-05-26
Allow camptocamp/systemd 3.x + use it for timers

ekohl:
This allows camptocamp/systemd 3.x. Additionally, it goes back to using systemd::timer now that the cyclic dependency can no longer happen.


ekohl:
https://github.com/theforeman/puppet-pulpcore/pull/198 & https://github.com/theforeman/puppet-foreman/pull/956 are the other 2 modules that need to be merged for the installer to be able to use 3.x.

ehelms (approved):

PR #791, opened 2021-06-01
Drop Puppet 5 support

ehelms:


ekohl (changes requested):
I think dropping Puppet 5 should also drop all code. That's a much bigger patch.

alexjfisher:
If you're going to remove code, maybe just remove it for the puppetserver side of the module, but leave the agent code alone?
Puppetserver 6 still supports Puppet 3 agents, and I think this module still works with those.  Removing support can make it harder to upgrade.

After a lot of work, I've just upgraded by server infra to Puppet 6/Foreman 2.  Most agents are Linux and are still on Puppet 4, (which I'll be upgrading to 6 in the next few weeks), but I've also got some Solaris 10 and 11 with Puppet 3 agents. 😢

ehelms:
We can just leave this be for now, it was inspired by dropping the support across our modules since their largest use case is through the foreman-installer and cleaning up code paths and test matrices. Also, we had a few modules starting to need Puppet 6+ features (e.g. EL8 module support).

ekohl:
@alexjfisher how far along are you? I'd like to reconsider this because by now Puppet 5 AIO packages are no longer on yum.puppet.com so we no longer really test it.

ekohl:
I've merged #822.

PR #801, opened 2021-09-01
Implement optional Katello integration

ekohl:
Katello uses a different certificate structure. This moves over the integration bits from puppet-foreman_proxy_content to this module. It also means fewer variables need to be set in the installer itself.

This is still very early (needs a related change in puppet-foreman_proxy_content and an installer migration), but submitting for a general design idea.

### DIFF
--- a/manifests/server/foreman.pp
+++ b/manifests/server/foreman.pp
@@ -1,0 +1,18 @@
+# @summary Set up Foreman integration
+# @api private
+class puppet::server::foreman {
+  # Include foreman components for the puppetmaster
+  # ENC script, reporting script etc.
+  class { 'puppetserver_foreman':
+    foreman_url      => $puppet::server::foreman_url,
+    enc_upload_facts => $puppet::server::server_foreman_facts,
+    enc_timeout      => $puppet::server::request_timeout,
+    puppet_home      => $puppet::server::puppetserver_vardir,
+    puppet_basedir   => $puppet::server::puppet_basedir,
+    puppet_etcdir    => $puppet::dir,
+    ssl_ca           => pick($puppet::server::foreman_ssl_ca, $puppet::server::ssl_ca_cert),
+    ssl_cert         => pick($puppet::server::foreman_ssl_cert, $puppet::server::ssl_cert),
+    ssl_key          => pick($puppet::server::foreman_ssl_key, $puppet::server::ssl_cert_key),
+  }
+  contain puppetserver_foreman
+}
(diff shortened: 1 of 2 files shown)


manifests/server/foreman.pp
```diff
@@ -0,0 +1,33 @@
...
+    include certs::puppet
+    Class['certs::puppet'] -> Class['puppetserver_foreman']
+
+    $ssl_ca = $certs::puppet::ssl_ca_cert
+    $ssl_cert = $certs::puppet::client_cert
+    $ssl_key = $keys::puppet::client_key
+  } else {
+    $ssl_ca = $puppet::server::ssl_ca_cert
+    $ssl_cert = $puppet::server::ssl_cert
+    $ssl_key = $puppet::server::ssl_cert_key
```
ehelms, line 16: I am wondering if this should now be:

```
    $ssl_ca = pick($puppet::server::foreman_ssl_ca, $puppet::server::ssl_ca_cert)
    $ssl_cert = pick($puppet::server::foreman_ssl_cert, $puppet::server::ssl_cert)
    $ssl_key = pick($puppet::server::foreman_ssl_key, $puppet::server::ssl_cert_key)
```

To prevent a user from setting `katello=true` and then overriding it by setting those parameters and ending up in a weird state given the `katello=true` is an explicit request. Then down below it would become:

```
    ssl_ca           => $ssl_ca,
    ssl_cert         => $ssl_cert,
    ssl_key          => $ssl_key,
```

manifests/server/foreman.pp
```diff
@@ -0,0 +1,33 @@
+# @summary Set up Foreman integration
+# @api private
+class puppet::server::foreman (
+  Boolean $katello = false,
+) {
+  if $katello {
+    include certs::puppet
```
ehelms, line 7: Does this imply https://github.com/theforeman/puppet-puppet/blob/master/metadata.json#L17 now needs to require `katello-certs` or do we allow 'implied' or 'if present' dependencies?

ekohl: We already have puppet-puppetserver_foreman as a soft dependency: https://github.com/theforeman/puppet-puppet#foreman-integration

manifests/server/foreman.pp
ekohl: I debated this. On the other hand, we have the parameter and if a user does want to override it ... Still not 100% sure where I land on it.

ehelms: I think my proposed change still allows overriding it, users would just set `katello=false` and provide them and prevent the whole need to generate unnecessary certs and relationships.

ehelms:
This is looking good. What are the transition steps across our modules to enable this method and drop the puppet-fpc method?

ekohl:
Closing for now, we'll likely go in a different direction where Puppetserver will only talk to a Smart Proxy.

PR #803, opened 2021-09-20
Support multiple envs_dir directories

### LINKED ISSUE #708: Support multiple path in environmentpath parameter
Currently only one path is accepted. puppet.conf is correctly updated but it fails on folder creation. We probably need a loop at https://github.com/theforeman/puppet-puppet/blob/master/manifests/server/config.pp#L225 

ref: https://puppet.com/docs/puppet/6.4/environments_creating.html#environmentpath

optiz0r:
This module adds support for the puppet environmentpath setting to be configured
using multiple directories. An example use-case is where r10k is in use to
deploy some but not all environments, to avoid unmanaged environments being
purged by r10k.

- Each listed environmentpath directory will be created and managed by puppet
- The default post-receive hook script uses the first listed directory in
  `envs_dir` to create initial environments to maintain backward compatibility.
- The config_version_cmd is updated to use only the first listed directory from
  `envs_dir`, this is only used in the environment.conf.erb template which does
  not appear to be deployed anywhere in the current version of this module

Tests are included to ensure that all listed directories are created and
managed, and that any other references to the envs_dir parameter behave the
same as if only a single directory were specified to maintain backward
compatibility.

Fixes #708

### DIFF
--- a/manifests/server/config.pp
+++ b/manifests/server/config.pp
@@ -36,6 +36,7 @@ class puppet::server::config inherits puppet::config {
   $server_external_nodes       = $puppet::server::external_nodes
   $server_environment_timeout  = $puppet::server::environment_timeout
   $trusted_external_command    = $puppet::server::trusted_external_command
+  $primary_envs_dir            = $puppet::server::envs_dir[0]
 
   if $server_external_nodes and $server_external_nodes != '' {
     class{ 'puppet::server::enc':
@@ -56,7 +57,7 @@ class puppet::server::config inherits puppet::config {
 
   puppet::config::main {
     'reports':            value => $puppet::server::reports;
-    'environmentpath':    value => $puppet::server::envs_dir;
+    'environmentpath':    value => $puppet::server::envs_dir.join(':');
   }
   if $puppet::server::hiera_config and !empty($puppet::server::hiera_config){
     puppet::config::main {
@@ -258,7 +259,7 @@ class puppet::server::config inherits puppet::config {
       mode    => $puppet::server::git_repo_mode,
       user    => $puppet::server::git_repo_user,
       group   => $puppet::server::git_repo_group,
-      require => File[$puppet::vardir, $puppet::server::envs_dir],
+      require => File[$puppet::vardir, $primary_envs_dir],
     }
 
     $git_branch_map = $puppet::server::git_branch_map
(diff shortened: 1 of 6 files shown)


ekohl (changes requested):
I think keeping the data type of `envs_dir` on `Stdlib::Absolutepath` is misleading. If it's an array, really make it an array: `Variant[Array[Stdlib::Absolutepath, 1], Stdlib::Absolutepath]`.

optiz0r:
> I think keeping the data type of `envs_dir` on `Stdlib::Absolutepath` is misleading. If it's an array, really make it an array: `Variant[Array[Stdlib::Absolutepath, 1], Stdlib::Absolutepath]`.

A wrinkle: this is going to need corresponding changes to the `foreman-installer` too, which currently takes a single string parameter and passes it straight through.

ekohl:
I'm very close to releasing a major version of the module. Would you like to update this so it's an array? I can write the migration on the installer side.

optiz0r:
@ekohl Do you mean switch from `Variant[Array[Stdlib::Absolutepath, 1], Stdlib::Absolutepath]` to just `Array[Stdlib::Absolutepath, 1]`? If so yes, I can do that this weekend.

ekohl:
That is indeed what I mean. I can wait with a major version until Monday.

optiz0r:
Updated to drop taking a single path as a string (breaking change), and rebased against current master.

manifests/server/config.pp
```diff
@@ -230,13 +231,15 @@
...
+  $puppet::server::envs_dir.each |$dir| {
+    file { $dir:
+      ensure => $ensure,
+      owner  => $puppet::server::environments_owner,
+      group  => $puppet::server::environments_group,
+      mode   => $puppet::server::environments_mode,
+      target => $puppet::server::envs_target,
+      force  => true,
+    }
   }
```
ekohl, line 243: You don't need to modify this since you can pass an array to `file`.

ekohl (changes requested):
Very minor thing, other than that :+1:

ekohl:
I think https://github.com/theforeman/foreman-installer/pull/740 would be the installer migration.

optiz0r:
Will fix in the morning. This is leftover from handling the more complex
variant case. I will revert this hunk and force push.

ekohl (approved):
Thanks!

PR #805, opened 2021-09-21
Drop config_version and environment.conf

optiz0r:
Support for dynamic and config environments was dropped in a61e010
however the `config_version` related parameters and the
`environment.conf` template used by these features were not removed
at the same time. This commit drops the unused code.

As parameters are being removed, this will be a breaking change and
require a major version bump.


ekohl:
The unit test failures are odd. I submitted https://github.com/voxpupuli/puppet-systemd/pull/229 which I think should fix them.

ekohl (approved):
Test suite is not complete yet, but I think the failures that do show up are unrelated.

ekohl:
Hmm, looks like the patch was wrong and I need to look at why it's failing.

PR #806, opened 2021-09-22
Support "environment-vars" setting in puppetserver.conf

### LINKED ISSUE #792: Support "environment-vars" setting in puppetserver.conf
The puppetserver.conf has a [jruby-puppet setting "environment-vars"](https://puppet.com/docs/puppet/7/server/config_file_puppetserver.html) which allows the puppet code to access defined environment variables. This option is currently not supported. 

A use-case is defining environment variables in `/etc/default/puppetserver` and making it accessible to puppet ruby functions.

I can create a PR if you agree to support this setting.

phihos:
Fixes #792


phihos:
@ekohl Is it possible to get this reviewed?

manifests/init.pp
```diff
@@ -699,6 +703,7 @@
   Optional[Variant[String,Array[String]]] $server_jvm_extra_args = $puppet::params::server_jvm_extra_args,
   Optional[String] $server_jvm_cli_args = $puppet::params::server_jvm_cli_args,
   Optional[Stdlib::Absolutepath] $server_jruby_gem_home = $puppet::params::server_jruby_gem_home,
+  Hash[String, String] $server_environment_vars = $puppet::params::server_environment_vars,
```
ekohl, line 706: Thinking out loud: Do we also want to accept `Integer` here? Users may write in hiera:
```yaml
puppet::server_environment_vars:
  myvar: 1
```

I think in Puppet this ends up as an Integer unless they write `"1"`.

templates/server/puppetserver/conf.d/puppetserver.conf.erb
```diff
@@ -73,6 +73,20 @@ jruby-puppet: {
...
+    # Ruby code running within JRuby, for example, via the Ruby ENV class.
+    # By default, the only environment variables whose values are set into JRuby from the shell are HOME and PATH.
+    # The default value for the GEM_HOME environment variable in JRuby is set from the value provided for the
+    # jruby-puppet.gem-home key.
+    # Any variable set from the map for the environment-vars key overrides these defaults.
+    # Avoid overriding HOME, PATH, or GEM_HOME here because these values are already configurable via the shell
+    # or jruby-puppet.gem-home.
+    environment-vars: {
+    <%- @server_environment_vars.each do |env_key, env_val| -%>
+      "<%= env_key %>" : <%= env_val %>
```
ekohl, line 87: Is it needed to quote the value?

manifests/init.pp
phihos: I was thinking from a technical perspective that environment variables are always strings by nature. But we can allow more types I you like and rely on implicit conversion during templating.

templates/server/puppetserver/conf.d/puppetserver.conf.erb
phihos: The more I read about HOCON the more complicated it gets. My use case is to set the following
```hocon
  ...
  environment-vars: { 
    "FOO" : ${FOO}
  }
  ...
```
`${FOO}` will be replaced by the corresponding environment variable making it possible to pass through variables set in `/etc/default/puppetserver`. Quoting it in any way prevents substitutions from being evaluated.
So I guess we somehow must let the user decide whether to quote the value or not. So how about this:

We render the value verbatim into the config and adapt the docs to show that the following is possible:
```yaml
puppet::server_environment_vars:
  myvar: 'some string'
  myvar_from_env: '${myvar_from_env}'
  myvar_quoted: '"somestring with forbidden chars $[]{}"'
  myvar_multiline: |-
    """this
    is
    a
    multi-line
    string
    """
````
Would that be ok?

phihos: I finally found the time to improve the variable doc comment. Can you take a look?

phihos: @ekohl Any update on this?

hugendudel:
Hi guys, I need this feature too. It would be really great if it continues here.

Thanks and greetings
Harm

templates/server/puppetserver/conf.d/puppetserver.conf.erb
phihos: @ekohl Can you take another look?

ekohl (approved):
Thanks!

PR #807, opened 2021-09-27
Remove Puppet start workaround on Debian

ekohl:
Back with Puppet 3 there was a workaround to not start Puppet by default but since Debian Jessie ([854487]) this no longer exists. It now uses regular init integration. This also doesn't affect AIO which never implemented it this way.

[854487]: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=854487


manifests/agent/config.pp
```diff
@@ -48,24 +48,4 @@
...
-  }
-
-  if $facts['os']['family'] == 'Debian' {
-    augeas {'puppet::set_start':
-      context => '/files/etc/default/puppet',
-      changes => "set START ${should_start}",
-      incl    => '/etc/default/puppet',
-      lens    => 'Shellvars.lns',
-    }
-    if $puppet::remove_lock {
```
wbclark, line 65: `$puppet::remove_lock` is no longer used anywhere with this change, so the param should be removed entirely from init.pp and params.pp

ekohl: Good point.

ehelms (approved):

manifests/init.pp
```diff
@@ -616,7 +614,6 @@
   Array[String] $auth_allowed = $puppet::params::auth_allowed,
   Variant[String, Array[String]] $client_package = $puppet::params::client_package,
   Boolean $agent = $puppet::params::agent,
-  Boolean $remove_lock = $puppet::params::remove_lock,
```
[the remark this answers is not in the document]
ekohl: Correct so I added the `Backwards incompatible` label to it.

PR #808, opened 2021-09-29
Stop coercing Strings into Numerical

bassonj:
During Puppet run warnings are presented that the string value is coerced into a numeric.
This change addresses that warning


ekohl:
@bassonj can we close this or can you provide a reproducer?

PR #818, opened 2021-12-27
Improve support for chocolatey package provider on Windows

fraenki:
Currently the allowed data types for parameter `$package_install_options` do not include arrays. However, when using the chocolatey package provider on Windows, arrays should be used:
https://forge.puppet.com/modules/puppetlabs/chocolatey/readme#packages


ekohl (approved):
Our CI is currently broken (I think it's Pysch 4.0) but I can't see how this would break things. Thanks!

PR #822, opened 2022-02-03
Drop Puppet 5 support, require >= 6.15.0 + drop pluginsync & jruby9k parameters

ekohl:
Replaces https://github.com/theforeman/puppet-puppet/pull/791.

### DIFF
--- a/manifests/agent/config.pp
+++ b/manifests/agent/config.pp
@@ -35,16 +35,6 @@ class puppet::agent::config inherits puppet::config {
     }
   }
 
-  unless $puppet::pluginsync {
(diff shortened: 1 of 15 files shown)


ekohl:
Right now the failing tests:
* Puppet 6 unit tests has a problem where Psych 4 is in the gem cache and we should purge it
* CentOS 6 fails because it's no longer on the mirrors
* Ubuntu fails because the lsbdistrelease fact is empty so the package version is incomplete and it can't install specific versions

These all also fail on master so I wouldn't consider them blocking right now, but I'll look into the last one.

bastelfreak (approved):
I haven't tested this locally, but it looks sane to me

PR #825, opened 2022-02-21
Change `server_puppetserver_profiler` and `server_puppetserver_metrics` defaults to true

alexjfisher:
Puppet Inc. enabled the profiler by default in Puppetserver 5.0.0
https://tickets.puppetlabs.com/browse/SERVER-1761

This module has been disabling by default though causing users to be
missing metrics they might reasonably expect to be present after reading
through the [Puppetserver documentation](https://puppet.com/docs/puppet/6/server/puppet_server_metrics.html#compiler-metrics).

**EDIT**
Also changed `server_puppetserver_metrics` (which enables http-client metrics) to `true`

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -439,10 +439,10 @@
 #                                           Defaults to 30000, using the Jetty default of 30s
 #
 # $server_puppetserver_metrics::            Enable puppetserver http-client metrics
-#                                           Defaults to false because that's the Puppet Inc. default behaviour.
+#                                           Defaults to true, matching defaults in Puppetserver 5+.
 #
 # $server_puppetserver_profiler::           Enable JRuby profiling.
-#                                           Defaults to true because that's the Puppet Inc. default behaviour (since Puppetserver 5.0.0).
+#                                           Defaults to true, matching defaults in Puppetserver 5+.
 #                                           If set to false, compiler and function metrics will not be available, (eg. when enabling graphite metrics)
 #
 # $server_metrics_jmx_enable::              Enable or disable JMX metrics reporter. Defaults to true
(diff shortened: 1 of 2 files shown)


ekohl:
In 12154ce68e51a4e83eb26cea7867a2153bf370d8 it was split into two different parameters.

If metrics depend on the profiler, should we make it catalog compilation fail on it?

alexjfisher:
Only _some_ metrics depend on this setting.  eg. the compiler and function metrics.  With the profiler off, you still get the jvm and jruby metrics etc.

manifests/init.pp
```diff
@@ -439,10 +439,11 @@
 #                                           Defaults to 30000, using the Jetty default of 30s
 #
 # $server_puppetserver_metrics::            Enable puppetserver http-client metrics
-#                                           Defaults to false because that's the Puppet Inc. default behaviour.
+#                                           Defaults to true because that's the Puppet Inc. default behaviour (since Puppetserver 5.0.0).
```
ekohl, line 442: I wonder if we even need to mention the default at all, but perhaps shorten it?
```suggestion
#                                           Defaults to true, matching defaults in Puppetserver 5+.
```

manifests/init.pp
```diff
@@ -439,10 +439,11 @@
...
 #
 # $server_puppetserver_profiler::           Enable JRuby profiling.
-#                                           Defaults to false because that's the Puppet Inc. default behaviour.
+#                                           Defaults to true because that's the Puppet Inc. default behaviour (since Puppetserver 5.0.0).
```
ekohl, line 445: Same comment here about the defaults.

ekohl (approved):
Overall :+1: but some bikeshedding about the description.

ekohl:
Also, could you rebase? I'd expect the tests to pass then.

manifests/init.pp
alexjfisher: Updated and rebased.

ekohl (approved):
Thanks!

PR #826, opened 2022-03-04
Add `server_jolokia_metrics_whitelist` parameter

alexjfisher:
When set, allows listed clients access to the V2 (jolokia) metrics
endpoint.

See https://puppet.com/docs/puppet/7/server/metrics-api/v2/metrics_api.html

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -537,6 +537,9 @@
 #                                           invokes when on static_file_content requests.
 #                                           Defaults to undef
 #
+# $server_jolokia_metrics_whitelist::       The whitelist of clients that
+#                                           can query the jolokia /metrics/v2 endpoint
+#
 # === Usage:
 #
 # * Simple usage:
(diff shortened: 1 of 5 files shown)


alexjfisher:
Should I also accept a special value that turns on unauthenticated access, instead of populating the `allow` section?

ie 
```puppet
Variant[Enum['NOAUTH'],Array[String[1]]]
```

Or maybe also accept `true`??
```puppet
Variant[Boolean, Array[String[1]]]
```

Or maybe a whitelist is overly complicated and I should have gone for a simple boolean!

manifests/init.pp
```diff
@@ -537,6 +537,10 @@
 #                                           invokes when on static_file_content requests.
 #                                           Defaults to undef
 #
+# $server_jolokia_metrics_whitelist::       The whitelist of clients that
+#                                           can query the jolokia /metrics/v2 endpoint
+#                                           Defaults to []
```
ekohl, line 542: IMHO there's no need to document the default.

ekohl (approved):
I'm OK with this. I haven't used the metrics myself so I'll trust you on the specifics. The defaults don't change anything so :+1: 

Do we want to drop the comment describing the default?

ekohl (approved):

PR #828, opened 2022-03-15
Disable FIPS support within JVM for Puppet

ehelms:
Puppet does not currently support FIPS on EL8 and needs to be disabled in order to run on a FIPS enabled EL8+ machine within the Java stack. This solution includes the disable flag out right as it does not break on environments where the flag is not present.

A couple questions to consider:

 a) should this be a configurable parameter?
 b) should this only be included conditionally? (complicates the code a little bit)

### DIFF
--- a/manifests/server/puppetserver.pp
+++ b/manifests/server/puppetserver.pp
@@ -60,6 +60,9 @@
 # @param server_multithreaded
 #   Configures the puppetserver to use multithreaded jruby.
 #
+# @param disable_fips
+#   Disables FIPS support within the JVM
+#
 # @example
 #
 #   # configure memory for java < 8
@@ -140,6 +143,7 @@ class puppet::server::puppetserver (
   $max_open_files                         = $puppet::server::max_open_files,
   $versioned_code_id                      = $puppet::server::versioned_code_id,
   $versioned_code_content                 = $puppet::server::versioned_code_content,
+  $disable_fips                           = $facts['os']['family'] == 'RedHat' and $facts['os']['release']['major'] == '8',
 ) {
   include puppet::server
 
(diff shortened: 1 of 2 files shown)


ekohl (changes requested):
I think we discussed this should be limited to EL8. I'd still prefer a class parameter:
```puppet
Boolean $disable_fips = $facts['os']['family'] == 'RedHat' && $facts['os']['release']['major'] == '8',
```

That way it can at least be modified via Hiera.

I also expect that the tests will fail on this.

evgeni:
Does the JVM on EL7 even know about that parameter, or would it ignore it? No need to make it conditional then ;)

manifests/server/puppetserver.pp
```diff
@@ -149,7 +153,11 @@
 
   $puppetserver_package = pick($puppet::server::package, 'puppetserver')
 
-  $jvm_cmd_arr = ["-Xms${jvm_min_heap_size}", "-Xmx${jvm_max_heap_size}", $jvm_extra_args]
+  if $disable_fips {
+    $jvm_cmd_arr = ["-Xms${jvm_min_heap_size}", "-Xmx${jvm_max_heap_size}", '-Dcom.redhat.fips=false', $jvm_extra_args]
+  } else {
+    $jvm_cmd_arr = ["-Xms${jvm_min_heap_size}", "-Xmx${jvm_max_heap_size}", $jvm_extra_args]
+  }
```
evgeni, line 160: I would've probably not duplicate the heap size stuff and do something like
```suggestion
  $jvm_heap_arr = ["-Xms${jvm_min_heap_size}", "-Xmx${jvm_max_heap_size}"]
  if $disable_fips {
    $jvm_cmd_arr = $jvm_heap_arr + ['-Dcom.redhat.fips=false', $jvm_extra_args]
  } else {
    $jvm_cmd_arr = $jvm_heap_arr + [$jvm_extra_args]
  }
```

if that's even valid syntax…
but in the end I don't care too much as long this chapter gets closed

evgeni (approved):

ehelms:
@ekohl could you have another look?

evgeni:
Why didn't CI run on this (anymore)?

evgeni:
Ubuntu failure looks unrelated.

ekohl (approved):
> Ubuntu failure looks unrelated.

Yes, I haven't looked too deeply into why it doesn't work. I think that particular one has never worked.

ekohl:
> Yes, I haven't looked too deeply into why it doesn't work. I think that particular one has never worked.

Oh, and to add to that: in a VM it does work. So it's somehow the interaction between docker and systemd. Note that on Focal we only test puppetserver on 7 so that it passes on 6 is just because the agent works on 6. Perhaps it's the same thing that we see PostgreSQL fail on puppet-foreman with Focal.

PR #829, opened 2022-03-16
puppet agent timer: depend on network/start it if offline

bastelfreak:
Two changes: With adding network-online.target as dependency we ensure
that the network is really available. That might not be the case for the
network.target. In case the system has no proper network manager,
network-online.target isn't available. For that situation we still
depend on network.target. Also we now have Wants= defined. That will
start the targets if the agent is started.

We had a similar discussion upstream:
https://github.com/puppetlabs/puppet/pull/8893

### DIFF
--- a/templates/agent/systemd.puppet-run.service.erb
+++ b/templates/agent/systemd.puppet-run.service.erb
@@ -3,7 +3,8 @@
 #
 [Unit]
 Description=Systemd Timer Service for Puppet Agent
-After=network.target
+Wants=network.target network-online.target
+After=network.target network-online.target
 
 [Service]
 Type=oneshot


ekohl:
I wonder if we should push the timer upstream.

evgeni:
I'm wondering why the `Wants` change -- you don't have the same in the linked `puppet.service` PR.

My biggest "fear" (hard word) here is: a user has intentionally dropped out of `network` target, but because the timer is active, it will turn the network back on again? Feels like this is something we should have no business in. Or am I misunderstanding timers (possible).

evgeni:
(depending on the discussion, we could squeeze that in into #834 ;))

ekohl:
@bastelfreak since you linked the PR to me, any idea about what @evgeni said?

PR #830, opened 2022-03-16
metadata.json: Use https URL to git repo

bastelfreak:
using plaintext git URLs isn't supported by GitHub anymore. People might
parse the metadata.json to get the git URL which will then fail.


ekohl (approved):
Thanks!

PR #835, opened 2022-04-22
Add Array of Integers as valid type for run_minute

danifr:
If no value for `run_minute` is specified the module internally uses https://forge.puppet.com/modules/puppet/extlib/reference#extlibip_to_cron
This function returns an array of integers as values that then are passed to the cron/systemd timer resources.
Having a list of values is not currently possible if you set the value of `run_minute` via the class.

This change allows specifying the minute you want as an Array of integers.

### DIFF
--- a/manifests/agent/service/cron.pp
+++ b/manifests/agent/service/cron.pp
@@ -1,9 +1,9 @@
 # Set up running the agent via cron
 # @api private
 class puppet::agent::service::cron (
-  Boolean                 $enabled = false,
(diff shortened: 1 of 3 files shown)


danifr:
I don't think that one failed test has something to do with my change. Can you please check @ekohl ? Thank you very much

DriesVerachtert:
It would be great to have this merged: at the moment this bug causes puppet to change the crontab entry at each run.

ehelms (approved):

PR #841, opened 2022-06-22
Update to voxpupuli-test 5

ekohl:
I'm still a bit unsure about the fact handling. https://puppet.com/docs/puppet/7/lang_facts_builtin_variables.html is a very useful resource, but I'm not entirely sure what the intended values should be and how to deal with it in various environments. See the TODOs inline.


manifests/params.pp
```diff
@@ -33,19 +33,15 @@
...
+  $srv_domain = fact('networking.domain')
 
   # lint:ignore:puppet_url_without_modules
   $pluginsource        = 'puppet:///plugins'
   $pluginfactsource    = 'puppet:///pluginfacts'
   # lint:endignore
   $classfile           = '$statedir/classes.txt'
   $syslogfacility      = undef
-  $environment         = $::environment
+  $environment         = $server_facts['environment']
```
ekohl, line 44: While https://www.puppet.com/docs/puppet/7/lang_facts_builtin_variables.html#lang_facts_builtin_variables-server-facts doesn't mention it, `$server_facts['environment']` is present when using `puppet apply`.

manifests/params.pp
```diff
@@ -198,13 +194,8 @@
...
   # Will this host be a puppet agent ?
   $agent                      = true
-  $client_certname            = $::clientcert
-
-  if defined('$::puppetmaster') {
-    $puppetmaster             = $::puppetmaster
-  } else {
-    $puppetmaster             = undef
-  }
+  $client_certname            = $trusted['certname']
```
ekohl, line 197: This is also present when using `puppet apply`

manifests/params.pp
```diff
@@ -198,13 +194,8 @@
...
+  $puppetmaster               = $server_facts['servername']
```
ekohl, line 198: This is one I wonder about. `puppetmaster` is a global variable we set in Foreman's ENC, which allows you to change the puppetserver easily. Also, setting this outside of Foreman's ENC. For example, when using DNS lookups. Should we keep the old `defined` code/update it to `getvar()`?

ekohl (commented):
@alexjfisher @bastelfreak would you mind taking a look at these comments?

manifests/params.pp
ekohl: I ended up changing it to `getvar('puppetmaster')`, to remain compatible.

spec/support/trusted.rb
```diff
@@ -0,0 +1,4 @@
+RSpec.configure do |c|
+  c.trusted_node_data = true
+  c.trusted_server_facts = true
```
ekohl, line 3: Surprisingly https://github.com/puppetlabs/rspec-puppet#trusted_node_data and https://github.com/puppetlabs/rspec-puppet#trusted_server_facts are off by default.

spec/classes/puppet_server_spec.rb
```diff
@@ -503,6 +496,7 @@
           end
 
           it 'should not sync the crl' do
+            pending("rspec-puppet always sets $server_facts['servername']")
```
ekohl, line 499: https://github.com/puppetlabs/rspec-puppet/issues/37

tuxmea (approved):

ekohl:
No content change compared to the previous version, just extracted part of it to https://github.com/theforeman/puppet-puppet/pull/841 and rebased it on top. Also added a link to the rspec-puppet GH issue in the code itself.

Issue #842, opened 2022-07-20
FreeBSD puppetserver gem-home is missing the system Gem.dir


llua:
Which is present in the configuration file provided by the package, removing it seems to cause the puppetserver to die shortly after starting.

ekohl:
Can you provide a bit more information, like how to reproduce this? Which versions are you using?

llua:
FreeBSD 13.0-RELEASE, puppetserver7 7.8.0
```
pkg install -y puppetserver7 rubygem-ruby-augeas
if ! getent hosts "${host:=$(hostname)}"; then
  printf '%s\n' "127.0.0.1 $host" >> /etc/hosts
fi >/dev/null 2>&1
puppet module install theforeman/puppet
puppet module install puppetlabs/cron_core
puppet module install puppetlabs/augeas_core
puppet apply -e '
  class { "puppet":
    server => true,
    server_foreman => false,
  }
'
```
Should reproduce it from a clean install
puppetserver.log: http://ix.io/44Ym
```
# pkg info --dependencies puppetserver7-7.8.0
puppetserver7-7.8.0:
        rubygem-puppetserver-ca-2.3.6
        puppet7-7.17.0
        bash-5.1.16
        sudo-1.9.11p3
        ruby-3.0.4_2,1
        openjdk8-8.332.09.1_1
        rubygem-semantic_puppet-1.0.4

```

llua:
though i am unsure how to fix the testing in `spec/classes/puppet_server_puppetserver_spec.rb`, this patch fixes the problem for me.
```
diff --git a/manifests/server/puppetserver.pp b/manifests/server/puppetserver.pp
index fcf98ee..b51083e 100644
--- a/manifests/server/puppetserver.pp
+++ b/manifests/server/puppetserver.pp
@@ -164,7 +164,7 @@ class puppet::server::puppetserver (
   $jvm_cmd = strip(join(flatten($jvm_cmd_arr), ' '))

   if $facts['os']['family'] == 'FreeBSD' {
-    $server_gem_paths = [ '${jruby-puppet.gem-home}', "\"${server_puppetserver_vardir}/vendored-jruby-gems\"", ] # lint:ignore:single_quote_string_with_variables
[...]

PR #843, opened 2022-08-03
Keep ruby's system Gem.dir in gem-path on FreeBSD

llua:
without it puppetserver will not successfully start.

Fixes: https://github.com/theforeman/puppet-puppet/issues/842


ekohl (commented):
Bleh, looks like lint fails. I addressed that in https://github.com/theforeman/puppet-puppet/pull/841 but that needs a bit of further work.

Right now I'm very busy with Foreman 3.4's branching so I may need a few days to get back to it.

ekohl (approved):
I took the liberty of rebasing this.

PR #844, opened 2022-08-17
Puppet 7 no longer ships with a cron built-in

yakatz:
It now requires `cron_core` and seems like it won't work with any other module that provides the `cron` type (like `puppet-cron` which defines a different type with the same name)

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -24,6 +24,10 @@
       "version_requirement": ">= 4.18.0 < 9.0.0"
     },
     {
+      "name": "puppetlabs/cron_core",
+      "version_requirement": ">= 1.0.0 < 2.0.0"
+    },
+    {
       "name": "puppet/extlib",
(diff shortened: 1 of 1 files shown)


ekohl (commented):
I'm not sure this is correct. In https://github.com/puppetlabs/puppet-agent/blob/main/configs/components/module-puppetlabs-cron_core.json it looks like it should still be part of the AIO install. I do think that FreeBSD doesn't include the bundled modules.

Can you share more about your environment?

yakatz:
Our environment is mostly Redhat 7, 8, and 9, using the Puppet-supplied yum repos. The cron_core module is included in vendor_modules (actually as of Puppet 6), but it is an older version and there are other modules that have a `cron` type that clobber it (for example, voxpupuli's cron module), so I thought it made sense to list it.

PR #845, opened 2022-08-17
Allow setting cron target

yakatz:
For example, this is useful if you have `/var/spool/cron` mounted over NFS and use the `-c` option which would mean Puppet cronjobs would not run on other machines.

### DIFF
--- a/spec/classes/puppet_agent_spec.rb
+++ b/spec/classes/puppet_agent_spec.rb
@@ -254,6 +254,54 @@ describe 'puppet' do
           end
         end
 
+        describe 'when runmode => cron with specified target' do
+          let :params do
+            super().merge(runmode: 'cron',
+                          cron_target: '/etc/cron.d/puppet'
+                         )
+          end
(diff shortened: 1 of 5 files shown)


spec/classes/puppet_agent_spec.rb
```diff
@@ -254,6 +254,54 @@
...
+              is_expected.to contain_service('puppet')
+                .with_ensure('stopped')
+                .with_name('puppet')
+                .with_hasstatus('true')
+                .with_enable('false')
+            end
+            it { is_expected.to contain_class('puppet::agent::service::systemd').with_enabled(false) }
+            it { is_expected.to contain_service('puppet-run.timer').with_ensure(false) }
+            it do
+              is_expected.to contain_cron('puppet')
```
ekohl, line 282: Should this also have the `with_target()`?

ekohl (commented):
I'll have a look at the failing tests first.

PR #847, opened 2022-09-26
Allow permissions change of puppet.conf

teluq-pbrideau:
There is confidential info in my `puppet.conf` (`http_proxy_password` to be precise) that I would like to secure from unprivileged users.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -162,6 +162,9 @@
 #                                           of the classes associated with the retrieved
 #                                           configuration.
 #
+# $puppetconf_mode::                        The permissions for /etc/puppetlabs/puppet/puppet.conf
+#                                           default to '0644' and '0674' on windows
+#
 # == puppet::agent parameters
 #
 # $agent::                                  Should a puppet agent be installed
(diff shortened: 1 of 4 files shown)


teluq-pbrideau:
I don’t understand why the check failed…  The two warnings in `manifest/server/config.pp` are not changed I made.  And if I run `bundle exec rake validate lint check` on my laptop, there is about 2000 lines of errors, and none seems to be about my changes.  What do I miss?

ekohl:
I think I need to finish https://github.com/theforeman/puppet-puppet/pull/841 first.

manifests/init.pp
```diff
@@ -752,6 +755,7 @@
   Optional[Stdlib::Absolutepath] $server_versioned_code_id = undef,
   Optional[Stdlib::Absolutepath] $server_versioned_code_content = undef,
   Array[String[1]] $server_jolokia_metrics_whitelist = [],
+  Pattern[/^[0-9]{3,4}$/] $puppetconf_mode = $puppet::params::puppetconf_mode,
```
ekohl, line 758: Would `Stdlib::Filemode` be better?
```suggestion
  Stdlib::Filemode $puppetconf_mode = $puppet::params::puppetconf_mode,
```

teluq-pbrideau: Thanks for pointing this out, I was not aware this existed!

PR #849, opened 2022-10-27
Add option to allow unauthenticated access to metrics, and for generic additions to auth

sshipway:
This minor change allows generic control of the auth.conf file, and permitting unauthenticated retrieval of metrics (as is required by our monitoring system)
* 2 new parameters to server class
* Usage comments added
* Spec tests added and run successfully


sshipway:
Closing because #904 replaces it

PR #851, opened 2022-11-07
Fixes #35728 - Disable telemetry by default

ekohl:
Puppetserver 7.0 adopted [opt-out telemetry](https://puppet.com/docs/puppet/7/server/puppet_server_usage_telemetry.html), meaning telemetry is sent by default. Foreman traditionally has had a policy where no telemetry was sent by default. Disabling the telemetry would bring align Puppetserver to this policy.

Still a draft since I haven't checked all the implications. Users of the installer will also need to change their answer. Possibly this could be a migration, but then we'd disrespect users who wanted to opt-in. Possibly an upgrade note.


ekohl:
That may actually be better yes. At least for the future to explicitly record the opt-in/opt-out.

evgeni (approved):

PR #852, opened 2023-01-10
Permit setting `default_schedules` to `true`

jcpunk:
Also allow override for `localconfig`, but I don't know why someone would change it....

I've been using the default schedules for a few things and, while I get not everyone likes them, I'd like to be able to keep using them.

Without this flag, this module is mostly useless for my environment.

### DIFF
--- a/manifests/agent/config.pp
+++ b/manifests/agent/config.pp
@@ -3,8 +3,8 @@
 class puppet::agent::config inherits puppet::config {
   puppet::config::agent{
     'classfile':         value => $puppet::classfile;
-    'localconfig':       value => '$vardir/localconfig';
-    'default_schedules': value => false;
+    'localconfig':       value => $puppet::localconfig;
+    'default_schedules': value => $puppet::agent_default_schedules;
     'report':            value => $puppet::report;
     'masterport':        value => $puppet::port;
     'environment':       value => $puppet::environment;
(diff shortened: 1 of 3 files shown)


jcpunk:
The CI failures don't appear to be related to my changes....

I'm not sure which redmine project to file this under....

ekohl:
> I'm not sure which redmine project to file this under....

Technically not required, but https://projects.theforeman.org/projects/puppet-foreman if you want to.

ekohl (approved):
I'm happy to not hardcode these. Let me know if you want to create a Redmine issue, but I'd be fine to merge it without.

jcpunk:
I've tried to link up to https://projects.theforeman.org/issues/35932

ekohl (changes requested):
You should use `Fixes #... - ` as a prefix. The dash is currently missing, so not recognized. The Redmine issues status should link to it if it is recognized.

jcpunk:
I think I've got the commit line sorted now.

ekohl (approved):
Yes indeed. You can see https://github.com/theforeman/puppet-puppet/pull/852/checks?check_run_id=10608471594 now shows a link to the right issue and the Redmine issue was flipped to Ready For Testing with a link to this PR. Thanks!

PR #853, opened 2023-01-10
Add trivial `r10k` integration point.

jcpunk:
Installation and configuration of `r10k` is out of scope. This merely provides a simple way to use it with the git repo workflow.

In my experience `r10k` also tends to use the isolated environment types from https://www.puppet.com/docs/puppet/7/environment_isolation.html


templates/server/post-receive.erb
```diff
@@ -111,12 +111,33 @@ $stdin.each_line do |line|
         end
         do_cmd("git submodule sync")
         do_cmd("git submodule update --init --recursive")
+        <% if @git_repo_r10k -%>
+        if File.exists? 'Puppetfile'
```
ekohl, line 115: `File.exists?` is deprecated in favor of `File.exist?`. Ruby 3.2 removed the deprecated variant.
```suggestion
        if File.exist? 'Puppetfile'
```

And yes, I now see we do the same thing earlier in the file.

templates/server/post-receive.erb
```diff
@@ -111,12 +111,33 @@ $stdin.each_line do |line|
...
+      <% if @git_repo_r10k -%>
+      Dir.chdir environment_path
+      if File.exists? 'Puppetfile'
+        puts("Installing modules from r10k")
+        do_cmd("r10k puppetfile install")
+      end
+      <% end -%>
+      <% if @git_repo_gen_types -%>
+      puts("Generating types for #{environment_name})"
+      do_cmd("puppet generate types --environment #{environment_name}"
```
ekohl, line 139: I recall in the past that I also used an explicit directory path. In theory it shouldn't make a difference, but it doesn't hurt. We already have that path anyway so I wonder if it would be the right thing to do.

templates/server/post-receive.erb
```diff
@@ -111,12 +111,33 @@ $stdin.each_line do |line|
         end
         do_cmd("git submodule sync")
         do_cmd("git submodule update --init --recursive")
+        <% if @git_repo_r10k -%>
+        if File.exists? 'Puppetfile'
+          puts("Installing modules from r10k")
```
ekohl, line 116:
```suggestion
          puts("Installing modules using r10k")
```

templates/server/post-receive.erb
jcpunk: It seems that `puppet` only excepts environment names (and not full paths).

```
Could not find a directory environment named '/etc/puppetlabs/code/environments/test'
```

templates/server/post-receive.erb
jcpunk: Fixed for both.

templates/server/post-receive.erb
jcpunk: Done

templates/server/post-receive.erb
jcpunk: Filed ticket upstream as this behavior seems odd to me : https://tickets.puppetlabs.com/browse/PUP-11700

jcpunk:
Sounds good, I'll try to keep an eye out, but if I miss it let me know and I'll get rebasing :)

ekohl:
In case you missed it, it was merged.

jcpunk:
:) Rebased my various PRs

ekohl:
Something I've been thinking about: r10k (and g10k for that matter) can both update environments based on a git repo. Instead of running r10k for every environment that was updated, would it make sense to instead look at an entirely different hook script that just runs r10k as post-receive?

The one consideration may be that r10k may not be able to do git submodules, though that may be acceptable for many users.

jcpunk:
My understanding is the `r10k` native git integration requires a bunch more config stuff.  That could be an interesting point of integration, but is probably a bit of work...

PR #854, opened 2023-01-10
Default to TLS1.3 or TLS1.2 by default since puppet supports both

jcpunk:
TLSv 1.3 support has been in puppetserver since 2019, should be safe to start defaulting to that on.


ekohl (approved):

alexjfisher:
By default TLS 1.3 is _still_ disabled because there are no TLS 1.3 ciphers in the default list...  I'll open a PR to correct this.

alexjfisher:
ciphers added in https://github.com/theforeman/puppet-puppet/pull/878

Issue #855, opened 2023-01-10
Sync up with upstream naming (`master`->`server`)


jcpunk:
Upstream has switched a bunch of the settings from `master` to `server` for various parts of the puppet server.  It would be nice if this module followed as I've no idea how long the old names will work.

ekohl:
https://github.com/theforeman/puppet-puppet/pull/770 was opened but never finished. Needs a non-trivial rebase now and I'll admit that I lack the time to work on it, but PRs are welcome.

PR #859, opened 2023-01-12
Fixes #35933 Update for change from puppet master to server

jcpunk:
This is a somewhat naive stab at #855

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -24,7 +24,7 @@
 #                                           Can also install only server packages with value
 #                                           of 'server' or only agent packages with 'agent'.
 #
-# $port::                                   Override the port of the master we connect to.
+# $agent_server_port::                      Override the port of the server we connect to.
 #
 # $splay::                                  Switch to enable a random amount of time
 #                                           to sleep before each run.
@@ -99,7 +99,7 @@
 # $syslogfacility::                         Facility name to use when logging to syslog
 #
 # $use_srv_records::                        Whether DNS SRV records will be used to resolve
-#                                           the Puppet master
+#                                           the Puppet server
 #
 # $srv_domain::                             Search domain for SRV records
 #
@@ -171,7 +171,7 @@
 #
 # $agent_noop::                             Run the agent in noop mode.
 #
-# $puppetmaster::                           Hostname of your puppetmaster (server
+# $agent_server_hostname::                  Hostname of your puppetserver (server
 #                                           directive in puppet.conf)
 #
 # $prerun_command::                         A command which gets excuted before each Puppet run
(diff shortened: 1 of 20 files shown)


jcpunk:
With this change, I think the minimum version of the puppet server supported is probably in the 6.x or 7.x range (README.md has a min listed version).

I can't seem to find the version where they started supporting both names.

manifests/init.pp
```diff
@@ -24,7 +24,7 @@
 #                                           Can also install only server packages with value
 #                                           of 'server' or only agent packages with 'agent'.
 #
-# $port::                                   Override the port of the master we connect to.
+# $port::                                   Override the port of the server we connect to.
```
ekohl, line 27: I've thought about this and wondered if we need to rename it to `agent_server_port`, similar to how you changed `puppetmaster` to the more descriptive `agent_server_hostname`.

manifests/init.pp
```diff
@@ -258,16 +258,16 @@
 #
 # $server_manage_user::                     Whether to manage the server user resource
 #
-# $server_user::                            Name of the puppetmaster user.
+# $server_user::                            Name of the puppetserver user.
```
ekohl, line 261: Not sure I like the description. It never really was that good to begin with. Perhaps something like:
```suggestion
# $server_user::                            Username used for the puppetserver  process
```

Then similar for the group.

manifests/params.pp
```diff
@@ -197,14 +197,17 @@
   $client_certname            = $trusted['certname']
 
   # Set by the Foreman ENC
-  $puppetmaster               = getvar('puppetmaster')
+  $agent_server_hostname      = getvar('agent_server_hostname')
+  if $agent_server_hostname == '' {
+    $agent_server_hostname = getvar('puppetmaster')
+  }
```
ekohl, line 203: You can't reassign variables in Puppet. I also wouldn't expect it to return an empty string. This probably works:
```suggestion
  $agent_server_hostname = pick(getvar('agent_server_hostname'), getvar('puppetmaster'))
```

Then you also need something to modify the ENC output, which is set here:
https://github.com/theforeman/foreman_puppet/blob/5500ecd3d772b36fb80f71b2aff05822471167df/app/services/foreman_puppet/host_info_providers/puppet_info.rb#L38

For now I'd avoid this whole matter and keep it separate. So
```suggestion
  $agent_server_hostname = getvar('puppetmaster')
```

Then make an issue to tackle on its own. For example, we've talked about moving all Foreman details into its own hash. So you have a `foreman_enc` hash with all Foreman details, rather than a bunch at the top scope. That way you'll never introduce `agent_server_hostname`, but rather some other variable.

jcpunk: That looks beyond my current limits.... yeah I'll revert these lines.

jcpunk:
I've tried to address your feedback, let me know if I missed something or you find new things that could be improved.

jcpunk:
I'm confident the remaining test errors are obvious, but I'm not seeing what is wrong there....

PR #860, opened 2023-01-13
Fixes #35943 - Switch to puppetlabs vcsrepo for gitrepo tracking

jcpunk:
This PR switches to using the `puppetlabs/vcsrepo` module.  In theory this should reduce the maintenance overhead of this module for `git` related issues.

### DIFF
--- a/manifests/server.pp
+++ b/manifests/server.pp
@@ -80,9 +80,11 @@
 #
 # $common_modules_path::               Common modules paths
 #
-# $git_repo_path::                     Git repository path
+# $git_repo_path::                     Git repository path on disk
 #
-# $git_repo_mode::                     Git repository mode
+# $git_repo_umask::                    Umask used during git operations
+#
+# $git_repo_hook_mode::                Git repository hook mode
 #
 # $git_repo_group::                    Git repository group
 #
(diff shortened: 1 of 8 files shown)


manifests/server.pp
```diff
@@ -384,8 +388,10 @@
   Array[Stdlib::Absolutepath, 1] $envs_dir = $puppet::server_envs_dir,
   Optional[Stdlib::Absolutepath] $envs_target = $puppet::server_envs_target,
   Variant[Undef, String[0], Array[Stdlib::Absolutepath]] $common_modules_path = $puppet::server_common_modules_path,
-  Pattern[/^[0-9]{3,4}$/] $git_repo_mode = $puppet::server_git_repo_mode,
+  Pattern[/^[0-9]{3,4}$/] $git_repo_hook_mode = $puppet::server_git_repo_hook_mode,
   Stdlib::Absolutepath $git_repo_path = $puppet::server_git_repo_path,
+  Optional[String] $git_repo_source = $puppet::server_git_repo_source,
```
ekohl, line 393: This is actually a pretty massive difference in functionality. The old mode created a bare repository which you were intended to push to. Do you have have any idea how the hook interacts with that?

Before this you had:
* Empty repo is initialized
* You push all data
* Hook creates all environments

If you now provide something to clone from, do you end up with all environments? If not, you can't trigger it unless you create some dummy commit, right?

ekohl: As you say: without the parameter the behavior is unchanged. Often I end up then writing one patch to move over and then another to change the behavior. So we could leave out the source parameter at first and then introduce that as a new parameter in a separate feature.

jcpunk: I'll split this into two as my test VMs seem to be sad today.

ekohl (commented):
We'll also want to update our installer to pull in `puppetlabs/vcsrepo` instead of `theforeman/git`. Today we have this line:

https://github.com/theforeman/foreman-installer/blob/e3212a7fbe79619a80b05ad3e8106539f5b5dbbe/Puppetfile#L15

Since it's an external module there is no need to pull it from git. I'd prefer to use a forge release instead.

You can use `Refs #35943` in that PR to relate the two.

manifests/server.pp
jcpunk: Opened https://projects.theforeman.org/issues/35945 to track the `git clone` bits.

jcpunk:
I do not understand those test failures...

jcpunk:
Ok I think it is sorted again... that ubuntu failure looks like a docker issue...

ekohl (approved):
Thanks! It may still be good to do a follow up to utilize `vcsrepo::manage::git` but for now I'm merging this.

PR #861, opened 2023-01-23
Fixes #35983 - Permit recursive ownership/permissions for environments

jcpunk:
This lets folks easily set owner/group/mode for the content of /etc/puppetlabs/code/environments.

There is a performance cost, but folks opting into this should know that already.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -235,6 +235,8 @@
 #
 # $server_environments_mode::               Environments directory mode.
 #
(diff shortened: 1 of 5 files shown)


jcpunk:
I don't think the CI failures on this are me....

ekohl:
Nope, that's https://tickets.puppetlabs.com/browse/PUP-11722 / https://github.com/ruby-concurrency/concurrent-ruby/issues/986.

ekohl (approved):

PR #862, opened 2023-01-24
Refs #35985 - Implement optional Katello integration

ekohl:
This is https://github.com/theforeman/puppet-puppet/pull/801 reopened. The whole effort to build PSF ended up going nowhere for various reasons. This revives the previous effort to it it more user friendly. We can remove the need to pass various parameters for end users, which makes life easier.

Additional PRs will be needed to make use of this.

### DIFF
--- a/manifests/server/foreman.pp
+++ b/manifests/server/foreman.pp
@@ -1,6 +1,21 @@
 # @summary Set up Foreman integration
 # @api private
-class puppet::server::foreman {
+class puppet::server::foreman (
+  Boolean $katello = false,
+) {
+  if $katello {
+    include certs::puppet
+    Class['certs::puppet'] -> Class['puppetserver_foreman']
+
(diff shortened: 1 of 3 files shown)


manifests/server/foreman.pp
```diff
@@ -0,0 +1,33 @@
+# @summary Set up Foreman integration
+# @api private
+class puppet::server::foreman (
+  Boolean $katello = false,
+) {
+  if $katello {
+    include certs::puppet
```
ehelms, line 7: I assume we'll have this as a soft dependency where we assume it's presence if someone is setting katello true that they will have the requisite dependencies present? I am fine with that, and it helps make things easier, I am wanting to ensure we call that out (maybe in the Readme?).

ekohl: How about this?

ehelms (approved):

PR #866, opened 2023-02-21
Release 16.4.0

ehelms:


evgeni:
Isn't the telemetry change technically a breaking one, as it changes a default?

ekohl (approved):

ekohl:
I also considered that, but I didn't want to do too many major releases. If we would do a major release, I'd probably also include https://github.com/theforeman/puppet-puppet/pull/859 but that's certainly breaking.

PR #868, opened 2023-03-11
Support new non-AIO puppet Debian packages

jcharaoui:
The upcoming release of Debian will ship with a new non-AIO puppetserver package, and these configuration tweaks are needed to make the module compatible with it.

https://packages.debian.org/search?keywords=puppetserver

### DIFF
--- a/manifests/server/puppetserver.pp
+++ b/manifests/server/puppetserver.pp
@@ -169,6 +169,8 @@ class puppet::server::puppetserver (
       context => '/files/etc/rc.conf',
       changes => ["set puppetserver_java_opts '\"${jvm_cmd}\"'"],
     }
+  } elsif $facts['os']['family'] == 'Debian' and !$puppet::params::aio_package {
+    $server_gem_paths = ['${jruby-puppet.gem-home}', '/usr/lib/puppetserver/vendored-jruby-gems'] # lint:ignore:single_quote_string_with_variables
   } else {
     if $jvm_cli_args {
       $changes = [
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -282,7 +282,7 @@ class puppet::params {
   $server_version = undef
 
   if $aio_package or
-    ($facts['os']['name'] == 'Debian' and Integer($facts['os']['release']['major']) >= 12) {
+    ($facts['os']['name'] == 'Debian' and versioncmp($facts['os']['release']['major'], '12') >= 0) {
     $client_package = ['puppet-agent']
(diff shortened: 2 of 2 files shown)


manifests/server/puppetserver.pp
```diff
@@ -240,7 +240,7 @@
   }
 
   unless $facts['os']['family'] == 'FreeBSD' or
-    ($facts['os']['family'] == 'Debian' and $puppet::params::aio_package) {
+  ($facts['os']['family'] == 'Debian' and $puppet::params::aio_package) {
```
ekohl, line 243: I consider this a bug in the indent rule. Not going to dive into it in the weekend, just want to let you know.

manifests/server/puppetserver.pp
```diff
@@ -169,6 +169,8 @@
       context => '/files/etc/rc.conf',
       changes => ["set puppetserver_java_opts '\"${jvm_cmd}\"'"],
     }
+  } elsif $facts['os']['family'] == 'Debian' and !$puppet::params::aio_package {
+    $server_gem_paths = ['${jruby-puppet.gem-home}', '/usr/lib/puppetserver/vendored-jruby-gems'] # lint:ignore:single_quote_string_with_variables
```
ekohl, line 173: Given the repetition in declaring a vendored-jruby-gems dir, is this something that belongs in `params.pp`?

ekohl (commented):
Thanks for this! I appreciate how small it is.

manifests/server/puppetserver.pp
jcharaoui: Declaring `$server_gem_paths` in `puppet::params` was my first idea too, however it doesn't work because the variable isn't a class parameter of `puppet::server::puppetserver`.

manifests/server/puppetserver.pp
jcharaoui: OK, just let me know if you want me to remove that commit from the branch.

ekohl: In https://github.com/voxpupuli/puppet-lint-strict_indent-check/pull/28 I already added examples for this, though I suspect that in this case you should use braces around the statements.

anarcat (approved):
there's a bunch of CI warnings, but once those are fixed I fully support this proposal. )

manifests/server/puppetserver.pp
jcharaoui: I'm not sure where you're suggesting I use braces (`{}`) in there?

PR #874, opened 2023-05-16
Respect puppet::server::additional_settings parameter

### LINKED ISSUE #810: server::config class references wrong variable for additional_settings
I think [server::config](https://github.com/theforeman/puppet-puppet/blob/master/manifests/server/config.pp#L3) class uses the wrong variable for referencing to `additional_settings`.

In [this code](https://github.com/theforeman/puppet-puppet/blob/7534cbe648d6576509303da7be4f7724ee2d1fa4/manifests/server/config.pp#L100)  we configure additional settings for the `master` section and we use `$puppet::server_additional_settings` variable instead of `$puppet::server::additional_settings` variable:
```puppet
  $puppet::server_additional_settings.each |$key,$value| {
    puppet::config::master { $key: value => $value }
  }
```

Due to this the [additional_settings](https://github.com/theforeman/puppet-puppet/blob/master/manifests/server.pp#L401) parameter of `server` class is ignored.

ekohl:
Previously it was read from the main puppet class, instead of puppet::server (which in turn defaulted to the parameter from the main class).

Fixes #810

### DIFF
--- a/manifests/server/config.pp
+++ b/manifests/server/config.pp
@@ -95,7 +95,7 @@ class puppet::server::config inherits puppet::config {
     }
   }
 
-  $puppet::server_additional_settings.each |$key,$value| {
+  $puppet::server::additional_settings.each |$key,$value| {
     puppet::config::server { $key: value => $value }
   }
 


ekohl:
I think the idempotency issue is just a fluke and https://github.com/theforeman/puppet-puppet/pull/874 drops Puppet 6 support anyway.

PR #880, opened 2023-07-07
Refs #36573 - Drop default value for foreman_url

ekohl:
This default matches what's in theforeman/puppetserver_foreman and is not needed. Users can still override it, but the default just duplicates things.


evgeni (changes requested):
needs tests changes for the default case, otherwise LGTM

evgeni (approved):

PR #882, opened 2023-07-28
README.md: Add more PuppetDB examples & simplify `puppet apply` example

bastelfreak:


README.md
```diff
@@ -136,10 +180,7 @@ To use this in standalone mode, edit a file (e.g. install.pp), put in a class re
 as per the examples above, and the execute _puppet apply_ e.g:
 
 ```sh
-cat > install.pp <<EOF
-class { '::puppet': server => true }
-EOF
-puppet apply install.pp --modulepath /path_to/extracted_tarball
+puppet apply -e 'class { 'puppet': server => true }' --modulepath /path_to/extracted_tarball
```
ekohl, line 183:
```suggestion
puppet apply -e "class { 'puppet': server => true }" --modulepath /path_to/extracted_tarball
```

But I think the install.pp way is also useful because there are a lot of non-trivial examples above and the file approach may be easier. Or use stdin:
```suggestion
puppet apply --modulepath /path_to/extracted_tarball <<EOF
class { 'puppet': server => true }
EOF
```

ekohl (approved):

PR #887, opened 2023-09-13
Allow disabling CRL checking on agent

cm-jc:
In extrernal CA configurations, it is possible that a CRL is not present on the server. This allows to tell the agent to skip downloading it and checking it.


cm-jc:
@ekohl , Can I get a review?

ekohl (approved):
Sorry for missing this. You can also call `puppet::config::main` in your own module or use `$additional_settings` (via Hiera), but I like data types for proper validation so :+1:.

PR #888, opened 2023-09-29
Update gems to support testing Puppet 8

ekohl:


ekohl:
It still evaluates the class, so it hits the raise. Even though the real class wouldn't be executed on Archlinux.

ehelms (approved):

ehelms (approved):

ekohl:
The unit test failure was already addressed in a gem update. Once I fix the Archlinux tests I'd expect that to pass.

ekohl:
Hmm, still not so I think I need some other update.

evgeni (approved):

PR #892, opened 2023-11-03
set PrivateTmp=true for puppetserver

evgeni:


manifests/server/puppetserver.pp
```diff
@@ -214,6 +214,13 @@
         unit     => 'puppetserver.service',
         content  => "[Service]\nLimitNOFILE=${max_open_files}\n",
       }
+
+      systemd::dropin_file { 'puppetserver.service-privatetmp.conf':
```
ekohl, line 218: Perhaps link to the PR that makes it permanent?
```suggestion
      # https://github.com/puppetlabs/ezbake/pull/623
      systemd::dropin_file { 'puppetserver.service-privatetmp.conf':
```

PR #893, opened 2023-11-03
Fix CentOS 9 support in metadata.json

evgeni:


ehelms (approved):

spec/acceptance/puppetserver_upgrade_spec.rb
```diff
@@ -16,7 +16,12 @@
 
     case ENV['BEAKER_PUPPET_COLLECTION']
     when 'puppet7'
-      from_version = '7.6.0'
+      from_version =
+        if fact('os.family') == 'RedHat' && fact('os.release.major') == '9'
+          '7.13.0'
```
ekohl, line 21: Doesn't this mean 7.13.0 to 7.13.0, making it a noop? I guess this is a bit of an exception because there is no older release, but should we then skip the whole test instead at line 3?

Issue #832, opened 2022-03-17
Ubuntu Focal fails to set max open file limit


ekohl:
When running the tests (and enabling debug output) the following can be seen:

```
  Notice: /Stage[main]/Puppet::Server::Puppetserver/Systemd::Dropin_file[puppetserver.service-limits.conf]/File[/etc/systemd/system/puppetserver.service.d]/ensure: created
  Debug: /Stage[main]/Puppet::Server::Puppetserver/Systemd::Dropin_file[puppetserver.service-limits.conf]/File[/etc/systemd/system/puppetserver.service.d]: The container Systemd::Dropin_file[puppetserver.service-limits.conf] will propagate my refresh event
  Notice: /Stage[main]/Puppet::Server::Puppetserver/Systemd::Dropin_file[puppetserver.service-limits.conf]/File[/etc/systemd/system/puppetserver.service.d/limits.conf]/ensure: defined content as '{sha256}af70de30e188923816fe1eb77708cdc603f53918edd1f7815bea5bc48690f062'
  Debug: /Stage[main]/Puppet::Server::Puppetserver/Systemd::Dropin_file[puppetserver.service-limits.conf]/File[/etc/systemd/system/puppetserver.service.d/limits.conf]: The container Systemd::Dropin_file[puppetserver.service-limits.conf] will propagate my refresh event
  Debug: Systemd::Dropin_file[puppetserver.service-limits.conf]: The container Class[Puppet::Server::Puppetserver] will propagate my refresh event
  Debug: Class[Puppet::Server::Puppetserver]: The container Stage[main] will propagate my refresh event
  Debug: Class[Puppet::Server::Puppetserver]: The container Class[Puppet::Server::Config] will propagate my refresh event
  Debug: Class[Puppet::Server::Config]: The container Stage[main] will propagate my refresh event
  Debug: Class[Puppet::Server::Config]: The container Class[Puppet::Server] will propagate my refresh event
  Info: Class[Puppet::Server::Config]: Scheduling refresh of Class[Puppet::Server::Service]
  Info: Class[Puppet::Server::Service]: Scheduling refresh of Service[puppetserver]
  Debug: Executing: '/usr/bin/systemctl is-active -- puppetserver'
  Debug: Executing: '/usr/bin/systemctl is-enabled -- puppetserver'
  Debug: Executing: '/usr/bin/systemctl is-active -- puppetserver'
[...]

PR #894, opened 2023-11-03
Add support for OracleLinux/AlmaLinux/Rocky

evgeni:


ekohl (commented):
The number of open files issue might be the same as https://github.com/theforeman/puppet-puppet/issues/832

evgeni:
I've split out the EL9 support into https://github.com/theforeman/puppet-puppet/pull/900 as that's what works already.

evgeni:
and split out the other two into #901 and #902

PR #895, opened 2023-11-03
Add Ubuntu 22.04 support

evgeni:


spec/acceptance/puppetserver_upgrade_spec.rb
```diff
@@ -16,7 +16,12 @@
 
     case ENV['BEAKER_PUPPET_COLLECTION']
     when 'puppet7'
-      from_version = '7.6.0'
+      from_version =
+        if fact('os.name') == 'Ubuntu' && fact('os.release.major') == '22.04'
+          '7.13.0'
```
evgeni, line 21: this makes the upgrade test on Ubuntu 22 essentially a noop, but at least it passes now :D

spec/acceptance/puppetserver_upgrade_spec.rb
```diff
@@ -17,7 +17,8 @@
     case ENV['BEAKER_PUPPET_COLLECTION']
     when 'puppet7'
       from_version =
-        if fact('os.family') == 'RedHat' && fact('os.release.major') == '9'
+        if (fact('os.family') == 'RedHat' && fact('os.release.major') == '9') ||
+          (fact('os.name') == 'Ubuntu' && fact('os.release.major') == '22.04')
           '7.13.0'
```
ekohl, line 22: Like https://github.com/theforeman/puppet-puppet/pull/893#discussion_r1382209344. Perhaps we should skip the whole test if there is just a single version?

evgeni: I was hoping 7.14 just pops up soon and we can use 13->14 as the test.
But yeah, in the meantime we can also skip it totally for EL9 and U22.

spec/acceptance/puppetserver_upgrade_spec.rb
```diff
@@ -1,6 +1,8 @@
 require 'spec_helper_acceptance'
 
-unless unsupported_puppetserver
+unless unsupported_puppetserver ||
+    (fact('os.family') == 'RedHat' && fact('os.release.major') == '9') ||
+    (fact('os.name') == 'Ubuntu' && fact('os.release.major') == '22.04')
```
evgeni, line 5: or should I move this to a `unsupported_puppetserver_upgrade` helper?

ekohl: I was also thinking about that.

ekohl (approved):

PR #891, opened 2023-10-19
allow stdlib 9.x & puppetlabs/concat 9.x

jhoblitt:

PR #897, opened 2023-11-07
Fix Deprecation-Warning

cocker-cc:
Fix Deprecation-Warning "This function is deprecated, please use stdlib::to_json_pretty instead."

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -21,7 +21,7 @@
     },
     {
       "name": "puppetlabs/stdlib",
-      "version_requirement": ">= 4.18.0 < 9.0.0"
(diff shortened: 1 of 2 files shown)


ekohl:
I think this needs to e combined with https://github.com/theforeman/puppet-puppet/pull/891.

ekohl:
Done in https://github.com/theforeman/puppet-puppet/pull/898

cocker-cc:
Not quite, at least concearning `metadata.json`, because the Function `stdlib::to_json_pretty` is available from `stdlib` Version `9.0.0` only, and not earlier.

PR #900, opened 2023-11-15
Add support for OracleLinux/AlmaLinux/Rocky 9

evgeni:


ekohl (approved):
Any particular reason you're only adding version 9?

ekohl:
And now I see https://github.com/theforeman/puppet-puppet/pull/894#issuecomment-1812264519 ...

spec/classes/puppet_agent_spec.rb
```diff
@@ -94,7 +94,7 @@
 
         it { is_expected.to contain_class('puppet::agent::service::systemd').with_enabled(false) }
         case os
-        when /\A(debian|redhat|centos|scientific|fedora|ubuntu|sles|archlinux)-/
+        when /\A(debian|redhat|centos|scientific|fedora|ubuntu|sles|archlinux|oraclelinux|almalinux|rocky)-/
```
evgeni, line 97: you can argue this should rather match on family, not the os itself, but I am lazy

PR #903, opened 2023-11-21
Add the ability to modify metrics.conf with a jolokia metrics policy location file

nickgw:


manifests/server/puppetserver.pp
```diff
@@ -271,4 +272,11 @@
     ensure  => 'file',
     content => template('puppet/server/puppetserver/conf.d/metrics.conf.erb'),
   }
+
+  if $jolokia_metrics_policy != undef {
+    file { '/etc/puppetlabs/puppetserver/jolokia-access.xml' :
+      ensure => file,
```
ekohl, line 278: Does it make sense to ensure it's `absent` when `$jolokia_metrics_policy` is `undef`?

ekohl (commented):
Overall it looks like it makes sense.

manifests/server/puppetserver.pp
nickgw: Sorry for the super late reply, but I'd argue the absence of `$jolokia_metrics_policy` should not dictate enforcing the absence of the file. 

I could see ensuring `jolokia-access.xml` as absent if $jolokia_metrics_policy == `absent`, but not undef.

PR #905, opened 2023-12-28
Add puppet8 certificate renewal endpoint.

jcpunk:
From https://www.puppet.com/docs/puppet/8/server/http_certificate_renew

### DIFF
--- a/manifests/server/puppetserver.pp
+++ b/manifests/server/puppetserver.pp
@@ -73,6 +73,7 @@
 #   }
 #
 class puppet::server::puppetserver (
+  Optional[Pattern[/^[\d]\.[\d]+\.[\d]+$/]] $puppetserver_version = $puppet::server::puppetserver_version,
   String $config = $puppet::server::jvm_config,
   String $java_bin = $puppet::server::jvm_java_bin,
   Variant[String, Array[String]] $jvm_extra_args = $puppet::server::real_jvm_extra_args,
@@ -139,6 +140,8 @@ class puppet::server::puppetserver (
   Boolean $ca_allow_sans = $puppet::server::ca_allow_sans,
   Boolean $ca_allow_auth_extensions = $puppet::server::ca_allow_auth_extensions,
   Boolean $ca_enable_infra_crl = $puppet::server::ca_enable_infra_crl,
+  Boolean $server_ca_allow_auto_renewal = $puppet::server::server_ca_allow_auto_renewal,
+  String $server_ca_allow_auto_renewal_cert_ttl = $puppet::server::server_ca_allow_auto_renewal_cert_ttl,
   Optional[Integer[1]] $max_open_files = $puppet::server::max_open_files,
   Optional[Stdlib::Absolutepath] $versioned_code_id = $puppet::server::versioned_code_id,
   Optional[Stdlib::Absolutepath] $versioned_code_content = $puppet::server::versioned_code_content,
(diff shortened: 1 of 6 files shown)


templates/server/puppetserver/conf.d/ca.conf.erb
```diff
@@ -7,4 +7,11 @@ certificate-authority: {
 
     # enable the separate CRL for Puppet infrastructure nodes
     enable-infra-crl: <%= @ca_enable_infra_crl %>
+
+    # Setup auto renewal of certs
+    allow-auto-renewal: <%= @server_ca_allow_auto_renewal %>
+    # This value determines the lifetime of the cert if auto-renewal is enabled
+    auto-renewal-cert-ttl: <%= @server_ca_allow_auto_renewal_cert_ttl %>
```
ekohl, line 14: These values are not known in the scope, so they don't get replaced. That causes the build to fail:
CI failures are related:
```
  Dec 28 19:30:38 centos9-64-puppet8.example.com puppetserver[2140]: /etc/puppetlabs/puppetserver/conf.d/ca.conf: 14: Expecting close brace } or a comma, got ':' (if you intended ':' to be part of a key or string value, try enclosing the key or value in double quotes)
```

You need to add them to `puppet::server` (and further) as well.

ekohl (changes requested):
What about the compatibility for older Puppet versions? Is this option available in Puppet 7?

jcpunk:
Hmmmm, that endpoint doesn't appear to work with puppet7.  Is the `puppetversion` fact sufficient to confine this?  That seems to be the agent version, but I don't see a great way to find the server version...

ekohl:
We used to have `$puppet::server::real_puppetserver_version`, but dropped it when it was no longer needed. I'd suggest to take it from https://github.com/theforeman/puppet-puppet/commit/689d12150e12017f0675e46204c39aef06ea9b45#diff-7a36979cb6200e11647d047c711717f1662615c5e39375073ded103e83d90678L465-L473 and update it as needed.

manifests/server/puppetserver.pp
```diff
@@ -147,6 +150,17 @@
 ) {
   include puppet::server
 
+  # For Puppetserver, certain configuration parameters are version specific.
+  # We need a method to determine what version is installed.
+  if $puppetserver_version {
+    $real_puppetserver_version = $puppetserver_version
+  } elsif versioncmp($facts['puppetversion'], '7.0.0') >= 0 {
+    $real_puppetserver_version = $facts['puppetversion']
```
ekohl, line 158: I don't think this is correct. There's puppet-agent 7.27.0 but puppetserver 7.14.0 is the highest version.

I'd simplify it down to just 8.0.0, because that's a safe assumption
```suggestion
  } elsif versioncmp($facts['puppetversion'], '8.0.0') >= 0 {
    $real_puppetserver_version = '8.0.0'
```

templates/server/puppetserver/conf.d/auth.conf.erb
```diff
@@ -77,6 +77,20 @@ authorization: {
             name: "puppetlabs csr"
         },
 <%- if @server_ca -%>
+<%- if versioncmp(@real_puppetserver_version, '8.0.0') >= 0 -%>
```
ekohl, line 80: Looking at 689d12150e12017f0675e46204c39aef06ea9b45 it needs to be this to call the function in ERB
```suggestion
<%- if scope.function_versioncmp([@real_puppetserver_version, '8.0.0']) >= 0 -%>
```

ekohl (changes requested):

ekohl (approved):

PR #906, opened 2024-02-16
Replace File.exists? with File.exist? for Ruby 3.2 compatibility

tuxmea:


ekohl (approved):
I guess I only relied on CI tests for Ruby 3.2 compatibility.

PR #908, opened 2024-03-06
Disable FIPS on EL8+

ehelms:


manifests/server/puppetserver.pp
```diff
@@ -145,7 +145,7 @@
   Optional[Integer[1]] $max_open_files = $puppet::server::max_open_files,
   Optional[Stdlib::Absolutepath] $versioned_code_id = $puppet::server::versioned_code_id,
   Optional[Stdlib::Absolutepath] $versioned_code_content = $puppet::server::versioned_code_content,
-  Boolean $disable_fips = $facts['os']['family'] == 'RedHat' and $facts['os']['release']['major'] == '8',
+  Boolean $disable_fips = $facts['os']['family'] == 'RedHat' and $facts['os']['release']['major'] != '7',
```
evgeni, line 148: Given we don't support EL7 anymore, should this just be `true`?

ekohl: In this module we still support EL7 for puppetserver. Considering the options I think the current implementation is good enough. You could use `versioncmp` to test for `>=`, but I think this is close enough now.

ehelms:
Are the test failures related?

ekohl:
I feel like this is a regression in puppetserver.
```
  Mar 06 15:43:30 centos7-64-puppet7.example.com puppetserver[4297]: SyntaxError: /opt/puppetlabs/puppet/lib/ruby/vendor_ruby/facter/util/api_debugger.rb:37: syntax error, unexpected '{ arg'
  Mar 06 15:43:30 centos7-64-puppet7.example.com puppetserver[4297]: exclude << option[1..].to_sym
  Mar 06 15:43:30 centos7-64-puppet7.example.com puppetserver[4297]: ^
```

manifests/server/puppetserver.pp
evgeni: Oh right, this is puppet, not candlepin 🙈

evgeni:
You need to update the tests ;-)

evgeni (approved):

ehelms:
I can't seem to run these tests locally that are failing and its not obvious to me if these failures are related.

evgeni:
It looks unrelated to me

evgeni:
https://github.com/theforeman/puppet-puppet/pull/909 has CI fixes

PR #909, opened 2024-03-21
raise minimum tested version for upgrades and test upgrades on EL9 and Ubuntu 22.04

evgeni:


spec/acceptance/puppetserver_upgrade_spec.rb
```diff
@@ -17,10 +17,10 @@
     case ENV['BEAKER_PUPPET_COLLECTION']
     when 'puppet8'
       from_version = '8.2.0'
-      to_version = '8.3.0'
+      to_version = '8.5.0'
```
evgeni, line 20: this one is (obviously) not necessary, but I thought if we're already bumping ranges, we can go all the way :)

spec/support/acceptance/puppetserver.rb
```diff
@@ -11,6 +11,6 @@ def unsupported_puppetserver
 
 # These versions only have a single version (x.y.z) released so no upgrade is possible
```
ekohl, line 12: This is now outdated

evgeni: You want me to remove the method completely?

ekohl: I think just the comment because when puppetserver is (hopefully) eventually published for Debian 12 then we also can't upgrade until there are at least 2 packages published.

evgeni: You mean to delete the comment, instead of reading it and the next one as "these versions only have a single …: currently none"

Can do

ekohl (approved):

PR #910, opened 2024-03-22
Fixes #37291 - Use explicit java on RH with Puppetserver 8

ekohl:
The /usr/bin/java file may point to any version. This changes the logic to determine the java version (unless explicitly specified) dynamically based on the puppetserver version and the OS.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -305,7 +305,7 @@
 # $server_envs_target::                     Indicates that $envs_dir should be
 #                                           a symbolic link to this target
 #
-# $server_jvm_java_bin::                    Set the default java to use.
+# $server_jvm_java_bin::                    Set the default java to use. If unspecified, it will be derived from the Puppet version.
 #
 # $server_jvm_config::                      Specify the puppetserver jvm configuration file.
 #
@@ -714,7 +714,7 @@ class puppet (
   Optional[Stdlib::Absolutepath] $server_puppet_basedir = $puppet::params::server_puppet_basedir,
   Enum['current', 'future'] $server_parser = $puppet::params::server_parser,
   Variant[Undef, Enum['unlimited'], Pattern[/^\d+[smhdy]?$/]] $server_environment_timeout = $puppet::params::server_environment_timeout,
-  String $server_jvm_java_bin = $puppet::params::server_jvm_java_bin,
+  Optional[Stdlib::Absolutepath] $server_jvm_java_bin = undef,
   String $server_jvm_config = $puppet::params::server_jvm_config,
   Pattern[/^[0-9]+[kKmMgG]$/] $server_jvm_min_heap_size = $puppet::params::server_jvm_min_heap_size,
   Pattern[/^[0-9]+[kKmMgG]$/] $server_jvm_max_heap_size = $puppet::params::server_jvm_max_heap_size,
--- a/manifests/server/install.pp
+++ b/manifests/server/install.pp
@@ -41,6 +41,19 @@ class puppet::server::install {
       install_options => $puppet::package_install_options,
     }
 
+    # Puppetserver 8 on EL 8 relies on JRE 11 or 17. This prefers JRE 17 by installing it first
+    if (
+      !$puppet::server::jvm_java_bin and
+      $facts['os']['family'] == 'RedHat' and $facts['os']['release']['major'] == '8' and
+      # This doesn't use server_version because we have 2 mechanisms to set the version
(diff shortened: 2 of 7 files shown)


manifests/server/install.pp
```diff
@@ -41,6 +41,19 @@
       install_options => $puppet::package_install_options,
     }
 
+    # Puppetserver 8 on EL 8 relies on JRE 11 or 17. This prefers JRE 17 by installing it first
+    if (
+      !$puppet::server::java_bin and
+      $facts['os']['family'] and $facts['os']['release']['major'] == '8' and
```
vchepkov, line 47: family missing `==`

vchepkov (approved):

manifests/server/puppetserver.pp
```diff
@@ -163,6 +163,27 @@
 
   $puppetserver_package = pick($puppet::server::package, 'puppetserver')
 
+  if $java_bin {
+    $_java_bin = $java_bin
+  } elsif versioncmp($real_puppetserver_version, '8.0.0') >= 0 {
+    # Follows logic that https://github.com/puppetlabs/ezbake/pull/627 suggests, but takes it a
+    # step further by also ensuring EL 8 has Java 17
+    $java_bin = case $facts['os']['family'] {
```
vchepkov, line 171: should be `$_java_bin`

vchepkov (changes requested):

manifests/server/puppetserver.pp
ekohl: Today is not my day it appears.

manifests/init.pp
```diff
@@ -714,7 +714,7 @@
   Optional[Stdlib::Absolutepath] $server_puppet_basedir = $puppet::params::server_puppet_basedir,
   Enum['current', 'future'] $server_parser = $puppet::params::server_parser,
   Variant[Undef, Enum['unlimited'], Pattern[/^\d+[smhdy]?$/]] $server_environment_timeout = $puppet::params::server_environment_timeout,
-  String $server_jvm_java_bin = $puppet::params::server_jvm_java_bin,
+  Optional[Stdlib::Absolutepath] $server_jvm_java_bin = undef,
```
ekohl, line 717: This will need an installer migration as well to wipe the value. I wonder if this a breaking change, but I think a relative path was never valid.

ehelms (approved):

manifests/init.pp
ehelms: Agreed.

ekohl: https://github.com/theforeman/foreman-installer/pull/922

ekohl:
Coming back to this: it didn't resolve our upgrade issues.

What happens is:
* `dnf upgrade` runs the scriptlet that calls `/usr/bin/systemctl try-restart puppetserver.service`
* puppetserver enters a restart loop
* Puppet runs and can't reliably restart the service (bug in systemd? Puppet service provider?)

I've opened https://github.com/puppetlabs/ezbake/pull/627 to avoid this. On EL8 it may not be reliable since there's no explicit java binary there, since it relies on a virtual package. For EL9 it should be though.

ekohl:
Contemplating the solutions: I think the best is for the foreman-installer to add a hook that detects the broken situation (`/usr/bin/java` being too old) and then stops `puppetserver`. That should prevent the restart loop.

evgeni:
That doesn't sound crazy.

Do we tell people to do `f-m services stop` before the upgrade?

ekohl:
Not sure, but there's no harm in an explicit stop if it prevents issues. At least our upgrade pipelines hit the problem so we're not doing it.

evgeni:
Yeah, not arguing against your solution. More wondering how many people would actually run into that.

ekohl:
https://github.com/theforeman/foreman-installer/pull/953 is now ready. This allows users to also upgrade to Puppet 8 as a separate procedure:
* Update puppet-release
* Upgrade puppetserver
* Run installer

Now I just need to see if this is also a problem for Debian.

ekohl:
> Now I just need to see if this is also a problem for Debian.

When I do a fresh installation of Puppet 7 on Debian 11 it install Java 11, so that shouldn't be a problem. I'm sure we'll find some cases, but not something to generally solve for everyone.

Fabian1976:
Why make it a hard-coded name? Now our installation on RedHat 8 broke because it is called `java-17-openjdk-headless` on those system.

PR #912, opened 2024-03-23
Remove pidfile_workaround and instead use unsupported_puppetserver

ekohl:
This at least tests the agent on EL8.

Not sure if this works now.


evgeni (approved):

ekohl:
Now updated to only skip EL8 so we still test on EL7 & EL9. I noticed this by looking at test runtimes.

evgeni (approved):

PR #918, opened 2024-07-04
Remove localconfig (see PUP-880)

m8t:
https://puppet.atlassian.net/browse/PUP-880

(PUP-880) Remove localconfig setting

The localconfig setting wasn't used anywhere so remove it, including references in sample puppet.conf files for various distros.


ekohl (approved):
Thanks!

PR #921, opened 2024-11-22
Allow to set node_terminus variable

d1nuc0m:
Allow to override default value of [server]/node_terminus in puppet.conf.
Useful in case of standalone (non-Foreman) Puppet Server installs without an ENC.

### DIFF
--- a/manifests/server/config.pp
+++ b/manifests/server/config.pp
@@ -33,6 +33,7 @@ class puppet::server::config inherits puppet::config {
   ## General configuration
   $ca_server                   = $puppet::ca_server
   $ca_port                     = $puppet::ca_port
+  $server_node_terminus        = $puppet::server::node_terminus
   $server_external_nodes       = $puppet::server::external_nodes
   $server_environment_timeout  = $puppet::server::environment_timeout
   $trusted_external_command    = $puppet::server::trusted_external_command
@@ -40,7 +41,13 @@ class puppet::server::config inherits puppet::config {
 
   if $server_external_nodes and $server_external_nodes != '' {
     class { 'puppet::server::enc':
-      enc_path => $server_external_nodes,
+      node_terminus => $server_node_terminus,
+      enc_path      => $server_external_nodes,
+    }
+  }
+  else {
+    class { 'puppet::server::enc':
+      node_terminus => $server_node_terminus,
     }
   }
 
--- a/manifests/server/config.pp
+++ b/manifests/server/config.pp
@@ -44,8 +44,7 @@ class puppet::server::config inherits puppet::config {
       node_terminus => $server_node_terminus,
       enc_path      => $server_external_nodes,
     }
-  }
-  else {
+  } else {
     class { 'puppet::server::enc':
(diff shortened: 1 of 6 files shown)


manifests/server/config.pp
```diff
@@ -33,14 +33,21 @@
...
   $primary_envs_dir            = $puppet::server::envs_dir[0]
 
   if $server_external_nodes and $server_external_nodes != '' {
     class { 'puppet::server::enc':
-      enc_path => $server_external_nodes,
+      node_terminus => $server_node_terminus,
+      enc_path      => $server_external_nodes,
+    }
+  }
+  else {
```
ekohl, line 48: Nit: we put them on the same line
```suggestion
  } else {
```

manifests/server/config.pp
```diff
@@ -33,14 +33,21 @@
   ## General configuration
   $ca_server                   = $puppet::ca_server
   $ca_port                     = $puppet::ca_port
+  $server_node_terminus        = $puppet::server::node_terminus
   $server_external_nodes       = $puppet::server::external_nodes
   $server_environment_timeout  = $puppet::server::environment_timeout
   $trusted_external_command    = $puppet::server::trusted_external_command
   $primary_envs_dir            = $puppet::server::envs_dir[0]
 
   if $server_external_nodes and $server_external_nodes != '' {
```
ekohl, line 42: I think we should look at `$server_node_terminus` and only when it's set to `exec` this should we include the ENC. So perhaps:
```puppet
if $server_node_terminus == 'exec' {
  class { 'puppet::server::enc':
    node_terminus => $server_node_terminus,
    enc_path      => $server_external_nodes,
  }
}
```

Ideally we would change `$server_external_nodes` to `Optional[Stdlib::Absolutepath]` then and drop empty string handling. That could be considered a breaking change, but I think it's for the better.

I haven't look at all at the consequences of this so there's a good chance I'm missing something.

ekohl (commented):
Test failures look related.

Orest-Worhacz:
I bumped into this when setting new environment. As a workaround until solution is merged into new version of the module. I used this ugly code for dummy node.rb:

```
# Workaround for https://github.com/theforeman/puppet-puppet/pull/921
    file { '/etc/puppetlabs/puppet/node.rb':
      source => 'puppet:///modules/module_name/node.rb',
      mode   => '0777',
      notify => Service['puppetserver'],
    }
```
and ```cat modules/module_name/files/node.rb```
```
#!/bin/bash
echo "node $1 { notify { 'hello': dummy }}"
```
Ugly but simple and works. We use just plain puppet no foreman or whatsoever.

oworhacz-vtech (approved):

aba-lrettler:
I would love to see this merged...
I also wanna use the module without foreman.
Anything that blocks the issue?

bastelfreak:
@aba-lrettler are you interested in checking out the branch, rebase it and submit it as a new PR?

aba-lrettler:
@bastelfreak https://github.com/theforeman/puppet-puppet/pull/995

Issue #922, opened 2025-01-03
constant changing of ownership of vardir


nomennesc-io:
This happens on debian bookworm:

At the start of every agent run, the puppet agent seems to change the ownership of /var/cache/puppet to puppet:puppet

Deploying puppet::server::config causes the ownership to be changed to root:root

This change in ownership then causes a restart of puppetserver, which is kindof annoying to have happening on every puppet run.

I'm not sure why vardir is set to ownership root:root, is there any harm in having it owned by user puppet?

PR #923, opened 2025-01-03
stop forcing root ownership of vardir

nomennesc-io:
this should fix https://github.com/theforeman/puppet-puppet/issues/922

### DIFF
--- a/manifests/server/config.pp
+++ b/manifests/server/config.pp
@@ -231,8 +231,6 @@ class puppet::server::config inherits puppet::config {
   if $puppet::server::git_repo {
     file { $puppet::vardir:
       ensure => directory,
-      owner  => 'root',
-      group  => 'root',
     }
 
     vcsrepo { 'puppet_repo':


ekohl (changes requested):
In the AIO case this is what's shipped by default. From the EL9 RPM:
```console
# rpm -q --dump puppet-agent | grep /opt/puppetlabs/puppet/cache
/opt/puppetlabs/puppet/cache 0 1729190692 0000000000000000000000000000000000000000000000000000000000000000 040750 root root 0 0 0 X
```

I get the impression you're using the Debian OS packages. Perhaps we should have variables for this. We already detect the AIO package:

https://github.com/theforeman/puppet-puppet/blob/b4c69a6d25cc5ecc8a3923118841fbad2ea66c4c/manifests/params.pp#L122-L168

We also have a similar case for directory ownership:
https://github.com/theforeman/puppet-puppet/blob/b4c69a6d25cc5ecc8a3923118841fbad2ea66c4c/manifests/params.pp#L185-L191

I'm not sure we need parameters, but I'd prefer to have some code that's probably something like this:
```puppet
$vardir_owner = if !$puppet::aio_package and $facts['os']['family'] == 'Debian' { 'puppet' } else { 'root' }

file { $puppet::vardir:
  ensure => directory,
  owner  => $vardir_owner,
  group  => $vardir_owner,
}
```

nomennesc-io:
thanks for the review, i am indeed using the Debian OS packages!

spec/classes/puppet_server_spec.rb
```diff
@@ -353,7 +353,6 @@
         it do
           should contain_file(vardir)
             .with_ensure('directory')
-            .with_owner('root')
```
ekohl, line 356: If you add this back, does it still pass?

nomennesc-io: good point, looks like it does!

ekohl:
@evgeni mind checkimg my suggestion?

evgeni:
I have no idea about that specific piece of code and why it needs to manage the vardir, but the if/else logic seems fine to me

ekohl:
Perhaps it's no longer needed these days and we can drop it instead. By default it's in packages anyway and if you move it then all bets are off anyway. Dropping it doesn't sound too bad either

PR #926, opened 2025-02-12
Add convenience facts for puppet config/ssl

jcpunk:
This adds two facts that I find helpful for working with the agent.


lib/facter/puppet_settings.rb
```diff
@@ -0,0 +1,13 @@
+require 'puppet'
+
+Facter.add('puppet_ssldir') do
+  setcode do
+    Puppet.settings['ssldir']
```
bastelfreak, line 5: extlib already has a fact for this: https://github.com/voxpupuli/puppet-extlib/blob/master/lib/facter/extlib__puppet_config.rb

jcpunk: Neat!  I'll drop that one.

lib/facter/puppet_settings.rb
```diff
@@ -0,0 +1,7 @@
+require 'puppet'
+
+Facter.add('puppet_confdir') do
+  setcode do
+    Puppet.settings['confdir']
```
bastelfreak, line 5: Personally I would prefer it to move the facts from puppet_agent to extlib/stdlib, or to extend the extlib fact with `confdir`, instead of another module having facts related to the puppet config.

jcpunk: Fair enough, I'll see about getting extlib updated.

PR #931, opened 2025-03-05
Drop EL7 support

bastelfreak:


ekohl (commented):
Perhaps also simplify this condition:
https://github.com/theforeman/puppet-puppet/blob/ed5d6b45b56a5bd57b33ae420c4b06118c437a5e/spec/acceptance/puppetserver_latest_spec.rb#L28

bastelfreak:
Updated it

alexjfisher:
Personally, I’d prefer it if only server support was actively removed. I’m sure most EL users have some el7 yet to be upgraded. But if this is just params.pp and can be fixed up in calling profile code, then less of an issue.

ekohl (approved):

alexjfisher:
> > Personally, I’d prefer it if only server support was actively removed. I’m sure most EL users have some el7 yet to be upgraded. But if this is just params.pp and can be fixed up in calling profile code, then less of an issue.
> 
> Technically the non-systemd EL case in `params.pp` that's removed is really for EL6, not EL7.

ah yes; (was reading on my phone earlier).  I've still got a couple of those two, but since it's only module defaults it's easy enough to work-around.
In general, this is one of those modules I might treat differently to others.  It's in the catalog of _all_ of my nodes and it's not easy to just 'use an older version' so that it works with the old stuff as the newer version is needed for a modern Puppetserver Infrastructure.

alexjfisher (approved):

ekohl:
Looks like you'd lose service reloads and they turn into full restarts on EL6.

PR #929, opened 2025-03-05
Update to gha-puppet v3 & voxpupuli-test 9.x & Replace legacy facts in rspec tests

bastelfreak:

PR #932, opened 2025-03-05
Add Ubuntu 24.04 support

bastelfreak:
should be rebased after #929 got merged. the current facterdb version in master has no ubuntu 24.04 factsets.


bastelfreak:
Perforce doesn't provide packages, so we would need to use openvox here.

bwitt:
@bastelfreak want to rebase this?

PR #933, opened 2025-03-11
[WIP] Add OpenVox support

binford2k:
This appeaers to _mostly_ work. I cannot get a clean foreman-installer
run though. It will create the postgres `foreman` user, but not the
database. I can't imagine it being related to these changes, but still
troubleshooting.

I'll add tests once I figure out what's going on.


binford2k:
turns out I'm getting the same database creation error with unmodified installer on stock Debian 12 (as provisioned by GCE). I'll post a bug tomorrow, apparently redmine uses a different login.

PR #935, opened 2025-03-11
Add OpenVox support

bastelfreak:
This adds a fact that discovers if the Perforce or OpenVox agent is installed. If none are installed yet, we default to the Perforce package for backwards compatibility.

### DIFF
--- a/lib/facter/puppet_flavor.rb
+++ b/lib/facter/puppet_flavor.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+Facter.add('puppet_flavor') do
+  confine { Facter::Core::Execution.which('puppet') }
+  setcode do
+    output = Facter::Core::Execution.execute('puppet --help')
+    output.split[-2] if output
+  end
+end
+
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -263,22 +263,25 @@ class puppet::params {
 
   $puppet_major = regsubst($facts['puppetversion'], '^(\d+)\..*$', '\1')
 
-  if ($facts['os']['family'] =~ /(FreeBSD|DragonFly)/) {
-    $server_package = "puppetserver${puppet_major}"
+  # Add support for OpenVox. Default to puppet if nothing is installed yet
+  $puppet_flavor       = pick($facts['puppet_flavor'], 'puppet').downcase()
+  $puppetserver_flavor = regsubst("${puppet_flavor}server", 'openvox', 'openvox-')
+
+  $server_package = if ($facts['os']['family'] =~ /(FreeBSD|DragonFly)/) {
+    "${puppetserver_flavor}${puppet_major}"
   } else {
-    $server_package = undef
+    $puppetserver_flavor
   }
 
   $server_ssl_dir = $ssldir
   $server_version = undef
 
-  if $aio_package or
-  ($facts['os']['name'] == 'Debian' and versioncmp($facts['os']['release']['major'], '12') >= 0) {
-    $client_package = ['puppet-agent']
+  $client_package = if $aio_package or ($facts['os']['name'] == 'Debian' and versioncmp($facts['os']['release']['major'], '12') >= 0) {
+    ["${puppet_flavor}-agent"]
   } elsif ($facts['os']['family'] =~ /(FreeBSD|DragonFly)/) {
-    $client_package = ["puppet${puppet_major}"]
+    ["${puppet_flavor}${puppet_major}"]
   } else {
-    $client_package = ['puppet']
+    [$puppet_flavor]
   }
 
   # Puppet service name


bastelfreak:
This is a resubmission of https://github.com/theforeman/puppet-puppet/pull/933/

lib/facter/puppet_flavor.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+Facter.add('puppet_flavor') do
```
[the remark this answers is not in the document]
ekohl: Any thoughts on enhancing openvox itself to provide an `openvox_version` fact, just like there's `puppet_version`? I imagine other software could also use it.

Going the structured route a `puppet` fact could work:
```json
{
  "version": "x.y.z",
  "implementation": "openvox",
}
```

`vendor` might also be an option.

Feels like something for the language steering committee.

manifests/params.pp
```diff
@@ -263,22 +263,25 @@
 
   $puppet_major = regsubst($facts['puppetversion'], '^(\d+)\..*$', '\1')
 
-  if ($facts['os']['family'] =~ /(FreeBSD|DragonFly)/) {
-    $server_package = "puppetserver${puppet_major}"
+  # Add support for OpenVox. Default to puppet if nothing is installed yet
+  $puppet_flavor       = pick($facts['puppet_flavor'], 'puppet').downcase()
```
ekohl, line 267: Could we downcase the fact?

bastelfreak: fixed it

Stricken1670:
So I just installed this module and it's active, and I applied the 935.diff for this PR:

How can I uninstall `puppet-agent` and install `openvox-agent` with this module?

I tried this recipe, and it doesn't work.

```ruby
  if $facts['networking']['hostname'] == 'labrat' {
    class {
      'puppet':
        runmode => 'cron',
        puppet_flavor => 'openvox-8',
        agent_server_hostname => 'theforeman.example.com',
    }
  }
```

markeganfuller:
I think it's also required that `terminus_package` is set to `openvoxdb-termini` on `puppetdb::master::config` in [`puppet::server::puppetdb`.](https://github.com/theforeman/puppet-puppet/blob/b207a35cbef02777e83950386ba5e0a22781d60a/manifests/server/puppetdb.pp#L31)

lib/facter/puppet_flavor.rb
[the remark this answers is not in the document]
krisavi: Isn't facter responsible for providing facts? It does provide `aio_agent_version` already, why not add `variant`, `vendor`, `flavor`, `implementation` or whatever you are going to call it as extra fact there. Changing how version is printed there would not be backwards compatible. Adding separate fact for puppet-agent could help.

puppet_metadata seems to be another binary that has to be called out by custom fact, which doesn't qutie solve it for out of the box solution. 

Oh and about the check foreman is using, then that is not consistent. It may not be enough to know which one is used. At least in case of RHEL7-9 and puppet/openvox v7.*
```
 dnf info openvox-agent | head -n3
Installed Packages
Name         : openvox-agent
Version      : 7.36.1

 dnf list installed | grep puppet
puppet7-release.noarch                 7.0.0-20.el9                                @puppet7.   # that is only repo, not actual binary

 puppet --help | tail -n1
Puppet v7.36.1

 facter | grep version
aio_agent_version => 7.36.1
augeas => {
  version => "1.14.1"
}
dmi => {
  bios => {
    version => "6.00"
  },
}
facterversion => 4.2.8
kernelmajversion => 5.14
kernelversion => 5.14.0
os => {
  selinux => {
    policy_version => "33"
  },
}
ruby => {
  platform => "x86_64-linux",
  sitedir => "/usr/local/share/ruby/site_ruby",
  version => "3.0.7"
}
```
Cleaned up some output, but at least on v7 branch openvox agent doesn't report much differently from puppet agent.

manifests/server/puppetdb.pp
```diff
@@ -23,17 +23,23 @@
...
 #   Whether to enable soft write failure
+#
+# @param terminus_package
+#   The PuppetDB terminus package
+#
 class puppet::server::puppetdb (
   Stdlib::Host $server = undef,
   Stdlib::Port $port = 8081,
   Boolean $soft_write_failure = false,
+  String[1] $terminus_package = 'puppetdb-termini',
```
ekohl, line 34: https://github.com/theforeman/puppet-puppet/pull/936 is open for this same change and there I also suggested to make it `Optional`
```suggestion
  Optional[String[1]] $terminus_package = undef,
```

bastelfreak:
@sbernhard as a first step, https://github.com/theforeman/puppet-puppet/pull/955 should be merged to remove the broken puppet 7 integration (and before that maybe do a minor release).

Stricken1670:
🥳

PR #938, opened 2025-03-31
Add bolt task to install the openvox agent

lbetz:
I have written a first draft for a task to install the OpenVox Agent on Linux systems.
Tested on Debian 10-12, Ubuntu 18.04-24.04, EL (Alma) 8/9.

As I understand it, would only acceptence tests be possible to write?

### DIFF
--- a/tasks/install_agent_shell.sh
+++ b/tasks/install_agent_shell.sh
@@ -422,6 +422,8 @@ if [ -f "$PT__installdir/facts/tasks/bash.sh" ]; then
   # Use facts module bash.sh implementation
   os=$(bash $PT__installdir/facts/tasks/bash.sh "platform")
   os_version=$(bash $PT__installdir/facts/tasks/bash.sh "release")
+  # Major OS Release
+  platform_version=$(echo $os_version | cut -d. -f1)
 
   case $os in
     "RedHat"|"Almalinux"|"Rocky"|"OracleLinux"|"CentOS")
@@ -449,6 +451,9 @@ if [ -f "$PT__installdir/facts/tasks/bash.sh" ]; then
       info "${os} platform! Lets get you a DEB..."
       pkg_type=deb
       platform=$(echo $os | tr '[:upper:]' '[:lower:]')
+      if [ $os == "Ubuntu" ]; then
+        platform_version=$os_version
+      fi
       ;;
     "Linuxmint"|"LinuxMint")
       info "Mint platform! Lets get you a DEB..."
@@ -467,9 +472,6 @@ if [ -f "$PT__installdir/facts/tasks/bash.sh" ]; then
       cleanup 1
       ;;
   esac
-
-  # Major OS Release
-  platform_version=$(echo $os_version | cut -d. -f1)
 else
   echo "This module depends on the puppetlabs-facts module"
   cleanup 1


tasks/install_agent_shell.sh
```diff
@@ -0,0 +1,690 @@
+#!/usr/bin/env bash
+
+set -e
+
+#####
+# Install openvox-agent as a task
+#
+
+# Error if non-root
+if [ `id -u` -ne 0 ]; then
```
bastelfreak, line 10: we don't have shellcheck in our pipeline right now. Could you/did you validate the file with shellcheck and check it for potential styleguide validations? I think it e.g. prefers `$()` over ``

tasks/install_agent.json
```diff
@@ -0,0 +1,50 @@
...
+      "description": "optional install arguments to the windows installer (defaults to REINSTALLMODE=\"amus\")",
+      "type": "Optional[String]"
+    },
+    "stop_service": {
+      "description": "Whether to stop the voxpupuli agent service after install",
+      "type": "Optional[Boolean]"
+    },
+    "retry": {
+      "description": "The number of retries in case of network connectivity failures",
+      "type": "Optional[Integer]",
```
bastelfreak, line 44: I think optional doesn't make sense when there's a default value? This could just be `Integer[0]`?

tasks/install_agent.json
```diff
@@ -0,0 +1,50 @@
+{
+  "description": "Install the OpenVox agent package",
+  "private": true,
+  "input_method": "environment",
+  "parameters": {
+    "version": {
+      "description": "The version of openvox-agent to install",
+      "type": "Optional[String]"
```
bastelfreak, line 8: Do empty strings make sense here? Otherwise you could update it to:
```suggestion
      "type": "Optional[String[1]]"
```

This prohibits empty strings.

tasks/install_agent_shell.sh
```diff
@@ -0,0 +1,690 @@
...
+}
+
+#####
+# Source options
+#
+
+if [ -n "$PT_yum_source" ]; then
+  yum_source=$PT_yum_source
+else
+  if [ "$nightly" = true ]; then
```
ekohl, line 372: Isn't `elif` also available in POSIX sh?
```suggestion
elif [ "$nightly" = true ]; then
```

tasks/install_agent_shell.sh
```diff
@@ -0,0 +1,690 @@
...
+    "Linuxmint"|"LinuxMint")
+      info "Mint platform! Lets get you a DEB..."
+      pkg_type=deb
+      case $os_version in
+        "4")  platform="debian"; platform_version="10";;
+        "5")  platform="debian"; platform_version="11";;
+        "6")  platform="debian"; platform_version="12";;
+        "21") platform="ubuntu"; platform_version="22.04";;
+        "20") platform="ubuntu"; platform_version="20.04";;
+        "19") platform="ubuntu"; platform_version="18.04";;
```
ekohl, line 446: Should we keep the list sorted?

tasks/install_agent_shell.sh
lbetz: Is this relevant if a bash is required?

tasks/install_agent_shell.sh
lbetz: I think your right.

tasks/install_agent_shell.sh
ekohl: I thought you were going for maximum compatibility. If you can rely on bash, I'd certainly use the shorter `elif`.

And if you require bash, I'd encourage bash checks.
```suggestion
elif [[ $nightly == true ]]; then
```

This is safer, in case there is some weird garbage in the variable.

tasks/install_agent_shell.sh
lbetz: done

tasks/install_agent.json
lbetz: correct, done.

tasks/install_agent_shell.sh
lbetz: done

tasks/install_agent_shell.sh
lbetz: But it seems to be
https://queirozf.com/entries/posix-shell-tests-and-conditionals-examples-and-reference

lbetz: maximum compatibility would be fine. But do you think there will be a platform without bash for the OpenVox agent?

lbetz: But what would be the behavior if no bash exists on a platform because line 1 is '/usr/bin/env bash'?

tasks/install_agent.json
```diff
@@ -0,0 +1,55 @@
+{
+  "description": "Install the OpenVox agent package",
+  "parameters": {
+    "version": {
+      "description": "The version of openvox-agent to install (defaults to latest when no agent is installed)",
+      "type": "Optional[String[1]]"
+    },
+    "collection": {
+      "description": "The OpenVox collection to install from (defaults to openvox8, which maps to the latest collection released)",
+      "type": "Optional[Enum[openvox7, openvox8]]"
```
bastelfreak, line 10: Does the enum need to contain `nightly`? And I think the logic in the task itself is easier if the default is already set in the json?

tasks/install_agent.json
```diff
@@ -0,0 +1,55 @@
...
+      "description": "The number of retries in case of network connectivity failures",
+      "type": "Integer[0]",
+      "default": 5
+    }
+  },
+  "implementations": [
+    {
+      "name": "install_agent_shell.sh",
+      "requirements": ["shell"],
+      "files": ["facts/tasks/bash.sh"],
```
bastelfreak, line 50: is that file coming from the puppetlabs/facts module? Then it needs to be added to metadata.json?

tasks/install_agent.json
lbetz: At the time, OpenVox hasn't any nightly builds. But default to json it's ok.

tasks/install_agent.json
lbetz: This module comes with bolt itself.

tasks/install_agent.json
lbetz: I had a look at the code. And neither version, collection or nightly can get a default.
- nightly: is overridden if nightly is included in collection parameter. The enum doesn't have this collections at the time because OpenVox doesn't offer nightly builds, yet. I will remove this nightly stuff since we have nightly builds.
- collection: if the puppet agent is installed and NO collection is set, the major version of the puppet agent is taken
- version: default to latest would result in an update every runtime

lbetz: done

tasks/install_agent_shell.sh
lbetz: I removed parameter nightly since OpenVox offers such builds.

PR #939, opened 2025-04-25
Add timezone param for systemd.timer

zipkid:


manifests/agent/service/systemd.pp
```diff
@@ -4,6 +4,7 @@
   Boolean                 $enabled                             = false,
   Optional[Integer[0,23]] $hour                                = undef,
   Variant[Integer[0,59], Array[Integer[0,59]], Undef] $minute  = undef,
+  Optional[String] $timezone                                   = undef,
```
bastelfreak, line 7: are emtpy strings allowed, or can we narrow down the datatype a bit?

zipkid: it works as expected when not providing that param.
It does change the timer config as a space is then added, but the timer still works.

zipkid: `(/Stage[main]/Puppet::Agent::Service::Systemd/Systemd::Timer[puppet-run.timer]/Systemd::Unit_file[puppet-run.timer]/File[/etc/systemd/system/puppet-run.timer]/content) -OnCalendar=*-*-* *:24,54:00
(/Stage[main]/Puppet::Agent::Service::Systemd/Systemd::Timer[puppet-run.timer]/Systemd::Unit_file[puppet-run.timer]/File[/etc/systemd/system/puppet-run.timer]/content) +OnCalendar=*-*-* 10:0:00 Europe/Brussels

result in systemctl show puppet-run.timer
TimersCalendar={ OnCalendar=*-*-* 10:00:00 Europe/Brussels ; next_elapse=Sat 2025-04-26 08:00:00 UTC }`

TomRitserveldt (approved):

manifests/agent/service/systemd.pp
bastelfreak: yes, but do we want to allow `''` as values? I think that doesn't make sense and it should be `String[1]`, so empty strings aren't allowed?

zipkid: I can add that, but having an empty, or multiple spaces even, string would still be ok, as this just gets added to the end of the line in the systemd timer config file, so that will not make any difference at all.

bastelfreak (approved):

templates/agent/systemd.puppet-run.timer.erb
```diff
@@ -5,7 +5,7 @@
 Description=Systemd Timer for Puppet Agent
 
 [Timer]
-OnCalendar=*-*-* <%= Array(@_hour).join(',') %>:<%= Array(@_minute).join(',') %>:00
+OnCalendar=*-*-* <%= Array(@_hour).join(',') %>:<%= Array(@_minute).join(',') %>:00 <%= @timezone %>
```
ekohl, line 8: I'd like to minimize the changes on updates. By default timezone will be empty so I wonder what's the least ugly way to avoid the trailing whitespace when it's empty.

manifests/init.pp
```diff
@@ -72,6 +72,10 @@
 # $run_minute::                             The minute at which to run the puppet agent
 #                                           when runmode is cron or systemd.timer.
 #
+# $run_timezone::                           The timezon for the run the puppet agent
```
ekohl, line 75:
```suggestion
# $run_timezone::                           The timezone for the run the puppet agent
```

ekohl (approved):

PR #940, opened 2025-05-06
Drop support for puppetserver on non-systemd

bastelfreak:
Foreman doesn't support operatingsystems for puppetserver without systemd anymore, so we can remove the logic.


ekohl (changes requested):
FreeBSD doesn't have system and have explicit code in the same file for FreeBSD. https://cgit.freebsd.org/ports/tree/sysutils/puppetserver7 still exists so for now this isn't correct.

bastelfreak:
Ah I assumed you currently only support the puppet agent on FreeBSD.

PR #944, opened 2025-06-13
Add support for Sensitive values in server additional_settings

### LINKED ISSUE #848: Add support for sensitive config
I use a proxy with a password in my `puppet.conf`, and would like it not to be displayed in the logs when the config is pushed.  It would be nice for this class to support the `Sensitive` type.

```
$secret = Sensitive('mysecretpassword')
class { 'puppet':
  agent_additional_settings => {
    'http_proxy_user      => 'proxyuser',
    'http_proxy_password' => $secret,
  }
}
```

I’ve opened an issue on the `concat` module that need a change before it could be addressed on this module: https://github.com/puppetlabs/puppetlabs-concat/issues/742

For now, I’ve disabled the diff completely with a default value for `concat` module
```
Concat {
  show_diff => false,
}
```

optiz0r:
Support for Sensitive values was implemented in concat for 7.4.0+. This module currently has a dependency on concat >=4.10.0.

This change introduces a soft dependency on concat 7.4 to make use of Sensitive values for the `server_additional_settings`, `additional_settings` parameters. I don't think this change justifies bumping the minimum version of concat.

Only support for `Sensitive[String]` is added to avoid overcomplicating the parameter type definitions. I don't think other values would necessarily be sensitive, but would be fairly easy to add.

Fixes #848


optiz0r:
Looked like it worked, in fact did not, just plaintexted `Sensitive [value redacted]` into the file :(

Issue #593, opened 2018-05-08
Make agent 'environment' optional


nick-george:
Hi, 

It would be great if you could make the agent's environment setting in puppet.conf be optional (see https://github.com/theforeman/puppet-puppet/blob/master/manifests/agent/config.pp#L10). Quite often when I change an agent's environment (using Foreman), puppet fails all subsequent agent runs until I manually fix the puppet.conf on the agent to remove the environment setting, or change it to be the correct environment. I imagine this is a bug in puppet (agent v5.5.1), but it would also be handy if we could drop this setting entirely from the puppet.conf. 

Given that most Puppet Foreman users would be forcing the directory environment from the server side, I'm surprised to see this setting is being forced. 

Many thanks, 
Nick George

vchepkov:
As a workaround I have been doing this trick:
```
environment => $server_facts['environment'],
```

nick-george:
Thanks for that tip Vadym. 

I suspect if do this, I will still get errors when I change a host's directory environment. It seems the puppet master (or agent) can't handle the fact that the agent is specifying one environment when the master is specifying another. Unfortunately I don't have the error message on hand. When I'm next in the office, I'll see if I can get it. 

Cheers, 
Nick

mmoll:
PRs welcome :)

nick-george:
For reference, the issue I mentioned earlier in this thread is PUP-7734 (https://tickets.puppetlabs.com/browse/PUP-7734). I'd be able to work around it if I could remove the environment setting from puppet.conf. 

I'm still waiting for word on whether or not I can spend the time on this. I imagine the change itself would take a minute or two, but it's the associated tests that will take the time (for me). 

Cheers, 
Nick

achevalet:
In the current implementation, the default value refers to the fact `environment`. To keep this behaviour, we could add a new flag `use_environment_fact` which is `true` by default. If set to `false` and `$environment` is empty the key is removed from puppet.conf. Thoughts?

nod0n:
I do change the environment by running the agent once with `--environment` and then it changes the environment in the puppet agent config. But beeing able to drop the line from the config entirely, would also be nice. Some times I just run the agent twice:
1. `--environment temporary`
2. `--environment default`

jplindquist:
Sorry to dig up an old issue, but I'm wondering if there's a good solution for this to make `environment` option, or configure it in some other way.  

[...]

PR #951, opened 2025-09-10
Manage agent environment based on param

bwitt:
Only set the agent environment if `agent_manage_environment` is true

Fixes https://github.com/theforeman/puppet-puppet/issues/593


bwitt:
those failures do not seem related?

bwitt:
thanks!

PR #952, opened 2025-09-17
Allow puppet/systemd 9.x

evgeni:


bastelfreak (approved):

bastelfreak:
Not sure where those failures are coming from :thinking:

PR #965, opened 2025-10-02
puppet::server::puppetdb: Fix puppet-strings documentation

bastelfreak:
Each example needs a title, otherwise the REFERENCE.md is a bit broken


bastelfreak:
@sbernhard done

ekohl (approved):

PR #982, opened 2026-04-03
Switch puppetdb->openvoxdb module

bastelfreak:
Vox Pupuli provides their own module to manage openvoxdb. Also openvox-server has the privatetmp setting that was explicitly configred.

---

This is still work in progress and we need to figure out a migration path.

### DIFF
--- a/manifests/server/puppetdb.pp
+++ b/manifests/server/puppetdb.pp
@@ -1,7 +1,7 @@
 # @summary PuppetDB integration
 #
-# This class relies on the puppetlabs/puppetdb and essentially wraps
-# puppetdb::master::config with the proper resource chaining.
+# This class relies on the puppet/openvoxdb and essentially wraps
+# openvoxdb::master::config with the proper resource chaining.
 #
 # Note that this doesn't manage the server itself.
 #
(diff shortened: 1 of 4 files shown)


manifests/server/puppetdb.pp
```diff
@@ -33,13 +33,13 @@
   Boolean $soft_write_failure = false,
   Optional[String[1]] $terminus_package = undef,
 ) {
-  class { 'puppetdb::master::config':
+  class { 'openvoxdb::master::config':
```
bastelfreak, line 36: Or we duplicate the class into `puppet::server::openvoxdb`? 🤔 Or call `puppetdb::master::config` vs `openvoxdb::master::config` depending on if we have `puppetdb` vs `openvoxdb` modules in our environment?

seriv:
We've been running this exact unconditional switch in production since 2026-04-30 with puppet/openvoxdb 9.0.1 on RHEL 10. The catalog compiles cleanly, the puppetdb.service was unaffected during the Puppet agent run, and PuppetDB has been healthy since. Happy to confirm the approach works for sites that have already migrated their packages to OpenVox.

kenyon (approved):

manifests/server/puppetdb.pp
ehelms: Are you asking because:

- Preventing users from having to update their puppet code due to openvox support?
- Expecting there may be differences between openvox and puppet that require separate implementations?

bastelfreak: I don't expect a different implementation. But I don't know your support policy. Do you want to keep puppet support and openvox in parallel? Can we drop puppet in favour of openvox?

PR #985, opened 2026-04-30
Gemfile: Add Ruby 3.4 support & update gems

bastelfreak:
The older versions were only working on Ruby 3.2. That made it challenging to run the tests locally on a modern operating system. This PR updates all dependencies and aligns them with the Vox Pupuli setup.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,13 +7,12 @@ gem 'puppet', ENV.fetch('PUPPET_GEM_VERSION', '>= 8'), groups: ['development', '
 gem 'rake'
 
 gem 'kafo_module_lint', {"groups"=>["test"]}
-gem 'puppet-lint-spaceship_operator_without_tag-check', '~> 1.0', {"groups"=>["test"]}
-gem 'voxpupuli-test', '~> 9.0', {"groups"=>["test"]}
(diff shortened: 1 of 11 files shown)


Gemfile
```diff
@@ -7,13 +7,12 @@ gem 'puppet', ENV.fetch('PUPPET_GEM_VERSION', '>= 8'), groups: ['development', '
 gem 'rake'
 
 gem 'kafo_module_lint', {"groups"=>["test"]}
-gem 'puppet-lint-spaceship_operator_without_tag-check', '~> 1.0', {"groups"=>["test"]}
-gem 'voxpupuli-test', '~> 9.0', {"groups"=>["test"]}
+gem 'puppet-lint-spaceship_operator_without_tag-check', '~> 2.0', {"groups"=>["test"]}
+gem 'voxpupuli-test', '~> 14.0', {"groups"=>["test"]}
```
evgeni, line 11: This would be overridden on the next msync again, as https://github.com/theforeman/foreman-installer-modulesync/blob/master/config_defaults.yml is not updated. Care to do so too? :)

bastelfreak: https://github.com/theforeman/foreman-installer-modulesync/pull/178

PR #995, opened 2026-07-20
feat: allow to set node_terminus variable

aba-lrettler:
Allow to override default value of [server]/node_terminus in puppet.conf.
Useful in case of standalone (non-Foreman) Puppet Server installs without an ENC.

Copy of: #921 
Rebased on to master

### DIFF
--- a/manifests/server/config.pp
+++ b/manifests/server/config.pp
@@ -33,6 +33,7 @@ class puppet::server::config inherits puppet::config {
   ## General configuration
   $ca_server                   = $puppet::ca_server
   $ca_port                     = $puppet::ca_port
+  $server_node_terminus        = $puppet::server::node_terminus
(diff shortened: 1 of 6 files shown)


manifests/server/config.pp
```diff
@@ -33,14 +33,38 @@
...
+  }
+
   if $server_external_nodes and $server_external_nodes != '' {
     class { 'puppet::server::enc':
-      enc_path => $server_external_nodes,
+      node_terminus => $server_node_terminus,
+      enc_path      => $server_external_nodes,
+    }
+  } else {
+    class { 'puppet::server::enc':
```
bastelfreak: this leads to duplicate class declaration

aba-lrettler: @bastelfreak fixed.
All tests now passes

PR #1001, opened 2026-09-18
permissions for openvox primary

marcusdots:
OpenVox has decided that a primary should be able to clean its own cache after code deploy. pp_cli_auth-cert-extensions has been indentified as an existing path available on primaries after certificate generation.


bastelfreak (approved):
this will be introduced in openvox-server 9.0.0-rc2 and 8.17.0. Another review from a Foreman maintainer would be appreciated.

